Issue #517, opened 2019-10-14
Add shortcut for commenting in script editor


noramullen1:
https://github.com/influxdata/influxdb/pull/15313

PR #519, opened 2019-10-14
Add shortcut info (addresses #517 )

noramullen1:


content/v2.0/visualize-data/explore-metrics.md
```diff
@@ -35,7 +35,7 @@ See [Get started with Flux](/v2.0/query-data/get-started) to learn more about Fl
    To switch back to the query builder, click **Query Builder**. Note that your updates from the Script Editor will not be saved.
 3. Use the **Functions** list to review the available Flux functions.
    Click on a function from the list to add it to your query.
-4. Click **Submit** to run your query. You can then preview your graph in the above pane.
+4. Click **Submit** (or press `Control+Enter`) to run your query. You can then preview your graph in the above pane.
```
sanderson: This is more of a question than I comment. Should we format keypress combinations as code or make them bold? I've done both. This is something we should probably standardize.

noramullen1: @sanderson I had the same thought so I checked out Google's style guide and it looks like they use code blocks: https://developers.google.com/style/ui-elements

sanderson: Cool, I'm good with that.

PR #528, opened 2019-10-16
Redesigned homepage

sanderson:
The original homepage design was put together fairly quickly for the initial alpha release. The content near the bottom of the original was put there simply to fill vertical space. This updated design is meant to simplify the homepage and make the two getting started paths the focal point.

**Old design**  
image

**New design**  
_For the "task categories" on the right, you get a bit more information when you hover._

image
image


kelseiv:
🏆 LOVE the new look!

Issue #245, opened 2019-05-21
Best practices docs


sanderson:
We need to create best practices and info docs for 2.0. Examples from 1.x:

- [x] [Schema and data layout](https://docs.influxdata.com/influxdb/v1.7/concepts/schema_and_data_layout/)
- [x] [Insights and tradeoffs](https://docs.influxdata.com/influxdb/v1.7/concepts/insights_tradeoffs/)
- [x] [Storage Engine](https://docs.influxdata.com/influxdb/v1.7/concepts/storage_engine/)

Other thoughts:
- [ ] Using Organization IDs vs Organization names in queries. (#204)

kelseiv:
**Schema and data layout:**
- InfluxDB data elements: https://v2.docs.influxdata.com/v2.0/reference/key-concepts/data-elements/
- InfluxDB table structure: https://v2.docs.influxdata.com/v2.0/reference/key-concepts/table-structure/

**Insights and tradeoffs**:
- InfluxDB design principles: https://v2.docs.influxdata.com/v2.0/reference/key-concepts/design-principles/

@pierwill working on Storage Engine best practices

kelseiv:
@sanderson @pierwill closing re work tracking in #399

PR #542, opened 2019-10-22
add InfluxDB table structure topic

kelseiv:
Updated per #245

### DIFF
--- a/content/v2.0/reference/key-concepts/table-structure.md
+++ b/content/v2.0/reference/key-concepts/table-structure.md
@@ -1,5 +1,4 @@
 ---
-draft: true
 title: InfluxDB table structure
 description: >
   InfluxDB uses a columnar system to structure tables.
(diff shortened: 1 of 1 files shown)


content/v2.0/reference/key-concepts/table-structure.md
```diff
@@ -17,6 +16,7 @@ InfluxDB 2.0 uses the following columnar table structure to store data:
 - **Header row:** describes the data labels for each column in a row.
 - **Data columns:** include the following columns: annotation, result, and table.
 - **Data rows:** all rows that contain time series data. For details about the type of data stored in InfluxDB, see [InfluxDB data elements](/v2.0/reference/key-concepts/data-elements/).
+- **Group keys** modify output tables in Flux by grouping records that share common values in specified columns. Learn more about [grouping your data with Flux](/v2.0/query-data/guides/group-data/).
```
sanderson: **Group keys** ~modify~ determine the contents of output tables...

This is a minor nit, so feel free to ignore. "Modify" feels very active, like it's something they're continually doing. Group keys are the result of user input and only change when acted upon. Like I said, really minor, feel free to ignore.

PR #573, opened 2019-10-30
add Flux intro

kelseiv:
add Flux intro to how-to guides


kelseiv:
Scott, I had started to do that and think adding headers will help. Will you check out these topics, specifically:
- exists.md
- sort-limit.md
-window-aggregates.md

PR #574, opened 2019-10-30
Task description

noramullen1:
Updating task description from the task list


content/v2.0/process-data/manage-tasks/update-task.md
```diff
@@ -30,6 +30,11 @@ Click on the name of a task to update it.
 In the list of tasks, click the {{< icon "toggle" >}} toggle to the left of the
 task you want to activate or inactivate.
 
+#### Update a task's description
```
sanderson: I want to saw that we're avoiding possessive form where possible (something Kelly and Steve talked about way back when). What about:

> Update a task description

Issue #565, opened 2019-10-30
Document predicate delete syntax


sanderson:
The `/delete` endpoint uses a SQL-like predicate syntax to identifies which points should be deleted. We need to document how to use this syntax. Users will need to know it if deleting data in raw API requests or when using the `influx delete` command.

##### Relevant URLs
https://v2.docs.influxdata.com/v2.0/reference/cli/influx/delete _(will be live when alpha-19 rolls out)_

PR #577, opened 2019-10-30
Reference/syntax

### LINKED ISSUE #566: RFC: New "Syntaxes" reference section
With the addition of a predicate delete syntax (#565), I think we should create a "Syntaxes" reference section that has the following:

- **Line protocol** - Just move the existing line protocol doc here.
- **Flux** - Provide an overview of Flux that links to the existing Flux docs. I think it's important to keep existing Flux docs on the top level reference.
- **Annotated CSV** - Move the existing annotated CSV doc here.
- **Delete predicates** _or whatever we end up calling this syntax_

sanderson:
Closes #566 and #565

Created a new "Syntax" reference section and moved the Annotated CSV and Line protocol docs there. Added a Flux summary doc and a new Delete predicate syntax doc.

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/syntax/delete-predicate.md
+++ b/content/v2.0/reference/syntax/delete-predicate.md
@@ -1,0 +1,82 @@
+---
+title: Delete predicate syntax
+list_title: Delete predicate
+description: >
+  The InfluxDB `/delete` endpoint uses a SQL-like predicate syntax to determine
+  what data points to delete.
+menu:
+  v2_0_ref:
+    parent: Syntax
+    name: Delete predicate
+weight: 104
+v2.0/tags: [syntax, delete]
+related:
+  - /v2.0/reference/cli/influx/delete/
+---
+
+The InfluxDB `/delete` endpoint uses a SQL-like predicate syntax to determine
+what data [points](/v2.0/reference/glossary/#point) to delete.
+InfluxDB uses the delete predicate to evaluate each point in the time range
+specified in the delete request.
+Points that evaluate to `true` are deleted.
+Points that evaluate to `false` are preserved.
+
+A delete predicate is comprised of one or more [predicate expressions](/v2.0/reference/glossary/#predicate-expression).
+The left operand of the predicate expression is the column name.
+The right operand is the column value.
+Operands are compared using [comparison operators](#comparison-operators).
+Use [logical operators](#logical-operators) to combine two or more predicate expressions.
+
+##### Example delete predicate
+```sql
+key1="value1" AND key2="value"
+```
+{{% note %}}
+Predicate expressions can use any column or tag except `_time` or `_value`.
+{{% /note %}}
+
+## Logical operators
+Logical operators join two or more predicate expressions.
+
+| Operator | Description                                                                  |
+|:-------- |:-----------                                                                  |
+| `AND`    | Both left and right operands must be `true` for the expression to be `true`. |
+
+## Comparison operators
+Comparison operators compare left and right operands and return `true` or `false`.
+
+| Operator | Description  | Example        | Result |
+|:-------- |:-----------  |:-------:       |:------:|
+| `=`      | Equal to     | `"abc"="abc"`  | `true` |
+| `!=`     | Not equal to | `"abc"!="def"` | `true` |
+
+
+## Delete predicate examples
+
+### Delete a measurement
+The following will delete points in the `sensorData` measurement:
+
+```sql
+_measurement="sensorData"
+```
+
+### Delete a field
+The following will delete points with the `temperature` field:
+
+```sql
+_field="temperature"
+```
+
+### Delete points with a specific tag set
+The following will delete points from the `prod-1.4` host in the `us-west` region:
+
+```sql
+host="prod-1.4" AND region="us-west"
+```
+
+## Limitations
+The delete predicate syntax has the following limitations.
+
+- Delete predicates do not support regular expressions.
+- Delete predicates do not support the `OR` logical operator.
+- Delete predicates can use any column or tag except `_time` or `_value`.
--- a/content/v2.0/reference/syntax/delete-predicate.md
+++ b/content/v2.0/reference/syntax/delete-predicate.md
@@ -2,7 +2,7 @@
 title: Delete predicate syntax
 list_title: Delete predicate
 description: >
-  The InfluxDB `/delete` endpoint uses a InfluxQL-like predicate syntax to determine
+  The InfluxDB `/delete` endpoint uses an InfluxQL-like predicate syntax to determine
   what data points to delete.
 menu:
   v2_0_ref:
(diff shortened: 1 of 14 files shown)


content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
...
+## Delete predicate examples
+
+### Delete a measurement
+The following will delete points in the `sensorData` measurement:
+
+```sql
+_measurement="sensorData"
+```
+
+### Delete a field
```
kelseiv: Delete points with a specific field?

content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
...
+
+| Operator | Description  | Example        | Result |
+|:-------- |:-----------  |:-------:       |:------:|
+| `=`      | Equal to     | `"abc"="abc"`  | `true` |
+| `!=`     | Not equal to | `"abc"!="def"` | `true` |
+
+
+## Delete predicate examples
+
+### Delete a measurement
```
kelseiv: Delete points with a specific measurement?

content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Delete a field
+The following will delete points with the `temperature` field:
+
+```sql
+_field="temperature"
+```
+
+### Delete points with a specific tag set
+The following will delete points from the `prod-1.4` host in the `us-west` region:
```
kelseiv: what command will delete the points?

content/v2.0/reference/syntax/flux.md
```diff
@@ -0,0 +1,36 @@
...
+v2.0/tags: [syntax, flux]
+---
+
+Flux is a functional data scripting language designed for querying, analyzing, and acting on data.
+
+## Flux design principles
+Flux takes a functional approach to data exploration and processing, but is designed
+to be usable, readable, flexible, composable, testable, contributable, and shareable.
+
+```js
```
kelseiv: perhaps explain this example

content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Delete predicate syntax
+list_title: Delete predicate
+description: >
+  The InfluxDB `/delete` endpoint uses a SQL-like predicate syntax to determine
```
e-dard: `influxql-like`

content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
...
+  v2_0_ref:
+    parent: Syntax
+    name: Delete predicate
+weight: 104
+v2.0/tags: [syntax, delete]
+related:
+  - /v2.0/reference/cli/influx/delete/
+---
+
+The InfluxDB `/delete` endpoint uses a SQL-like predicate syntax to determine
```
e-dard: `influxql-like`

content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
...
+what data [points](/v2.0/reference/glossary/#point) to delete.
+InfluxDB uses the delete predicate to evaluate each point in the time range
+specified in the delete request.
+Points that evaluate to `true` are deleted.
```
e-dard: I don't have a strong opinion on this but the predicate applies to the _series_ and timestamp only, not the value of fields on the point. So in that sense I wonder if it's clearer to say something like:

> Points within the specified time range who have series keys that evaluate to true for a given predicate are deleted.

content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
...
+|:-------- |:-----------                                                                  |
+| `AND`    | Both left and right operands must be `true` for the expression to be `true`. |
+
+## Comparison operators
+Comparison operators compare left and right operands and return `true` or `false`.
+
+| Operator | Description  | Example        | Result |
+|:-------- |:-----------  |:-------:       |:------:|
+| `=`      | Equal to     | `"abc"="abc"`  | `true` |
+| `!=`     | Not equal to | `"abc"!="def"` | `true` |
```
e-dard: I checked and `!=` isn't supported yet. We will add it I'm sure, but I would omit it for now.

content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+## Delete predicate examples
+
+### Delete a measurement
+The following will delete points in the `sensorData` measurement:
+
+```sql
+_measurement="sensorData"
```
e-dard: FYI - this should work but from my testing it looks like there is a defect. That will be fixed and shipped for the next Alpha. https://github.com/influxdata/influxdb/issues/15677

content/v2.0/reference/syntax/delete-predicate.md
e-dard: Same https://github.com/influxdata/influxdb/issues/15677

content/v2.0/reference/syntax/delete-predicate.md
```diff
@@ -0,0 +1,82 @@
...
+```sql
+host="prod-1.4" AND region="us-west"
+```
+
+## Limitations
+The delete predicate syntax has the following limitations.
+
+- Delete predicates do not support regular expressions.
+- Delete predicates do not support the `OR` logical operator.
+- Delete predicates can use any column or tag except `_time` or `_value`.
```
e-dard: Looks like this feature shipped with only equality support, so `!=` doesn't work right now.

content/v2.0/reference/syntax/delete-predicate.md
sanderson: I hesitate to include specific commands here since this docs focuses primarily on the syntax. I think it would be better to create a task-based "Delete data" doc that covers the different ways to delete data.

A delete predicate can be used in a couple of different ways (UI, CLI, direct API). The UI is still being built for deletes, but the CLI and API commands would look like this:

**influx CLI**
```sh
influx delete -o my-org -b my-bucket \
  --start 2000-01-01T00:00:00Z \
  --stop 2020-01-01T00:00:00Z \
  -p '_measurement="someMeasurement" and tag="someTag"'
```

**InfluxDB API**
```sh
curl -XPOST http://localhost:9999/api/v2/delete?org=my-org&bucket=mybucket \
  -H 'Authorization: Token <YOURAUTHTOKEN>' \
  -d '{
      "start": "string",
      "stop": "string",
      "predicate": "_measurement=\"someMeasurement\" and tag=\"someTag\""
    }
```

PR #592, opened 2019-11-07
Extract scalar values

### LINKED ISSUE #512: Document extracting a scalar value from results
We need a guide that walks through extracting scalar values from query results. This is done using [stream and table functions](https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/stream-table/). Here's an example:

```js
getFieldValue = (tables=<-, field) => {
  extract = tables
    |> tableFind(fn: (key) => key._field == field)
    |> getColumn(column: "_value")
  return extract[0]
}

n_cpus = from(bucket: "default")
  |> range(start: -1h)
  |> filter(fn: (r) => r._measurement == "system")
  |> getFieldValue(field: "n_cpus")

from(bucket: "default")
  |> range(start: -1h)
  |> filter(fn: (r) => r._measurement == "mem" and r._field == "used_percent")
  |> map(fn: (r) => ({ r with n_cpus: n_cpus }))
```

sanderson:
Closes #512 

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/query-data/guides/scalar-values.md
+++ b/content/v2.0/query-data/guides/scalar-values.md
@@ -1,0 +1,50 @@
+---
+title: Extract scalar values in Flux
+description: >
+  placeholder
+menu:
+  v2_0:
+    name: Extract scalar values
+    parent:  How-to guides
+weight: 210
+v2.0/tags: [scalar]
+---
+
+placeholder
+
+```js
+getFieldValue = (tables=<-, field) => {
+  extract = tables
+    |> tableFind(fn: (key) => key._field == field)
+    |> getColumn(column: "_value")
+  return extract[0]
+}
+
+getRowObject = (tables=<-, field) => {
+  extract = tables
+    |> tableFind(fn: (key) => key._field == field)
+    |> getRecord(idx: 0)
+  return extract
+}
+
+n_cpus = from(bucket: "default")
+  |> range(start: -1h)
+  |> filter(fn: (r) => r._measurement == "system")
+  |> getFieldValue(field: "n_cpus")
+
+row = from(bucket: "default")
+  |> range(start: -1h)
+  |> filter(fn: (r) => r._measurement == "system")
+  |> getRowObject(field: "n_cpus")
+
+from(bucket: "default")
+  |> range(start: -1h)
+  |> filter(fn: (r) => r._measurement == "mem" and r._field == "used_percent")
+  |> keep(columns: ["_time", "_field", "_value"])
+  |> map(fn: (r) => ({ r with
+    n_cpus: n_cpus,
+    rowTime: row._time,
+    rowCPUs: row._value,
+    rowField: row._field
+  }))
+```
--- a/content/v2.0/query-data/guides/scalar-values.md
+++ b/content/v2.0/query-data/guides/scalar-values.md
@@ -10,9 +10,167 @@ weight: 210
 v2.0/tags: [scalar]
 ---
 
-placeholder
+Use Flux [stream and table functions](/v2.0/reference/flux/stdlib/built-in/transformations/stream-table/)
+to extract scalar values from Flux query output for things like dynamically setting
+variables using query results.
+
+**To extract scalar values from output:**
+
+1. [Extract a table](#extract-a-table)
+2. Use the extracted table and [extract an array of column values](#extract-a-column-from-the-table)
+   or [extract a row as an object](#extract-a-row-from-the-table).
+
+_The samples on this page use the [sample data provided below](#sample-data)._
+
+{{% warn %}}
+#### Current limitations
+- The InfluxDB user interface (UI) does not support raw scalar output, but you can
+  use [`map()`](/v2.0/reference/flux/stdlib/built-in/transformations/map/) to add
+  scalar values to data.
+- The [Flux REPL ](/v2.0/reference/cli/influx/repl) does not currently support
+  Flux stream and table functions (also known as "dynamic queries").
+  See [#15321](https://github.com/influxdata/influxdb/issues/15231).
+
+{{% /warn %}}
+
+## Extract a table
+Flux formats query results as a stream of tables.
+To extract a scalar value, first first use
+[`tableFind()`](/v2.0/reference/flux/stdlib/built-in/transformations/stream-table/tablefind/)
+to extract a single table from the stream of tables.
+
+{{% note %}}
+If query results include only one table, it is still formatted as a stream of tables.
+{{% /note %}}
+
+The `tableFind()` extracts the **first** table in a stream of tables whose
+[group key](/v2.0/reference/glossary/#group-key) values match the
+[predicate function](/v2.0/reference/glossary/#predicate-expression) defined by
+the `fn` parameter.
+The predicate function expects a `key` object, which represents the group key of
+each table.
 
 ```js
+sampleData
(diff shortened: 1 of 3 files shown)


content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+menu:
+  v2_0:
+    name: Extract scalar values
+    parent:  How-to guides
+weight: 210
+v2.0/tags: [scalar]
+---
+
+Use Flux [stream and table functions](/v2.0/reference/flux/stdlib/built-in/transformations/stream-table/)
+to extract scalar values from Flux query output for things like dynamically setting
```
kelseiv: ...output. For example, extract...to dynamically set variables...

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+variables using query results.
+
+**To extract scalar values from output:**
+
+1. [Extract a table](#extract-a-table)
```
kelseiv: table).

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+2. Use the extracted table and [extract an array of column values](#extract-a-column-from-the-table)
```
kelseiv: From? the extracted table, [extract...

should we make links and headers match here?

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+   or [extract a row as an object](#extract-a-row-from-the-table).
+
+_The samples on this page use the [sample data provided below](#sample-data)._
+
+{{% warn %}}
+#### Current limitations
+- The InfluxDB user interface (UI) does not support raw scalar output, but you can
```
kelseiv: ...does not currently? support raw scalar output. Use map()...

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+  scalar values to data.
+- The [Flux REPL ](/v2.0/reference/cli/influx/repl) does not currently support
+  Flux stream and table functions (also known as "dynamic queries").
+  See [#15321](https://github.com/influxdata/influxdb/issues/15231).
+
+{{% /warn %}}
+
+## Extract a table
+Flux formats query results as a stream of tables.
+To extract a scalar value, first first use
```
kelseiv: Flux formats...as a stream of tables.
1. Use tableFind() to extract the first table in a stream where the group key matches the predicate function.

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+[`tableFind()`](/v2.0/reference/flux/stdlib/built-in/transformations/stream-table/tablefind/)
+to extract a single table from the stream of tables.
+
+{{% note %}}
+If query results include only one table, it is still formatted as a stream of tables.
```
kelseiv: what's the alternative to one table being formatted as a stream of tables?

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+
+```js
+sampleData
+  |> tableFind(fn: (key) =>
+      key._field == "temp" and
+      key.location == "sfo"
+  )
+```
+
+This returns a single table:
```
kelseiv: The example above returns...

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+| _time                | location | _field | _value |
+|:-----                |:--------:|:------:| ------:|
+| 2019-11-01T12:00:00Z | sfo      | temp   | 65.1   |
+| 2019-11-01T13:00:00Z | sfo      | temp   | 66.2   |
+| 2019-11-01T14:00:00Z | sfo      | temp   | 66.3   |
+| 2019-11-01T15:00:00Z | sfo      | temp   | 66.8   |
+
+{{% note %}}
+#### Minimize the number of input tables
+Flux functions do not guarantee the order in which tables are returned.
```
kelseiv: order tables

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+`tableFind()` simply extracts the first table that matches the `fn` predicate.
```
kelseiv: does tableFind() always return one table? Curious bc otherwise, why would the table order matter?

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+To ensure you retrieve the table that includes the data you actually want, filter
```
kelseiv: tables?

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+// Returns 65.1
+
+SFOTemps[2]
+// Returns 66.3
+```
+
+## Extract a row from the table
+Use the [`getRecord()` function](/v2.0/reference/flux/stdlib/built-in/transformations/stream-table/getrecord/)
+to output data from a single row in the extracted table.
+Specify the index of row to output using the `idx` parameter.
```
kelseiv: of the row

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+
+// Returns {
+//   _time:2019-11-11T12:00:00Z,
+//   _field:"temp",
+//   location:"sfo",
+//   _value: 65.1
+// }
+```
+
+### Use an extracted row object
```
kelseiv: In this example, which variable stores the extracted row object?

content/v2.0/query-data/guides/scalar-values.md
```diff
@@ -0,0 +1,246 @@
...
+
+tempInfo._time
+// Returns 2019-11-11T12:00:00Z
+
+tempInfo.location
+// Returns sfo
+```
+
+## Example helper functions
+Use the process of extracting scalar values from query output to create custom
```
kelseiv: Create custom helper functions to extract scalar values from query output?

content/v2.0/query-data/guides/scalar-values.md
sanderson: Yes, `tableFind()` only ever returns one table – the **first** table it finds that matches the predicate. For example, using the sample data in the guide, you have tables with the following group keys:

```
[location="sfo", _field="temp"]
[location="kjfk", _field="temp"]
[location="kord", _field="temp"]
```

There's no guarantee that the tables will come in that order. So if you're using the predicate:

```
(key) => key._field == "temp"
```

You'll get temp data from one of the three locations, but there's no guarantee as to which one.

I think I'll update this note a bit. You either need to minimize the number of tables going into `tableFind()` or be very specific in your predicate function.

content/v2.0/query-data/guides/scalar-values.md
sanderson: The alternative is pulling the table out of the stream. I've clarified this note a bit.

PR #599, opened 2019-11-12
add EU Frankfurt endpoint

kelseiv:
add `https://eu-central-1-1.aws.cloud2.influxdata.com` to InfluxDB Cloud instance endpoint section

### DIFF
--- a/content/v2.0/cloud/get-started.md
+++ b/content/v2.0/cloud/get-started.md
@@ -87,7 +87,7 @@ For example:
 ```bash
 
 # for US West (Oregon)
-https://us-west-2-1.aws.cloud2.influxdata.com 
+https://us-west-2-1.aws.cloud2.influxdata.com
 
 # for EU Frankfurt
 https://eu-central-1-1.aws.cloud2.influxdata.com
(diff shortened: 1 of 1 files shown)


content/v2.0/cloud/get-started.md
```diff
@@ -84,10 +84,16 @@ When using Telegraf, the InfluxDB v2 API, the `influx` CLI, or the client librar
...
+```bash
+
+# for US West (Oregon)
 https://us-west-2-1.aws.cloud2.influxdata.com
+
+# for EU Frankfurt
+https://eu-central-1-1.aws.cloud2.influxdata.com
 ```
 
+Currently, client libraries are available for the US West (Oregon) region.
```
sanderson: Is this true? I would think users can just set the endpoint URL in the client.

kelseiv: Asked Balaji to confirm. Client libraries tab isn't available for EU region in the UI

sanderson: I just double checked each of the libraries and they do allow users to specify the URL. I'm not sure why they're hiding the client libraries tab in the EU version, but I don't see any reason a user couldn't use a client library to write to InfluxDB in Frankfurt. They simply send data to the InfluxDB API.

kelseiv: Thank you, Scott. Sounds like we need to remove the statement. @bpalani thinks the UI may need to be deployed to EU region cluster and is checking. Will stay tuned.

PR #620, opened 2019-11-20
Flux experimental HTTP package and 'http.get()' function

sanderson:
Closes # 601

Adds the `experimental/http` package with the `http.get()` function.

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/flux/stdlib/experimental/http/_index.md
+++ b/content/v2.0/reference/flux/stdlib/experimental/http/_index.md
@@ -10,7 +10,7 @@ menu:
     name: HTTP
     identifier: HTTP-exp
     parent: Experimental
-weight: 202
+weight: 201
 v2.0/tags: [functions, http, package]
 ---
 
--- a/content/v2.0/reference/flux/stdlib/experimental/http/_index.md
+++ b/content/v2.0/reference/flux/stdlib/experimental/http/_index.md
@@ -1,0 +1,31 @@
+---
+title: Flux HTTP package
+list_title: HTTP package
+description: >
+  The Flux Experimental HTTP package provides experimental functions for transferring
+  data using the HTTP protocol.
+  Import the `experimental/http` package.
+menu:
+  v2_0_ref:
+    name: HTTP
+    identifier: HTTP-exp
+    parent: Experimental
+weight: 202
+v2.0/tags: [functions, http, package]
+---
+
+The Flux Experimental HTTP package provides experimental functions for transferring
+data using the HTTP protocol.
+
+{{% warn %}}
+The experimental HTTP package is subject to change at any time.
+By using this package, you accept the [risks of experimental functions](/v2.0/reference/flux/stdlib/experimental/#use-experimental-functions-at-your-own-risk).
+{{% /warn %}}
+
+Import the `experimental/http` package:
+
+```js
+import "experimental/http"
+```
+
+{{< children type="functions" show="pages" >}}
--- a/content/v2.0/reference/flux/stdlib/experimental/http/_index.md
+++ b/content/v2.0/reference/flux/stdlib/experimental/http/_index.md
@@ -2,8 +2,8 @@
 title: Flux HTTP package
 list_title: HTTP package
 description: >
-  The Flux Experimental HTTP package provides experimental functions for transferring
-  data using the HTTP protocol.
+  The Flux Experimental HTTP package provides functions for transferring data
+  using HTTP protocol.
   Import the `experimental/http` package.
 menu:
   v2_0_ref:
@@ -14,8 +14,8 @@ weight: 201
 v2.0/tags: [functions, http, package]
 ---
 
-The Flux Experimental HTTP package provides experimental functions for transferring
-data using the HTTP protocol.
+The Flux Experimental HTTP package provides functions for transferring data
+using HTTP protocol.
 
 {{% warn %}}
 The experimental HTTP package is subject to change at any time.
--- a/content/v2.0/reference/flux/stdlib/experimental/http/get.md
+++ b/content/v2.0/reference/flux/stdlib/experimental/http/get.md
@@ -2,7 +2,7 @@
 title: http.get() function
 description: >
   The `http.get()` function submits an HTTP GET request to the specified URL with headers
-  and returns the HTTP status code and body as a byte array.
+  and returns the HTTP status code as an integer and the response body as a byte array.
 menu:
   v2_0_ref:
     name: http.get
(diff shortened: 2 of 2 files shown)


content/v2.0/reference/flux/stdlib/experimental/http/get.md
```diff
@@ -0,0 +1,56 @@
...
+##### Get the status of InfluxDB
+```js
+import "experimental/http"
+
+response = http.get(
+    url: "http://localhost.com:9999/health",
+    headers: {}
+  )
+
+status = string(v: response.body)
```
nisley: I think you mean body = string(v: response.body)

You could also add
status = response.statusCode

content/v2.0/reference/flux/stdlib/experimental/http/_index.md
```diff
@@ -0,0 +1,31 @@
...
+menu:
+  v2_0_ref:
+    name: HTTP
+    identifier: HTTP-exp
+    parent: Experimental
+weight: 201
+v2.0/tags: [functions, http, package]
+---
+
+The Flux Experimental HTTP package provides experimental functions for transferring
```
pierwill: Consider: "The Flux Experimental HTTP package provides ~experimental~ functions for transferring"

content/v2.0/reference/flux/stdlib/experimental/http/_index.md
```diff
@@ -0,0 +1,31 @@
+---
+title: Flux HTTP package
+list_title: HTTP package
+description: >
+  The Flux Experimental HTTP package provides experimental functions for transferring
```
pierwill: Consider: "The Flux Experimental HTTP package provides ~experimental~ functions for transferring"

content/v2.0/reference/flux/stdlib/experimental/http/_index.md
```diff
@@ -0,0 +1,31 @@
...
+v2.0/tags: [functions, http, package]
+---
+
+The Flux Experimental HTTP package provides experimental functions for transferring
+data using the HTTP protocol.
+
+{{% warn %}}
+The experimental HTTP package is subject to change at any time.
+By using this package, you accept the [risks of experimental functions](/v2.0/reference/flux/stdlib/experimental/#use-experimental-functions-at-your-own-risk).
+{{% /warn %}}
```
pierwill: We might want to make a standard shortcode warning for experimental packages:

```
{{% warn %}}
This package is experimental and subject to change at any time.
By using this package, you accept the [risks of experimental functions](/v2.0/reference/flux/stdlib/experimental/#use-experimental-functions-at-your-own-risk).
{{% /warn %}}
```

Is there a way to include page titles as variables in those?

content/v2.0/reference/flux/stdlib/experimental/http/get.md
```diff
@@ -0,0 +1,56 @@
...
+The URL to send the GET request to.
+
+_**Data type:** String_
+
+### headers
+Headers to include with the GET request.
+
+_**Data type:** Object_
+
+## Examples
```
pierwill: Can we give an example of how headers are passed? How are multiple headers handled?

content/v2.0/reference/flux/stdlib/experimental/http/_index.md
sanderson: This warning is only used a handful of times and is unique to each pages it's on, so I don't feel it needs a shortcode.

content/v2.0/reference/flux/stdlib/experimental/http/get.md
sanderson: There are examples above and below this (lines 26 and 56), however I've updated the description to say that each header is a key-value pair.

content/v2.0/reference/flux/stdlib/experimental/http/get.md
```diff
@@ -0,0 +1,61 @@
...
+import "experimental/http"
+
+token = secrets.get(key: "READONLY_TOKEN")
+
+response = http.get(
+    url: "http://localhost.com:9999/health",
+    headers: {Authorization: "Token ${token}"}
+  )
+
+status = string(v: response.statusCode)
```
nisley: This can just be:
status = response.statusCode 

The result is status is an int, and I think that's what we want.

content/v2.0/reference/flux/stdlib/experimental/http/get.md
```diff
@@ -0,0 +1,61 @@
...
+```js
+import "influxdata/influxdb/secrets"
+import "experimental/http"
+
+token = secrets.get(key: "READONLY_TOKEN")
+
+response = http.get(
+    url: "http://localhost.com:9999/health",
+    headers: {Authorization: "Token ${token}"}
+  )
```
nisley: Have you actually made post calls with the headers this way?  I'm no expert on calling/using headers and I'm not saying it does not work, but it doesn't look like my working use case.  Here's the example of the types of header use I was testing:

http.get(url: "https://free-nba.p.rapidapi.com/games/44", headers: {"x-rapidapi-host": "free-nba.p.rapidapi.com","x-rapidapi-key": "a64aa239d5..."})

content/v2.0/reference/flux/stdlib/experimental/http/get.md
sanderson: Ok, so only the body is returned as a byte array?

content/v2.0/reference/flux/stdlib/experimental/http/get.md
sanderson: I haven't tested this function specifically (haven't built Flux from source), but it uses the same syntax as `http.post` and I have tested that. Keys inside of objects are always treated as strings, so you shouldn't have to explicitly wrap them in quotes.

content/v2.0/reference/flux/stdlib/experimental/http/get.md
```diff
@@ -0,0 +1,61 @@
+---
+title: http.get() function
+description: >
+  The `http.get()` function submits an HTTP GET request to the specified URL with headers
```
kelseiv: could a request be submitted w/o headers?

content/v2.0/reference/flux/stdlib/experimental/http/get.md
```diff
@@ -0,0 +1,61 @@
...
+## Parameters
+
+### url
+The URL to send the GET request to.
+
+_**Data type:** String_
+
+### headers
+Headers to include with the GET request.
+Headers are represented by key-value pairs.
```
kelseiv: Headers represent key...

PR #634, opened 2019-11-22
AWS marketplace integration

kelseiv:
- Added steps for AWS to "Sign up" section
- Added ref to AWS link for billing
- Added questions about terminology re plans/subscriptions/accounts

### DIFF
--- a/content/v2.0/cloud/account-management/billing.md
+++ b/content/v2.0/cloud/account-management/billing.md
@@ -16,7 +16,7 @@ menu:
 
   If you subscribed to an InfluxDB Cloud 2.0 plan through **AWS Marketplace**, click the **AWS** link to access your billing and subscription information.
   
-  If you subscribed to an InfluxDB Cloud 2.0 plan through **InfluxData**,, complete the following procedures as needed:
+  If you subscribed to an InfluxDB Cloud 2.0 plan through **InfluxData**, complete the following procedures as needed:
 
   - [Add or update your {{< cloud-name >}} payment method](#add-or-update-your-influxdb-cloud-2-0-payment-method)
   - [Add or update your contact information](#add-or-update-your-contact-information)
--- a/content/v2.0/cloud/account-management/billing.md
+++ b/content/v2.0/cloud/account-management/billing.md
@@ -14,7 +14,9 @@ menu:
 
     {{< nav-icon "cloud" >}}
 
-    Complete the following procedures as needed:
+  If you subscribed to a plan through AWS Marketplace, click the **AWS** link to access your billing and subscription information.
+  
+  If you subscribed to an InfluxData plan, complete the following procedures as needed:
 
   - [Add or update your {{< cloud-name >}} payment method](#add-or-update-your-influxdb-cloud-2-0-payment-method)
   - [Add or update your contact information](#add-or-update-your-contact-information)
--- a/content/v2.0/cloud/account-management/billing.md
+++ b/content/v2.0/cloud/account-management/billing.md
@@ -14,9 +14,9 @@ menu:
 
     {{< nav-icon "cloud" >}}
 
-  If you subscribed to a plan through AWS Marketplace, click the **AWS** link to access your billing and subscription information.
+  If you subscribed to an InfluxDB Cloud 2.0 plan through **AWS Marketplace**, click the **AWS** link to access your billing and subscription information.
   
-  If you subscribed to an InfluxData plan, complete the following procedures as needed:
+  If you subscribed to an InfluxDB Cloud 2.0 plan through **InfluxData**,, complete the following procedures as needed:
 
   - [Add or update your {{< cloud-name >}} payment method](#add-or-update-your-influxdb-cloud-2-0-payment-method)
   - [Add or update your contact information](#add-or-update-your-contact-information)
--- a/content/v2.0/cloud/account-management/offboarding.md
+++ b/content/v2.0/cloud/account-management/offboarding.md
@@ -38,7 +38,7 @@ For details, see [Export a dashboard](/v2.0/visualize-data/dashboards/export-das
 
 #### Save Telegraf configurations
 
-1. Click in the **Settings** icon in the navigation bar.
+1. Click the **Settings** icon in the navigation bar.
 
     {{< nav-icon "settings" >}}
 
@@ -52,12 +52,12 @@ To request a backup of data in your {{< cloud-name "short" >}} instance, contact
 
 ### Cancel service
 
-1. Hover over the Usage icon in the left navigation bar and select Billing.
+1. Hover over the **Usage** icon in the left navigation bar and select **Billing**.
 
     {{< nav-icon "usage" >}}
 
 2. Do one of the following:
 
-  - If you subscribed to an InfluxDB Cloud 2.0 **Usage-Based Plan** through **AWS Marketplace**, click the **AWS** link, and then click **Unsubscribe**.
+  - If you subscribed to an InfluxDB Cloud 2.0 **Usage-Based Plan** through **AWS Marketplace**, click the **AWS** link, click **Unsubscribe**, and then click **Yes, cancel subscription**.
   
   - If you subscribed to an InfluxDB Cloud 2.0 **Usage-Based Plan** through **InfluxData**, click **Cancel Service**. Select **I understand and agree to these conditions**, and then click **I understand, Cancel Service.** Click **Confirm and Cancel Service**. Your payment method is charged your final balance immediately upon cancellation of service.
--- a/content/v2.0/cloud/account-management/offboarding.md
+++ b/content/v2.0/cloud/account-management/offboarding.md
@@ -36,9 +36,7 @@ For details, see [Export a task](/v2.0/process-data/manage-tasks/export-task/).
 
 For details, see [Export a dashboard](/v2.0/visualize-data/dashboards/export-dashboard/).
 
-#### Telegraf configurations
-
-**To save a Telegraf configuration:**
+#### Save Telegraf configurations
 
 1. Click in the **Settings** icon in the navigation bar.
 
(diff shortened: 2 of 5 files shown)


content/v2.0/cloud/account-management/billing.md
```diff
@@ -14,7 +14,9 @@ menu:
 
     {{< nav-icon "cloud" >}}
 
-    Complete the following procedures as needed:
+  If you subscribed to a plan through AWS Marketplace, click the **AWS** link to access your billing and subscription information.
```
kelseiv: should we add the Manage Marketplace Subscriptions link in docs? (feels maybe a bit much since it's in UI) https://aws.amazon.com/marketplace/library/

content/v2.0/cloud/get-started.md
```diff
@@ -18,29 +18,34 @@ _See the differences between {{< cloud-name "short">}} and InfluxDB OSS
...
 Once you're ready to grow, [upgrade to the Usage-based Plan](/v2.0/cloud/account-management/upgrade-to-usage-based-plan/).
 
 ## Sign up
 
-1. Go to [InfluxDB Cloud 2.0]({{< cloud-link >}}), enter your email address and password, and click **Sign Up**.
-2. InfluxDB Cloud requires email verification to complete the sign up process. Verify your email address by opening
-   the email sent to the address you provided and clicking **Verify Your Email**.
-3. Choose your cloud provider. Currently, **AWS** is the only cloud provider available.
+1. Choose one of the following:
+   - To subscribe for a plan through InfluxData, go to [InfluxDB Cloud 2.0]({{< cloud-link >}}), enter your credentials, and then click **Sign Up**.
```
kelseiv: Thoughts in general about terminology?
We use “plan” and AWS uses “subscription.” Does it make sense to use the same term consistently for clarity (or keep more descriptive re context)?

content/v2.0/cloud/get-started.md
```diff
@@ -18,29 +18,34 @@ _See the differences between {{< cloud-name "short">}} and InfluxDB OSS
...
+   - To subscribe for a plan through AWS Marketplace, go to [Amazon Marketplace](https://aws.amazon.com/marketplace), search for **InfluxDB Cloud 2.0**, and then follow the prompts to subscribe. After you click **Set Up Your Account**, enter your credentials, and then click **Sign Up**.
+
+       {{%note%}} **Note:** You can sign up for a new InfluxDB Cloud 2.0 account through **either** InfluxData or AWS Marketplace. Currently, we do **not support** using an existing InfluxDB Cloud 2.0 account to sign up for an AWS Marketplace subscription.
```
kelseiv: then we have "account" 
In line 33 note, we could swap "account" with "plan" or "subscription" in the first sentence, but not really in the second sentence.

content/v2.0/cloud/get-started.md
kelseiv: Also, per discussion in meeting today, left button name as "Sign Up"

content/v2.0/cloud/account-management/billing.md
sanderson: I think this note is good enough, but you may want to bold "AWS Marketplace" so it's easy to see when scanning quickly.

content/v2.0/cloud/get-started.md
sanderson: I know there's already been some bikeshedding on this with the marketing team and "subscription" was one of the proposals. Before InfluxDB Cloud 2.0 launched, we decided to go with "Free Plan" and "Pay As You Go Plan" (now "Usage-Based Plan"). I think it's more important to align with the language of the marketing team than with AWS's terminology, but that's only my opinion.

content/v2.0/cloud/get-started.md
sanderson: Accounts are different from plans/subscriptions. A user has a user account and that account uses either the Free Plan or the Usage-Based Plan. I assume the AWS Marketplace subscriptions are just Usage-Based plans?

content/v2.0/cloud/get-started.md
```diff
@@ -18,29 +18,34 @@ _See the differences between {{< cloud-name "short">}} and InfluxDB OSS
...
+        {{%/note%}}
+
+2. InfluxDB Cloud requires email verification to complete the sign up process.
+   Verify your email address by opening the email sent to the address you provided and clicking **Verify Your Email**.
+3. (InfluxData subscriptions only) Choose your cloud provider. Currently, **AWS** is the only cloud provider available.
```
sanderson: I wonder if we even need to call out that AWS is the only provider available here? There are others coming VERY soon.

noramullen1: I think it'd be fine to leave it out given GCP is coming so soon.

content/v2.0/cloud/get-started.md
lukevmorris: I agree with @sanderson -- let's use Marketing terminology.

I do think there's a semantic difference -- we expose a Free Plan and a Usage-Based Plan. And you can subscribe to those plans through either InfluxData (Organic) or AWS (Marketplace).

content/v2.0/cloud/get-started.md
lukevmorris: An AWS Marketplace Subscription is established _for_ an InfluxData Usage-Based plan, within a particular user's account

lukevmorris: I suspect they may feel the same in some cases because there's a 1:1 relationship between Plan and Subscription.

content/v2.0/cloud/account-management/billing.md
barbaranelson: The line below this change is now a bit misleading - it says "If you subscribed to an Influxdata plan ..." Even if they went through the AWS marketplace, they are still subscribing to an Influxdata plan. Maybe we can make it clearer that it's an either/or - either you went through AWS or you subscribed directly with Influx.

content/v2.0/cloud/get-started.md
barbaranelson: "Currently, we do **not support** using an existing InfluxDB Cloud 2.0 account to sign up for an AWS Marketplace subscription" - maybe rephrase to "Currently we do not support using an existing InfluxDB Cloud 2.0 account to sign up for InfluxData via the AWS Marketplace".  I'm not sure that folks will figure out that signing up for an AWS Marketplace subscription is the same as signing up for Influx via the AWS marketplace.

content/v2.0/cloud/get-started.md
kelseiv: @lukevmorris, where will they be able to select Free plan vs Usage-Based Plan in AWS? After hitting the "Set Up Your Account" button?
...........................
- To subscribe to a **Free Plan** or **Usage-Based Plan** through InfluxData/AWS Marketplace
OR
- To subscribe to an **InfluxDB Cloud 2.0 plan** through InfluxData/AWS Marketplace...

content/v2.0/cloud/get-started.md
kelseiv: While the terms differ conceptually, the steps to create an account/plan/subscription (at least from UX point of view) are the same. Reason I bring up--for this note specifically, I wanted to point out I used "account" in the 1st and 2nd sentence, even though we could say "sign up for a new...plan" in the first sentence. Thoughts?

content/v2.0/cloud/get-started.md
kelseiv: sounds good! removed

content/v2.0/cloud/get-started.md
lukevmorris: @kelseiv They won't have a selection -- clicking Subscribe in AWS locks you into a Usage-Based plan.

We may at some point in the future allow users to begin with a free trial or free version. But not on launch.

content/v2.0/cloud/get-started.md
kelseiv: I removed the first sentence in the note, given we're saying as much in Step 1.

lukevmorris:
What did we decide re: documenting the recovery process for a lost session during registration?

kelseiv:
> What did we decide re: documenting the recovery process for a lost session during registration? 

We decided not to document this edge case preemptively. If an issue does occur (and we can detect there is an issue), then redirecting a user to AWS via UI text is optimal. 

 _How to best help customers that abandon and return to AWS registration
Do customers need a heads up for this edge case ahead of time? 
No_

For most docs, we want to limit the following type of statements (because they can be said for most procedures and w/o context aren't very helpful): 
"If you run into an error (or abandon in this case) xyz, refresh your browser and start at the beginning of the procedure").
Exception would be for troubleshooting docs, when we have specific information, for example, if you see error 123, then do xyx (like in troubleshooting docs).

content/v2.0/cloud/account-management/offboarding.md
```diff
@@ -58,6 +56,8 @@ To request a backup of data in your {{< cloud-name "short" >}} instance, contact
...
     {{< nav-icon "usage" >}}
 
-2. Click **Cancel Service**.
-3. Select **I understand and agree to these conditions**, and then click **I understand, Cancel Service.**
-4. Click **Confirm and Cancel Service**. Your payment method is charged your final balance immediately upon cancellation of service.
+2. Do one of the following:
+
+  - If you subscribed to an InfluxDB Cloud 2.0 plan through **AWS Marketplace**, click the **AWS** link, and then click **Unsubscribe**.
+  
+  - If you subscribed to an InfluxDB Cloud 2.0 plan through **InfluxData**, click **Cancel Service**. Select **I understand and agree to these conditions**, and then click **I understand, Cancel Service.** Click **Confirm and Cancel Service**. Your payment method is charged your final balance immediately upon cancellation of service.
```
kelseiv: - @bpalani - did we clean up/minimize this cancel text (I vaguely recall a PR)?

- Will Free Plan folks be looking to cancel their "service" being they're not paying for a subscription? Should we specify "Usage-Based Plan" here? Other thoughts?

lukevmorris: We don't provide a flow for free users to cancel their accounts -- there's really nothing to cancel. I think it's ok to specify `Usage-Based Plan`

If they want to _remove_ their account for GDPR reasons, that's a separate manual process requiring legal. Don't think we should document that one.

PR #676, opened 2019-12-18
Bugfix/kubernetes docs

rawkode:
Fixed the language used to "check" the service

Removed bash style "background" tasking, I don't believe this should be encouraged. Allow the user to keep it front and centre or open a new tab on their own (otherwise we need to explain how to `fg` and kill it?)

- [x] Rebased/mergeable


pierwill:
Thanks, @rawkode!

PR #680, opened 2019-12-20
Remove repeated content on Cloud2 getting started page

pierwill:
Deletes lines that look to me like a repetition of line 127.

### DIFF
--- a/content/v2.0/cloud/get-started.md
+++ b/content/v2.0/cloud/get-started.md
@@ -128,8 +128,6 @@ The primary differences between InfluxDB OSS 2.0 and InfluxDB Cloud 2.0 are:
 - Retrieving data from a file based CSV source using the `file` parameter of the
   [`csv.from()`](/v2.0/reference/flux/functions/csv/from) function is not supported;
   however you can use raw CSV data with the `csv` parameter.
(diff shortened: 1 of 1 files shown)


content/v2.0/cloud/get-started.md
```diff
@@ -128,8 +128,6 @@ The primary differences between InfluxDB OSS 2.0 and InfluxDB Cloud 2.0 are:
 - Retrieving data from a file based CSV source using the `file` parameter of the
   [`csv.from()`](/v2.0/reference/flux/functions/csv/from) function is not supported;
   however you can use raw CSV data with the `csv` parameter.
-- Multi-organization accounts and multi-user organizations are currently not
-  available in {{< cloud-name >}}.
```
sanderson: Where is this repeated?

pierwill: Line 127: `- {{< cloud-name "short" >}} instances are currently limited to a single organization with a single user.`

kelseiv: Nice catch, @pierwill! 
Maybe we could swap "instances" with "accounts"...and/or make singular and positive? An InfluxDB Cloud account currently supports a single....

PR #710, opened 2020-01-09
Update minikube getting started instructions

pierwill:


content/v2.0/get-started.md
```diff
@@ -218,7 +218,14 @@ The instructions below use Minikube, but the steps should be similar in any Kube
     kubectl apply -f https://raw.githubusercontent.com/influxdata/docs-v2/master/static/downloads/influxdb-k8-minikube.yaml
     ```
 
-    (**Note:** Always inspect YAML manifests before running `kubectl apply -f <url>`!)
+    <div class="warn block">
+      <p>
+        Always inspect YAML manifests before running <code>kubectl apply -f &lt;url&gt;</code>!
+      </p>
+    </div>
```
sanderson: I'm guessing you did this because the "nested" shortcode was breaking the layout. Shortcodes create a new "context" that don't know whether or not they're nested. They honor all indentation as if it was at the root level, so to get around it, you keep the shortcode tags indented, but all the contents of the shortcode should not be indented:

```
    {{% warn %}}
Always inspect YAML manifests before running `kubectl apply -f <url>`.
    {{% /warn %}}
```

pierwill: For me this methods breaks both the indentation and subsequent paragaph.
Screen Shot 2020-01-09 at 2 53 08 PM

sanderson: Ok, yeah, I've tested it too. Definitely a bug with Hugo's markdown/shortcode renderer. I guess the raw HTML works for now until we can figure out the issue with the shortcode.

PR #712, opened 2020-01-14
Differentiate 'join' and 'union'

### LINKED ISSUE #557: Differentiate `join()` and `union()`
There's some confusion about when one should use `join()` vs `union()`. We should update the function docs to explain the difference. Maybe even show the difference visually.

##### Relevant URLs
- https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/join/
- https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/union/

sanderson:
Closes #557 

This outlines the difference between `join()` and `union()` and includes an illustration to show the difference.

image

- [x] Rebased/mergeable


content/v2.0/reference/flux/stdlib/built-in/transformations/join.md
```diff
@@ -136,3 +138,14 @@ join(
   on: ["_time", "host"]
 )
 ```
+
+## join() versus union()
+`join()` outputs new rows comprised of data from each of the joined streams of tables.
+`union()` does not modify the data in each row, but unifies separate streams of tables
```
kelseiv: join() inputs and combines rows f specified tables and outputs all rows in one table?
union() inputs specified tables w/o combining rows and outputs tables in union?

PR #765, opened 2020-02-18
Template environment references

sanderson:
Updates the template docs with information about using environment references.

### DIFF
--- a/content/v2.0/influxdb-templates/create.md
+++ b/content/v2.0/influxdb-templates/create.md
@@ -114,12 +114,16 @@ influx pkg export all \
 ## Include user-definable resource names
 To let users customize resource names when installing your template, use
 **environment references** in place of names.
-Environment references are replaced with user-defined values when the template is installed.
 
-In your template manifest, replace a [supported resource field](#supported-resource-fields)
+In your template manifest, replace a [supported resource field](#resource-fields-that-support-environment-references)
 with an `envRef` object.
 A `envRef` object contains a single `key` property.
-`key` is a string that references a user-defined environment reference key.
+`key` is a string that references the key of a key-value pair a user should
+provide when installing the template.
+During installation, the `envRef` object is replaced by the value of the
+referenced key-value pair.
+If the user does not provide the environment reference key-value pair, InfluxDB
+uses the `key` string as the default value.
 
 {{< code-tabs-wrapper >}}
 {{% code-tabs %}}
@@ -163,16 +167,13 @@ influx pkg \
   --env-ref=bucket-name-1=myBucket
 ```
 
-If the user does not provide the environment reference key-value pair, InfluxDB
-uses the `key` string as the default value.
-
 {{% note %}}
-#### Supported resource fields
+#### Resource fields that support environment references
 Only the following fields support environment references:
 
 - `metadata.name`
 - `associations[].name`
-- `endpointName` _(unique to the `NotificationRule` resource)_
+- `endpointName` _(unique to `NotificationRule` resources)_
 {{% /note %}}
 
 ## Share your InfluxDB templates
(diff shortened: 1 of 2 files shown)


content/v2.0/influxdb-templates/create.md
```diff
@@ -111,6 +111,70 @@ influx pkg export all \
...
+If the user does not provide the environment reference key-value pair, InfluxDB
+uses the `key` string as the default value.
+
+{{% note %}}
+#### Supported resource fields
+Only the following fields support environment references:
+
+- `metadata.name`
+- `associations[].name`
+- `endpointName`
```
jsteenb2: it may be worth mentioning, `endpointName` is only valid for the `NotificationRule` resource. Your call, its the only place that key shows up atm.

sanderson: Updated

content/v2.0/influxdb-templates/create.md
```diff
@@ -111,6 +111,70 @@ influx pkg export all \
   --telegraf-configs=00000x0x000X0x0X0
 ```
 
+## Include user-definable resource names
```
kelseiv: when can this task happen? Perhaps link(s) in overview section where it makes sense...

content/v2.0/influxdb-templates/create.md
```diff
@@ -111,6 +111,70 @@ influx pkg export all \
...
+To let users customize resource names when installing your template, use
```
kelseiv: replace resource names with environment references.

content/v2.0/influxdb-templates/create.md
```diff
@@ -111,6 +111,70 @@ influx pkg export all \
...
+**environment references** in place of names.
+Environment references are replaced with user-defined values when the template is installed.
```
kelseiv: drop line 117

content/v2.0/influxdb-templates/create.md
```diff
@@ -111,6 +111,70 @@ influx pkg export all \
...
+
+In your template manifest, replace a [supported resource field](#supported-resource-fields)
```
kelseiv: 1. How to open/access template manifest...
2. Search for one of the following resource fields to update:
    - metadata.name
    - associations[].name
    - endpointName 
with an `envRef` object that contains a `key` property to reference the user-defined...

content/v2.0/influxdb-templates/create.md
```diff
@@ -111,6 +111,70 @@ influx pkg export all \
...
+when [installing the template](/v2.0/influxdb-templates/use/#install-templates) to
+set the bucket name to "myBucket".
+
+```sh
+influx pkg \
+  -f /path/to/template.yml \
+  --env-ref=bucket-name-1=myBucket
+```
+
+If the user does not provide the environment reference key-value pair, InfluxDB
```
kelseiv: 4. Enter a `key` string value to use if a user doesn't provide...

Scott, is this same `key` value referenced in step 3 above?

content/v2.0/influxdb-templates/create.md
```diff
@@ -111,6 +111,70 @@ influx pkg export all \
...
+uses the `key` string as the default value.
+
+{{% note %}}
+#### Supported resource fields
```
kelseiv: maybe drop this section or move to top before procedure?

content/v2.0/influxdb-templates/use.md
```diff
@@ -161,6 +162,21 @@ influx pkg \
   --recurse
 ```
 
+### Define environment references
+Some templates include [environment references](/v2.0/influxdb-templates/create/#include-user-definable-resource-names) that let you provide custom names for resources in the template.
+Template maintainers should document what environment references exist in the template
```
kelseiv: If you use environment refs, document the env refs available in your template and list the keys...

content/v2.0/influxdb-templates/create.md
sanderson: Yes, it's the same.

PR #773, opened 2020-02-21
Count rows in empty tables

### LINKED ISSUE #749: Update 'count()' transformation
We need to update the `count()` documentation to explain that count will return 0 for empty tables, but to include empty tables in your stream of tables, you need to explicitly keep/create them with filtering and windowing data.

```js
filter(onEmpty: "keep")
window(createEmpty: true)
```

##### Relevant URLs
- https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/aggregates/count/

sanderson:
Closes #749 

Adds information about preserving empty tables and using `count()`.

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/flux/stdlib/built-in/transformations/aggregates/count.md
+++ b/content/v2.0/reference/flux/stdlib/built-in/transformations/aggregates/count.md
@@ -21,6 +21,16 @@ _**Output data type:** Integer_
 count(column: "_value")
 ```
 
+{{% note %}}
+#### Count empty tables
+`count()` will return `0` for empty tables, but to include empty tables in your data,
+you must set the following parameters for specific functions:
+
+- [`filter(onEmpty: "keep")`](/v2.0/reference/flux/stdlib/built-in/transformations/filter/#onempty)
+- [`window(createEmpty: true)`](/v2.0/reference/flux/stdlib/built-in/transformations/window/#createempty)
+- [`aggregateWindow(createEmpty: true)`](/v2.0/reference/flux/stdlib/built-in/transformations/aggregates/aggregatewindow/#createempty)
+{{% /note %}}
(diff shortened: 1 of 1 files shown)


content/v2.0/reference/flux/stdlib/built-in/transformations/aggregates/count.md
```diff
@@ -21,6 +21,16 @@ _**Output data type:** Integer_
 count(column: "_value")
 ```
 
+{{% note %}}
+#### Count empty tables
+`count()` will return `0` for empty tables, but to include empty tables in your data,
+you must set the following parameters for specific functions:
```
kelseiv: `count()` returns `0` for an empty table. 
To include (column and row headers?) for an empty table, set the following parameters for (which?) specific functions:

content/v2.0/reference/flux/stdlib/built-in/transformations/aggregates/count.md
```diff
@@ -21,6 +21,18 @@ _**Output data type:** Integer_
...
 ```
 
+{{% note %}}
+#### Count empty tables
+`count()` returns `0` for empty tables.
+To keep empty tables in your data, set the following parameters for the following functions:
+
+| Function                                                                                              | Parameter           |
+|:--------                                                                                              |:---------           |
+| [filter()](/v2.0/reference/flux/stdlib/built-in/transformations/filter/)                              | `onEmpty: "keep"`   |
```
jsternberg: Can you add a warning here that keeping empty data on your first filter can have pretty severe performance costs? If the person wants to keep tables from a filter, they should do their normal filters first and then do whatever filter where they want to keep the empty tables as a second call to filter.

PR #809, opened 2020-03-10
Add menu links to client library repos

### LINKED ISSUE #808: Client Lib docs... menu lists only 1 lib
_Describe the issue here._

##### Relevant URLs
- https://v2.docs.influxdata.com/v2.0/reference/api/client-libraries/
Python lib appears in the nav menu...but the others do not.
Screen Shot 2020-03-10 at 9 41 25 AM

Current list of our Client libs can be seen here:
Screen Shot 2020-03-10 at 9 41 29 AM

pierwill:
Closes #808.

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/api/client-libraries/csharp-cl.md
+++ b/content/v2.0/reference/api/client-libraries/csharp-cl.md
@@ -1,0 +1,9 @@
+---
+title: InfluxDB C# client library
+description: Use the C# client library to interact with InfluxDB.
+menu:
+  v2_0_ref:
+    name: C#
+    parent: Client libraries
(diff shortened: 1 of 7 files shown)


content/v2.0/reference/api/client-libraries/csharp-cl.md
```diff
@@ -0,0 +1,9 @@
+---
+title: InfluxDB C# client library
+description: Use the C# client library to interact with InfluxDB.
+menu:
+  v2_0_ref:
+    name: C# client library
```
sanderson: This is nested under client libraries in the left nav, so "client library" here is redundant. I'd just just use the language as the menu name across all of the client libraries.

PR #813, opened 2020-03-10
Geo package strict filtering explanation

### LINKED ISSUE #802: Explain strict when describing geo.strictFilter() function
_Describe the issue here._

On the page:

https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/experimental/geo/strictfilter/

... the term "strict" is used, but not defined. 

Please explain that strict filters out points that might have bled outside the boundary, due to the way that the s2 geometry library works. It'd also be good to provide some visual examples of strict on, and strict off.

##### Relevant URLs
- _Provide relevant URLs_

https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/experimental/geo/strictfilter/

sanderson:
Closes #802 

Explains the difference between strict and non-strict filtering in the geo package with some visual representations:

image

- [x] Rebased/mergeable


content/v2.0/reference/flux/stdlib/experimental/geo/filterrows.md
```diff
@@ -35,6 +35,27 @@ geo.filterRows(
 )
 ```
 
+### Strict and non-strict filtering
+In most cases, the specified geographic region does not perfectly align with S2 grid cells.
+**Non-strict filtering** returns points that may be outside of the specified region but
```
kelseiv: maybe start w Strict filtering sentence?

- **Strict filtering** returns points inside the specified (filter?) region.
- **Non-strict filtering** returns points inside S2 grid cells covered by the specified region.

....not sure about adding "filter" in text, maybe extra? Either way, match text to graphic

PR #864, opened 2020-03-25
Note MySQL plugin supports MariaDB and Percona

kelseiv:
Scott, updated this page: https://v2.docs.influxdata.com/v2.0/reference/telegraf-plugins/

Should we note anywhere else in v2 docs?

### DIFF
--- a/data/telegraf_plugins.yml
+++ b/data/telegraf_plugins.yml
@@ -890,7 +890,7 @@ input:
   - name: MySQL
     id: mysql
     description: |
-      The MySQL input plugin gathers the statistics data from MySQL servers.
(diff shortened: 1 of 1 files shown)


data/telegraf_plugins.yml
```diff
@@ -890,7 +890,7 @@ input:
   - name: MySQL
     id: mysql
     description: |
-      The MySQL input plugin gathers the statistics data from MySQL servers.
+      The MySQL input plugin gathers the statistics data from MySQL servers. Input plugin supports MariaDB and Percona.
```
sanderson: What about:

The MySQL input plugin gathers the statistics data from MySQL, MariaDB, and Percona servers.

PR #910, opened 2020-04-02
Flux Pushbullet package

### LINKED ISSUE #909: Document Flux Pushbullet package
PR: https://github.com/influxdata/flux/pull/2697

sanderson:
Closes #909

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/flux/stdlib/pagerduty/endpoint.md
+++ b/content/v2.0/reference/flux/stdlib/pagerduty/endpoint.md
@@ -48,7 +48,6 @@ The returned object must include the following fields:
 - `client`
 - `client_url`
 - `class`
-- `dedupKey`
 - `eventAction`
 - `group`
 - `severity`
@@ -74,22 +73,18 @@ crit_statuses = from(bucket: "example-bucket")
   |> filter(fn: (r) => r._measurement == "statuses" and status == "crit")
 
 crit_statuses
-  |> e(mapFn: (r) => {
-      obj = mapFn(r: r)      
-      return {r with _sent: string(v: 2 == (sendEvent(pagerdutyURL: url,
-        routingKey: obj.routingKey,
-        client: obj.client,
-        clientURL: obj.clientURL,
-        dedupKey: r._pagerdutyDedupKey,
-        class: obj.class,
-        eventAction: obj.eventAction,
-        group: obj.group,
-        severity: obj.severity,
-        component: obj.component,
-        source: obj.source,
-        summary: obj.summary,
-        timestamp: obj.timestamp,
-      ) / 100))}
+  |> e(mapFn: (r) => ({ r with
+      routingKey: r.routingKey,
+      client: r.client,
+      clientURL: r.clientURL,
+      class: r.class,
+      eventAction: r.eventAction,
+      group: r.group,
+      severity: r.severity,
+      component: r.component,
+      source: r.source,
+      summary: r.summary,
+      timestamp: r._time,
     })
-  })()
+  )()
 ```
--- a/content/v2.0/reference/flux/stdlib/pagerduty/endpoint.md
+++ b/content/v2.0/reference/flux/stdlib/pagerduty/endpoint.md
@@ -32,14 +32,13 @@ Defaults to `https://events.pagerduty.com/v2/enqueue`.
 
 _**Data type:** String_
 
+## Usage
+`pagerduty.endpoint` is a factory function that outputs another function.
+The output function requires a `mapFn` parameter.
+
 ### mapFn
 A function that builds the object used to generate the POST request.
-
-{{% note %}}
-_You should rarely need to override the default `mapFn` parameter.
-To see the default `mapFn` value or for insight into possible overrides, view the
-[`pagerduty.endpoint()` source code](https://github.com/influxdata/flux/blob/master/stdlib/pagerduty/pagerduty.flux)._
-{{% /note %}}
+Requires an `r` parameter.
 
 _**Data type:** Function_
 
@@ -49,6 +48,8 @@ The returned object must include the following fields:
 - `client`
 - `client_url`
 - `class`
+- `dedupKey`
+- `eventAction`
 - `group`
 - `severity`
 - `component`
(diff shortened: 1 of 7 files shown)


content/v2.0/reference/flux/stdlib/pushbullet/pushdata.md
```diff
@@ -0,0 +1,79 @@
...
+_**Function type:** Output_
+
+```js
+import "pushbullet"
+
+pushbullet.pushData(
+  url: "https://api.pushbullet.com/v2/pushes",
+  token: "",
+  data: {
+    "type": "link",
```
kelseiv, line 23: Does Pushbullet support data types other than link and note?

affo: yep, I think it does: https://docs.pushbullet.com/#create-push

sanderson: Yes, it does. This is just an example of data being sent to the Pushbullet API. All the different options are documented in their API documentation. I'll add a link.

sanderson:
Thanks @affo, not sure how I missed `mapFn`. I've added it in.

content/v2.0/reference/flux/stdlib/pushbullet/endpoint.md
```diff
@@ -20,7 +20,12 @@ import "pushbullet"
 
 pushbullet.endpoint(
   url: "https://api.pushbullet.com/v2/pushes",
-  token: ""
+  token: "",
+  mapFn: (r) => ({
```
affo, line 24: oh, I am sorry it's not like that:

```
e = p.endpoint(token)

d
  |> e(mapFn: (r) =>
  ({r with
    title: "hello from REPL",
    text: "Value: ${r._value}"}))()
```

Th endpoint is a function, that given a `mapFn` returns a transformation that can be used when you pipe in data 👍

sanderson: @affo I'm not sure I understand. This specific codeblock is just meant to show the function parameters and example values. The example below this shows data being piped in. Are the examples incorrect?

affo: Yep.

`pushbullet.endpoint()` does not have a `mapFn` parameter.

It returns a function that has a `mapFn` param, as I show in the example above.

That's actually the same for pagerduty/slack.

I think it should be ok to do the same as for them in docs 👍 

I see now it is kind of wrong also there:

 - https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/slack/endpoint/
 - https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/pagerduty/endpoint/

If you look at theFlux code for them (pgduty for example:

```
endpoint = (url=defaultURL) =>
    (mapFn) =>
        (tables=<-) => tables
            |> dedupKey()
            |> map(fn: (r) => {
                obj = mapFn(r: r)
                return {r with _sent: string(v: 2 == (sendEvent(pagerdutyURL: url,
                    routingKey: obj.routingKey,
                   .......
                ) / 100))}
            })
```

You can see that `endpoint` only takes `url`.

It then returns a function that accepts `mapFn`, that returns a transformation (or a function that accepts streams of tables).

affo: The comments above `endpoint` are pretty explanatory:

```
// `endpoint` creates the endpoint for the PushBullet external service.
// It will push notifications of type `note`.
// If you want to push something else, see `genericEndpoint`.
// `url` - string - URL of the PushBullet endpoint. Defaults to: "https://api.pushbullet.com/v2/pushes".
// `token` - string - token for the PushBullet endpoint.
// The returned factory function accepts a `mapFn` parameter.
// The `mapFn` must return an object with `title` and `text` fields as defined in the `pushNote` function arguments.
```

I think you could use those 🤔

affo:
Clearer snippet:

```
import p "pushbullet"

data = from(...) |> range |> ...

factoryFn = p.endpoint(url: ..., token: ...)
transformation = factoryFn(mapFn: (r) => ...)

data |> transformation()
```

content/v2.0/reference/flux/stdlib/pagerduty/endpoint.md
```diff
@@ -62,14 +63,33 @@ _For more information, see [`pagerduty.sendEvent()`](/v2.0/reference/flux/stdlib
...
+        eventAction: obj.eventAction,
+        group: obj.group,
+        severity: obj.severity,
+        component: obj.component,
+        source: obj.source,
+        summary: obj.summary,
+        timestamp: obj.timestamp,
+      ) / 100))}
+    })
+  })()
```
affo, line 94: I think this is from the pagerduty package.

The correct could would not include 

```
obj = mapFn(r: r)      
      return {r with _sent: string(v: 2 == (sendEvent(pagerdutyURL: url,
```

but something like (from docs: `// The `mapFn` parameter must be a function that returns an object with `routingKey`, `client`, `client_url`, `class`, `group`, `severity`, `eventAction`, `component`, `source`, `summary`, and `timestamp` as defined in the sendEvent function.
`:

```
crit_statuses
  |> e(mapFn: (r) => ({ r with
        routingKey: r.routingKey,
        client: r.client,
        clientURL: r.clientURL,
        class: r.class,
        eventAction: r.eventAction,
        group: r.group,
        severity: r.severity,
        component: r.component,
        source: r.source,
        summary: r.summary,
        timestamp: r._time,
      })
 )()
```

Maybe the values could be hardcoded by the user, but that's the sense 👍

affo: This holds for also for slack and pushbullet.

content/v2.0/reference/flux/stdlib/pagerduty/endpoint.md
```diff
@@ -49,6 +48,8 @@ The returned object must include the following fields:
 - `client`
 - `client_url`
 - `class`
+- `dedupKey`
```
affo, line 51: This is not true:

```
// Note that while sendEvent accepts a dedup key, endpoint gets the dedupkey from the groupkey of the input table instead of it being handled by the `mapFn`.
```

content/v2.0/reference/flux/stdlib/pushbullet/endpoint.md
```diff
@@ -0,0 +1,87 @@
...
+
+_**Function type:** Output_
+
+```js
+import "pushbullet"
+
+pushbullet.endpoint(
+  url: "https://api.pushbullet.com/v2/pushes",
+  token: "",
+  mapFn: (r) => ({
```
affo, line 24: This should be removed 👍

affo:
@sanderson thanks for doing this! Just added a nit for something you forgot to update I think. 🙏

content/v2.0/reference/flux/stdlib/pushbullet/endpoint.md
affo: I mean, `mapFn` shouldn't be in the call to `endpoint`

sanderson: Nice catch @affo! Thanks! Fixed.

PR #912, opened 2020-04-03
Add 'epsilon' and 'verbose' parameters to 'testing.diff'

### LINKED ISSUE #911: Document 'epsilon' parameter for 'testing.diff' function
The `epsilon` parameter defines the margin of error allowed when comparing floating point values.

##### Relevant URLs
- https://github.com/influxdata/flux/blob/master/stdlib/testing/diff.go#L20-L26

sanderson:
Closes #911 

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/flux/stdlib/testing/diff.md
+++ b/content/v2.0/reference/flux/stdlib/testing/diff.md
@@ -51,6 +51,7 @@ _**Data type:** Object_
 
 ### verbose
 Enable verbose output.
+Defaults to `true`.
 
 _**Data type:** Boolean_
 
--- a/content/v2.0/reference/flux/stdlib/testing/diff.md
+++ b/content/v2.0/reference/flux/stdlib/testing/diff.md
@@ -17,7 +17,12 @@ _**Function type:** Test_
 ```js
 import "testing"
 
-testing.diff(got: stream2, want: stream1)
+testing.diff(
+  got: stream2,
+  want: stream1,
+  verbose: true,
+  epsilon: 0.000000001
+)
 ```
 
 It matches tables from each stream with the same group keys.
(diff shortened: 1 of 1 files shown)


content/v2.0/reference/flux/stdlib/testing/diff.md
```diff
@@ -34,16 +39,27 @@ _The `testing.diff()` function can be used to perform in-line diffs in a query._
...
 
 _**Data type:** Object_
 
 ### want
-The stream that contains the expected data to test against.
+Stream that contains the expected data to test against.
 
 _**Data type:** Object_
 
+### verbose
```
kelseiv, line 52: default true?

sanderson: I'm not sure 😄... @yzhang1991 ?

yzhang1991: Actually I don't know about `verbose`
From the code, it does not seem to be used anywhere. I think it is dead code.

sanderson: Nm, default is `true`. Added it to the description.

yzhang1991: Default is true, but we did not use it in the Process() code @nathanielc

sanderson: @yzhang1991 if it's not used, should I remove it or leave it there?

yzhang1991: @sanderson I suggest you remove it. Setting `verbose` does not change the function behavior in any way.

content/v2.0/reference/flux/stdlib/testing/diff.md
```diff
@@ -34,16 +38,22 @@ _The `testing.diff()` function can be used to perform in-line diffs in a query._
...
 
 ### want
-The stream that contains the expected data to test against.
+Stream that contains the expected data to test against.
 
 _**Data type:** Object_
 
+### epsilon
+Specifies how far apart two **float** values can be, but still considered equal.
+Defaults to `0.000000001`.
```
yzhang1991, line 53: I would recommend this to become something like `1e-9`, hard to count those zeros.

sanderson: I agree, but Flux doesn't support float values using scientific notation and I don't want to confuse users.

yzhang1991: I see, thanks!

yzhang1991: I created https://github.com/influxdata/flux/issues/2732, hoping that we could have scientific notation supported one day.

PR #946, opened 2020-04-13
Fill query guide

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/query-data/flux/fill.md
+++ b/content/v2.0/query-data/flux/fill.md
@@ -1,7 +1,7 @@
 ---
-title: Fill gaps in data
-seotitle: Fill gaps in data
-list_title: Fill gaps
+title: Fill null values in data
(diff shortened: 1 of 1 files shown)


content/v2.0/query-data/flux/fill.md
```diff
@@ -0,0 +1,117 @@
...
+v2.0/tags: [query, fill]
+list_code_example: |
+  ```js
+  data
+    |> fill(usePrevious: true)
+  ```
+---
+
+Use the [`fill()` function](/v2.0/reference/flux/stdlib/built-in/transformations/fill/)
+to replace _null_ values in query results.
```
kelseiv, line 22: Use the [`fill()` function](/v2.0/reference/flux/stdlib/built-in/transformations/fill/)
to replace _null_ values in query results with:
- Previous non-null values
- Specified value

PR #986, opened 2020-04-28
Calculating percentages in Flux

### LINKED ISSUE #981: Flux example: Calculate a % Utilization
A really common use case across a variety of monitoring scenarios is the calculation of a utilization %.  The query documentation is either too generic or too specific with the details of the available functions and doesn't provide enough examples for common query scenarios where many functions need to be used together to solve a problem.

Calculating a % utilization is a really common scenario that requires applying a number of functions together to achieve the desired result; typically a gauge for an immediate indicator or a line graph if you are visualizing the data over time.

I'm also sure there are a number of way to solve for this kind of use case -- selecting individual series into separate variables then joining, etc.  But, we should provide what we believe is the most efficient way of achieving this result.

#### Example
I'm currently looking at the GPU metrics provided by Nvidia's System Management Interface tooling.  It provides access to the various metrics about the GPUs and Telegraf has a handy plugin to pull that data into InfluxDB.

Two of the metrics that it reports are `memory_total` and `memory_used` (it also provides `memory_free`).  In order to calculate the % memory used vs. the total, memory_used is divided by memory_total and then multiplied by 100.

In some other use cases, you won't have the total.  You may only have the used and the free...which you will need to sum to infer the total.

It would be good to explore this example -- one path resulting in a gauge related visualization.
The other outcome would be a line graphed over time.

sanderson:
Closes #981

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/query-data/flux/calculate-percentages.md
+++ b/content/v2.0/query-data/flux/calculate-percentages.md
@@ -132,7 +132,7 @@ Use `map()` to re-map values in each row.
 Divide `mem_used` by `mem_total` and multiply by 100 to return the percentage.
 
 {{% note %}}
-To return a precise float percentage value that includes decimal points the example
+To return a precise float percentage value that includes decimal points, the example
 below casts integer field values to floats and multiplies by a float value (`100.0`).
 {{% /note %}}
 
@@ -177,8 +177,8 @@ Both `pivot()` and `join()` will do this, but there are important differences be
 
 #### Pivot is more performant
 `pivot()` only has to read and operate on a single stream of data.
-`join()` requires two streams of data and the overhead of reading and combing both
-streams can be high, especially for larger data sets.
+`join()` requires two streams of data and the overhead of reading and combining
+both streams can be significant, especially for larger data sets.
 
 #### Use join for multiple data sources
 `join()` is really only necessary when querying data from multiple buckets or
@@ -208,8 +208,8 @@ from(bucket: "example-bucket")
 
 1. Ensure measurements are in the same [bucket](/v2.0/reference/glossary/#bucket).
 2. Use `filter()` to include data from both measurements.
-3. Use `group()` to ungroup data to return a single table.
-4. Use `pivot()` to pivot fields into row-wise sets.
+3. Use `group()` to ungroup data and return a single table.
+4. Use `pivot()` to pivot fields into columns.
 5. Use `map()` to re-map rows and perform the percentage calculation.
 
 <!-- -->
--- a/content/v2.0/query-data/flux/calculate-percentages.md
+++ b/content/v2.0/query-data/flux/calculate-percentages.md
@@ -35,7 +35,7 @@ Then, use `map()` to re-map values in the row and calculate a percentage.
    to divide the numerator operand value by the denominator operand value and multiply by 100.
 
 {{% note %}}
-In the following examples use `pivot()` to align operands into rows because
+The following examples use `pivot()` to align operands into rows because
 `pivot()` works in most cases and is more performant than `join()`.
 _See [Pivot vs join](/v2.0/query-data/flux/mathematic-operations/#pivot-vs-join)._
 {{% /note %}}
--- a/content/v2.0/query-data/flux/calculate-percentages.md
+++ b/content/v2.0/query-data/flux/calculate-percentages.md
@@ -228,11 +228,16 @@ from(bucket: "example-bucket")
 #### Use join for multiple data sources
 ```js
 import "sql"
+import "influxdata/influxdb/secrets"
+
+pgUser = secrets.get(key: "POSTGRES_USER")
+pgPass = secrets.get(key: "POSTGRES_PASSWORD")
+pgHost = secrets.get(key: "POSTGRES_HOST")
 
 t1 = sql.from(
   driverName: "postgres",
-  dataSourceName: "postgresql://user:password@localhost",
-  query:"SELECT * FROM exampleTable"
+  dataSourceName: "postgresql://${pgUser}:${pgPass}@${pgHost}",
+  query:"SELECT id, name, available FROM exampleTable"
 )
 
 t2 = from(bucket: "example-bucket")
--- a/content/v2.0/query-data/flux/calculate-percentages.md
+++ b/content/v2.0/query-data/flux/calculate-percentages.md
@@ -1,0 +1,247 @@
+---
+title: Calculate percentages with Flux
+list_title: Calculate percentages
+description: >
+  Use [`pivot()` or `join()`](/v2.0/query-data/flux/calculate-percentages/#pivot-vs-join)
+  and the [`map()` function](/v2.0/reference/flux/stdlib/built-in/transformations/map/)
+  to align operand values into row-wise sets and calculate a percentage.
+menu:
+  v2_0:
+    name: Calculate percentages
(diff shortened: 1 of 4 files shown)


content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+{{< /code-tabs-wrapper >}}
+
+{{% note %}}
+This guide uses `pivot()` to align operand values into row-wise sets because
+`pivot()` works for the majority of use cases and is more performant than `join()`.
+_See [Pivot vs join](#pivot-vs-join) below._
+{{% /note %}}
+
+## GPU monitoring example
+The following example queries data collected from a GPU monitoring solution and
```
kelseiv, line 72: ...queries data from the `gpu-monitor` bucket and calculates...

content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+
+---
+
+## Pivot vs join
+To use values in mathematical operations in Flux, operand values must exists in a single row.
+Both `pivot()` and `join()` will do this, but there are important differences between the two:
+
+#### Pivot is more performant
+`pivot()` only has to read and operate on a single stream of data.
+`join()` requires two streams of data and the overhead of reading and combing both
```
kelseiv: join() reads and operates on two streams of data, requiring more overhead, especially

content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+---
+
+## Pivot vs join
+To use values in mathematical operations in Flux, operand values must exists in a single row.
+Both `pivot()` and `join()` will do this, but there are important differences between the two:
+
+#### Pivot is more performant
+`pivot()` only has to read and operate on a single stream of data.
```
kelseiv, line 179: pivot() reads and operates...

content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+`join()` requires two streams of data and the overhead of reading and combing both
+streams can be high, especially for larger data sets.
+
+#### Use join for multiple data sources
+`join()` is really only necessary when querying data from multiple buckets or
```
kelseiv, line 184: Use `join()` when querying data from different buckets or data sources.

content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+{{% /note %}}
+
+## GPU monitoring example
+The following example queries data collected from a GPU monitoring solution and
+calculates the percentage of GPU memory used over time.
+Data includes the following:
+
+- **`gpu` measurement**
+- **`mem_used` field**: used GPU memory in bytes
+- **`mem_total` field**: total GPU memory in bytes
```
kelseiv, line 78: Reword as numbered steps to indicate required order?

content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+    _time: r._time,
+    _measurement: r._measurement,
+    _field: "mem_used_percent",
+    _value: float(v: r.mem_used) / float(v: r.mem_total) * 100.0
+  }))
+```
+
+---
+
+## Pivot vs join
```
kelseiv, line 174: How many use cases for pivot vs join? If it isn't specific to "calculating percentages", I'd move to another topic and link to.

content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+
+#### Use join for multiple data sources
+`join()` is really only necessary when querying data from multiple buckets or
+different data sources.
+
+---
+
+## Examples
+
+#### Use pivot() to align multiple fields
```
kelseiv, line 191: Reframe example headings to the "Calculating percentages" perspective:
- Use pivot() to align multiple fields  >> Calculate a percentage using multiple fields
- Use pivot() for math across measurements >> Calculate a percentage using multiple measurements
- Use join() for multiple data sources >> Calculate a percentage from multiple data sources

content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+**To calculate percentages:**
+
+1. Query the necessary operand values.
+2. Use [`pivot()` or `join()`](#pivot-vs-join) to align operand values into row-wise sets.
+3. Use [`map()`](/v2.0/reference/flux/stdlib/built-in/transformations/map/)
+   to divide the numerator operand value by the denominator operand value and multiply by 100.
+
+{{< code-tabs-wrapper >}}
+{{% code-tabs %}}
+[pivot()](#)
```
kelseiv, line 37: It might be nice to include all examples up here at the top, and then reword to focus on user's data (rather than pivot or join): 

### Examples of calculating percentages:
-  from multiple data sources
- from two buckets
- from two measurements
- from 2 fields in one measurement
- from 3 fields in one measurement

content/v2.0/query-data/flux/calculate-percentages.md
sanderson: These aren't ordered steps. These are just points describing the schema.

content/v2.0/query-data/flux/calculate-percentages.md
```diff
@@ -0,0 +1,247 @@
...
+## GPU monitoring example
+The following example queries data collected from a GPU monitoring solution and
+calculates the percentage of GPU memory used over time.
+Data includes the following:
+
+- **`gpu` measurement**
+- **`mem_used` field**: used GPU memory in bytes
+- **`mem_total` field**: total GPU memory in bytes
+
+### Query mem_used and mem_total fields
```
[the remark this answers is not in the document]
sanderson: I'm not a fan of using inline code in headings. It never formats well and calls too much attention to itself.

content/v2.0/query-data/flux/calculate-percentages.md
sanderson: That's a good point. This should probably live in the more general "Transform data with math" doc.

content/v2.0/query-data/flux/calculate-percentages.md
sanderson: I move the note about using pivot vs join before this code block and simplified it to just use point (no code tabs)

content/v2.0/query-data/flux/calculate-percentages.md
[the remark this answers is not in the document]
kelseiv: I agree! It's like that throughout the v1 docs, but will stop doing that and establish this rule as our standard!

content/v2.0/query-data/flux/calculate-percentages.md
kelseiv: Ah, sorry, my comment wasn't too clear...my thought was it may be helpful to indicate order (think you did that above in first procedure), like:
1. Specify `gpu-monitor` bucket.
...
3. Use `filter()` to specify the `gpu` measurement and two fields to query: `mem_used` and `mem_total `

Issue #985, opened 2020-04-28
Link to token CLI instructions in GS CLI docs


noramullen1:
Per slack convo with @samhld and @timhallinflux:
Add info for [viewing a token](https://v2.docs.influxdata.com/v2.0/security/tokens/view-tokens/) to [getting started via CLI instructions](https://v2.docs.influxdata.com/v2.0/get-started/#set-up-influxdb-through-the-influx-cli).

noramullen1:
- Token gets generated but not displayed during first time setup
- Need to let people know the above and how they can view it from both the CLI and UI

noramullen1:
Closed via #993

PR #993, opened 2020-04-29
Add info re. token creation during setup and link to viewing tokens doc

noramullen1:
Addresses #985 

- Explain that tokens are created during setup
- Direct users on how to view token via CLI or UI

### DIFF
--- a/content/v2.0/get-started.md
+++ b/content/v2.0/get-started.md
@@ -385,14 +385,13 @@ To use the [`influx` CLI](/v2.0/reference/cli/influx) after setting up InfluxDB 
 
 Use one of the following methods to provide your authentication token to the CLI:
 
-1. Pass your token to the `influx` CLI using the `-t` or `--token` flag.
-2. Set the `INFLUX_TOKEN` environment variable using your token.
+- Create a new InfluxDB connection configuration using the [`influx config create` command](/v2.0/reference/cli/influx/config/create/).
+- Pass your token to the `influx` CLI using the `-t` or `--token` flag.
+- Set the `INFLUX_TOKEN` environment variable using your token.
 
     ```bash
     export INFLUX_TOKEN=oOooYourAuthTokenOoooOoOO==
     ```
-3. Store your token in `~/.influxdbv2/credentials`.
-   _The content of the `credentials` file should be only your token._
 
 _See [View tokens](/v2.0/security/tokens/view-tokens/) for information about
 retrieving authentication tokens._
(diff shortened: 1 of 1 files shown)


content/v2.0/get-started.md
```diff
@@ -381,8 +381,9 @@ You are ready to [write or collect data](/v2.0/write-data).
 
 {{% note %}}
 #### Using the influx CLI after setting up InfluxDB through the UI
-To use the [`influx` CLI](/v2.0/reference/cli/influx) after setting up InfluxDB through the UI,
-use one of the following methods to provide your [authentication token](/v2.0/users/tokens/) to the CLI:
+To use the [`influx` CLI](/v2.0/reference/cli/influx) after setting up InfluxDB through the UI, provide your [authentication token](/v2.0/users/tokens/), which is automatically generated during the setup process. For instructions on viewing your token via CLI or UI, see [View tokens](/v2.0/security/tokens/view-tokens/).
+
+Use one of the following methods to provide your authentication token to the CLI:
```
russorat, line 386: I think the preferred way to do this is to use the `influx config create` command. the -t option is available, but setting up the config is the better way.

timhallinflux: I think the original issue was....what if I use the CLI to perform the initial setup?  

So I think this section: https://v2.docs.influxdata.com/v2.0/get-started/#set-up-influxdb-through-the-influx-cli

needs to have a step 8... or state something about Before proceeding to use InfluxDB further via the CLI, you need a token.  Then...how do I get that?  You have to log in via the UI with the credentials you just established....once....to get a token to use for other CLI related action.

russorat: makes sense to me.

Specifically for this PR though, option 3 about the credentials file should be removed, and a new option 1 should be inserted describing how to set up a new influx cli profile with your credentials using the `influx config create` command.

content/v2.0/get-started.md
```diff
@@ -485,9 +488,6 @@ The primary differences between InfluxDB OSS 2.0 and InfluxDB Cloud 2.0 are:
 - [InfluxDB scrapers](/v2.0/write-data/scrape-data/) that collect data from specified
   targets are not available in {{< cloud-name "short" >}}.
 - {{< cloud-name "short" >}} instances are currently limited to a single organization with a single user.
-- **InfluxDB Cloud** does not support retrieving data from a file based CSV source
-  using the `file` parameter of the [`csv.from()`](/v2.0/reference/flux/functions/csv/from);
-  however you can use raw CSV data with the `csv` parameter.
```
russorat, line 490: OSS cannot read from a file either, so i removed this

content/v2.0/get-started.md
```diff
@@ -485,9 +488,6 @@ The primary differences between InfluxDB OSS 2.0 and InfluxDB Cloud 2.0 are:
 - [InfluxDB scrapers](/v2.0/write-data/scrape-data/) that collect data from specified
   targets are not available in {{< cloud-name "short" >}}.
 - {{< cloud-name "short" >}} instances are currently limited to a single organization with a single user.
```
russorat, line 490: this is no longer true

Issue #983, opened 2020-04-27
Document how to link to dashboards with variables pre-set in url


russorat:
if you use variables in a dashboard, those values are now available to be set in the URL so that you can link to dashboards with variables already set. 

For a dashboard url, you can add `&vars[variable_name]=value` and that value will be set when you get to the dashboard. Urls copied from the browser also store those variables in them so you can share the link easily.

PR #995, opened 2020-04-29
Mention variables in dashboard URLs

noramullen1:
Addresses #983 

@sanderson , wasn't sure if there was an additional/better place to include this, let me know if you have any thoughts.


russorat:
@sanderson yes. for example, since tables support url linking, you can build tables with deeplinks based off of query results.

sanderson:
@russorat That's awesome! @noramullen1, we need to call this out here. This is a fairly significant feature.

PR #1048, opened 2020-05-18
Add SQL Server support to SQL docs

### LINKED ISSUE #1047: Add SQL Server support to Flux 'sql' package
PR: https://github.com/influxdata/flux/pull/2847

##### Relevant URLs
- https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/sql/from/
- https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/sql/to/

sanderson:
Closes #1047

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/query-data/flux/sql.md
+++ b/content/v2.0/query-data/flux/sql.md
@@ -116,12 +116,17 @@ sql.from(
 
 {{% code-tab-content %}}
 ```js
+import "sql"
+
 sql.from(
   driverName: "sqlserver",
   dataSourceName: "sqlserver://user:password@localhost:1234?database=examplebdb",
   query: "GO SELECT * FROM Example.Table"
 )
 ```
+
+_For information about authenticating with SQL Server using ADO-style parameters,
+see [SQL Server ADO authentication](/v2.0/reference/flux/stdlib/sql/from/#sql-server-ado-authentication)._
 {{% /code-tab-content %}}
 {{< /code-tabs-wrapper >}}
 
--- a/content/v2.0/query-data/flux/sql.md
+++ b/content/v2.0/query-data/flux/sql.md
@@ -5,7 +5,7 @@ list_title: Query SQL data
 description: >
   The Flux `sql` package provides functions for working with SQL data sources.
   Use `sql.from()` to query SQL databases like PostgreSQL, MySQL, Snowflake,
-  SQLite, and MS SQL Server.
+  SQLite, and Microsoft SQL Server.
 v2.0/tags: [query, flux, sql]
 menu:
   v2_0:
@@ -32,7 +32,7 @@ The [Flux](/v2.0/reference/flux) `sql` package provides functions for working wi
 [`sql.from()`](/v2.0/reference/flux/stdlib/sql/from/) lets you query SQL data sources
 like [PostgreSQL](https://www.postgresql.org/), [MySQL](https://www.mysql.com/),
 [Snowflake](https://www.snowflake.com/), [SQLite](https://www.sqlite.org/index.html),
-and [MS SQL Server](https://www.microsoft.com/en-us/sql-server/default.aspx),
+and [Microsoft SQL Server](https://www.microsoft.com/en-us/sql-server/default.aspx),
 and use the results with InfluxDB dashboards, tasks, and other operations.
 
 - [Query a SQL data source](#query-a-sql-data-source)
--- a/content/v2.0/query-data/flux/sql.md
+++ b/content/v2.0/query-data/flux/sql.md
@@ -4,7 +4,8 @@ seotitle: Query SQL data sources with InfluxDB
 list_title: Query SQL data
 description: >
   The Flux `sql` package provides functions for working with SQL data sources.
-  Use `sql.from()` to query SQL databases like PostgreSQL, MySQL, Snowflake, and SQLite.
+  Use `sql.from()` to query SQL databases like PostgreSQL, MySQL, Snowflake,
+  SQLite, and MS SQL Server.
 v2.0/tags: [query, flux, sql]
 menu:
   v2_0:
@@ -30,7 +31,8 @@ list_code_example: |
 The [Flux](/v2.0/reference/flux) `sql` package provides functions for working with SQL data sources.
 [`sql.from()`](/v2.0/reference/flux/stdlib/sql/from/) lets you query SQL data sources
 like [PostgreSQL](https://www.postgresql.org/), [MySQL](https://www.mysql.com/),
-[Snowflake](https://www.snowflake.com/), and [SQLite](https://www.sqlite.org/index.html),
+[Snowflake](https://www.snowflake.com/), [SQLite](https://www.sqlite.org/index.html),
+and [MS SQL Server](https://www.microsoft.com/en-us/sql-server/default.aspx),
 and use the results with InfluxDB dashboards, tasks, and other operations.
 
 - [Query a SQL data source](#query-a-sql-data-source)
@@ -57,6 +59,7 @@ To query a SQL data source:
 [MySQL](#)
 [Snowflake](#)
 [SQLite](#)
+[SQL Server](#)
 {{% /code-tabs %}}
 
 {{% code-tab-content %}}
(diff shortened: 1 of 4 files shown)


alespour:
@sanderson 

> Flux doesn't provide access to system environment variables. How and where should these environment variables be set?

These environment variables could be set by a user/admin on a system where Flux REPL or InfluxDB runs. I'm honestly not sure whether this is suitable, it is just one of a few ways that Azure supports to get access token. 

> InfluxDB restricts access to the filesystem, so users wouldn't be able to use this method to access credentials if querying from InfluxDB. The only place this would really work is in the Flux REPL, unless this references a file stored in server SQL server is running on. Is this how it's supposed to be?

It references a file on the filesystem where Flux REPL or InfluxDB runs. So this looks like not a viable option. Similarly, connection string may contain `azure certificate path` param which value is expected to point to local filesystem.

---

I could remove support for connection string options that leads to local filesystem, ie. `azure auth` with filename value, and `azure certificate path` , which would render that code that handles these options inactive (but keep it in the source, just in case). As for Azure authentication using environment variables, I do not know. 

Isn't there a way to distinguish whether Flux runs in REPL or InfluxDB? In REPL, all options could be supported.

What do you suggest? Thank you.

sanderson:
@alespour 

I think we should leave out the environment variable method. For security reasons, Flux cannot read environment variables, whether in the REPL or in InfluxDB (can you confirm @nathanielc). I think we can and should leave the file-based options in since the restriction isn't Flux, it's InfluxDB. I'll just document that the file-based option is only available in the REPL like we do [here](https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/csv/from/#file) with `csv.from()`.

alespour:
@sanderson 

I guess I have not explained what `azure auth=ENV` properly, let me re-iterate. The environment variables are not read at Flux code but in native code that handles opening connection to db.

When user specifies such option in connection string in Flux `sql.from` or `sql.to` function, the code (added in this PR)  calls `auth.GetSettingsFromEnvironment` from Azure Go SDK in order to get access token. That method just calls `os.Getenv` and attempts to read a few variables. Similarly, it works when `azure auth` param that points to a local file. I've tested `azure auth=ENV` option and it works.

sanderson:
@alespour Thank you for the explanation! That makes perfect sense. I'll include notes about authenticating with Environment variables not being supported in InfluxDB Cloud and that file based authentication in the Flux REPL.

content/v2.0/query-data/flux/sql.md
```diff
@@ -4,7 +4,8 @@ seotitle: Query SQL data sources with InfluxDB
 list_title: Query SQL data
 description: >
   The Flux `sql` package provides functions for working with SQL data sources.
-  Use `sql.from()` to query SQL databases like PostgreSQL, MySQL, Snowflake, and SQLite.
+  Use `sql.from()` to query SQL databases like PostgreSQL, MySQL, Snowflake,
+  SQLite, and MS SQL Server.
```
kelseiv, line 8: Microsoft SQL Server

content/v2.0/query-data/flux/sql.md
```diff
@@ -30,7 +31,8 @@ list_code_example: |
 The [Flux](/v2.0/reference/flux) `sql` package provides functions for working with SQL data sources.
 [`sql.from()`](/v2.0/reference/flux/stdlib/sql/from/) lets you query SQL data sources
 like [PostgreSQL](https://www.postgresql.org/), [MySQL](https://www.mysql.com/),
-[Snowflake](https://www.snowflake.com/), and [SQLite](https://www.sqlite.org/index.html),
+[Snowflake](https://www.snowflake.com/), [SQLite](https://www.sqlite.org/index.html),
+and [MS SQL Server](https://www.microsoft.com/en-us/sql-server/default.aspx),
```
kelseiv, line 35: SQL Server

content/v2.0/reference/flux/stdlib/sql/from.md
```diff
@@ -139,3 +146,52 @@ sql.from(
...
+
+sql.from(
+  driverName: "sqlserver",
+  dataSourceName: "sqlserver://${username}:${password}@localhost:1234?database=examplebdb",
+  query: "GO SELECT * FROM Example.Table"
+)
+```
+
+#### SQL Server ADO authentication
+Use one of the the following methods to provide SQL Server authentication credentials as
```
kelseiv, line 166: double the

content/v2.0/reference/flux/stdlib/sql/from.md
```diff
@@ -139,3 +146,52 @@ sql.from(
...
+access to the underlying file system and do not support reading credentials from a file.
+To retrieve SQL Server credentials from a file, execute the query in the
+[Flux REPL](/v2.0/reference/cli/influx/repl/) on your local machine.
+{{% /warn %}}
+
+```
+azure auth=c:\secure\azure.auth
+```
+
+##### Specify authentication credentials in the connection string
```
kelseiv, line 187: is this example of how to specify auth credentials?

content/v2.0/reference/flux/stdlib/sql/from.md
```diff
@@ -139,3 +146,52 @@ sql.from(
...
+```
+azure tenant id=77...;azure client id=58...;azure client secret=0cf123..
+azure tenant id=77...;azure client id=58...;azure certificate path=C:\secure\...;azure certificate password=xY...
+azure tenant id=77...;azure client id=58...;azure username=some@myorg;azure password=a1...
+```
+
+##### Use a Managed identity in an Azure VM
```
kelseiv, line 194: is there more content for using a managed identity?

content/v2.0/reference/flux/stdlib/sql/to.md
```diff
@@ -60,6 +61,12 @@ username[:password]@hostname:port/dbname/schemaname?account=<your_account>&param
 
 # SQLite Driver DSN
 file:/path/to/test.db?cache=shared&mode=rw
+
+# MS SQL Server Driver DSNs
```
kelseiv, line 65: first occurrence

content/v2.0/reference/flux/stdlib/sql/from.md
```diff
@@ -61,6 +62,12 @@ username[:password]@hostname:port/dbname/schemaname?account=<your_account>&param
 
 # SQLite Driver DSN
 file:/path/to/test.db?cache=shared&mode=ro
+
+# MS SQL Server Driver DSNs
```
kelseiv, line 66: Microsoft SQL Server on first occurrence

content/v2.0/reference/flux/stdlib/sql/to.md
```diff
@@ -148,3 +155,52 @@ sql.to(
...
+
+sql.to(
+  driverName: "sqlserver",
+  dataSourceName: "sqlserver://${username}:${password}@localhost:1234?database=examplebdb",
+  table: "Example.Table"
+)
+```
+
+#### SQL Server ADO authentication
+Use one of the the following methods to provide SQL Server authentication credentials as
```
kelseiv, line 175: double

content/v2.0/reference/flux/stdlib/sql/from.md
```diff
@@ -139,3 +146,52 @@ sql.from(
...
+  driverName: "sqlserver",
+  dataSourceName: "sqlserver://${username}:${password}@localhost:1234?database=examplebdb",
+  query: "GO SELECT * FROM Example.Table"
+)
+```
+
+#### SQL Server ADO authentication
+Use one of the the following methods to provide SQL Server authentication credentials as
+[ActiveX Data Objects (ADO)](https://docs.microsoft.com/en-us/sql/ado/guide/ado-introduction?view=sql-server-ver15)
+connection string parameters:
```
kelseiv, line 168: list options here in bullet list and link to headers

content/v2.0/reference/flux/stdlib/sql/to.md
```diff
@@ -148,3 +155,52 @@ sql.to(
...
+[ActiveX Data Objects (ADO)](https://docs.microsoft.com/en-us/sql/ado/guide/ado-introduction?view=sql-server-ver15)
+connection string parameters:
```
kelseiv, line 177: same as above

content/v2.0/reference/flux/stdlib/sql/from.md
sanderson: Yes, I added comments to the code block to explain.

content/v2.0/reference/flux/stdlib/sql/from.md
sanderson: I added a link to the Microsoft documentation.

PR #1051, opened 2020-05-20
Update table structure page

### LINKED ISSUE #997: InfluxDb Table Structure -- Misleading Info
On the URl below the docs state:
[https://v2.docs.influxdata.com/v2.0/reference/key-concepts/table-structure/](https://v2.docs.influxdata.com/v2.0/reference/key-concepts/table-structure/)

> InfluxDB 2.0 uses the following columnar table structure to store data: 

This is potentially misleading as the DB actually stores data in a TSM file.  Furthermore Influx does not really have "tables".

Rather we may want the docs to state that this is for [annnotated csv](https://v2.docs.influxdata.com/v2.0/reference/syntax/annotated-csv/#tables), and is how data is presented in a response; not how it is actually stored in the DB

This is sort of nuanced but an important delineation.

kelseiv:
Closes #997

- Rename file table-structure.md > data-schema.md 
- Update content to:
   - Clarify data is stored in TSM and TSI files
   - Specify what the tabular data schema is used for

@sanderson, I think you're reworking the CSV content? We can prob reorg/drop some duplicate info between "Data schema" page and "Annotated CSV" wherever that lands.


content/v2.0/reference/key-concepts/data-schema.md
```diff
@@ -0,0 +1,61 @@
+---
+title: InfluxDB data schema
+description: >
+  InfluxDB uses a tabular data schema for displaying raw data in Data Explorer and for returning query results in annotated CSV syntax.
+weight: 103
+menu:
+  v2_0_ref:
+    parent: Key concepts
+    name: Data schema
+v2.0/tags: [key concepts]
```
sanderson, line 10: Add the alias for the old file name:

```yml
aliases:
  - /v2.0/reference/key-concepts/table-structure
```

Issue #1006, opened 2020-05-01
Document how to use the Flux REPL


sanderson:
Currently we only document how to start and end the REPL. We need to document using it.

- [ ] Setting variables and how they behave (they're immutable)
- [ ] Returning values (don't need yields, press `return`)
    - Mention what `yield()` does in the REPL (provides a name in the result column)
- [ ] Single-line vs multiple lines
- etc.

pierwill:
Closed in https://github.com/influxdata/docs-v2/pull/1355.

PR #1055, opened 2020-05-21
Add note on immutable variables in REPL

pierwill:
Small change toward addressing #1006.


pierwill:
Ah! Makes sense to me.

Issue #1077, opened 2020-05-29
More details for our Cloud 2 locations


alsargent:
We have customers who want to know where our Cloud 2 on Azure instance is. Currently it states "West Europe". 

![image](https://user-images.githubusercontent.com/20701/83312086-14a25180-a1c6-11ea-8d94-12381f833cb0.png)

But that's not helpful for determining compliance requirements. For instance, Germany is in western Europe, but has data privacy laws that go beyond EU regulations:

https://www.complianceweek.com/data-privacy/germanys-dual-approach-to-data-regulation-under-the-gdpr/28386.article

--

Other issues:

#2

Admittedly nitpicky, but we say "Frankfurt" and not "Frankfurt, Germany". I deal with enough customer emails to not be surprised if someone might scan this and think it's Frankfurt Kentucky/Alabama/Illinois/Indiana/Kansas/Maine/etc.

![image](https://user-images.githubusercontent.com/20701/83311927-99d93680-a1c5-11ea-89f2-b971f6529e7e.png)

#3 

We don't explicitly name which regions we're in. We state "US West", not "US West 2" for instance. At volume, I worry about someone saying that our doc only said "US West", and claim they purchased thinking we were in Northern California (AWS US West 1 region). 

Yes, these would be silly objections, but it'd burn up cycles needlessly defending against. It the history of business disputes, there have been sillier claims made.

##### Relevant URLs
https://v2.docs.influxdata.com/v2.0/reference/urls/

##### Proposed solution

Be explicit about where we are running. In the **Region** column, state the exact region that each of the cloud providers use the following, respectively:

us-west-1
eu-central-1
us-central1
westeurope-1

Create a new column, **Location**, with the city (if well-known), state (if well-known) and country. So use the following, respectively:

Oregon, USA
Frankfurt, Germany
Iowa, USA
Amsterdam, Netherlands

noramullen1:
Hi @alsargent , thanks for this issue.

We can definitely include the country for Frankfurt and add the city/country for the Azure data center. 

But not sure what you're asking for re. # 3 above:

> We don't explicitly name which regions we're in. We state "US West", not "US West 2" for instance. At volume, I worry about someone saying that our doc only said "US West", and claim they purchased thinking we were in Northern California (AWS US West 1 region).

The screenshot you took very clearly says Oregon? I'm not sure if you're referring to a different place in the docs or you somehow didn't see that. Can you clarify?

PR #1080, opened 2020-06-01
Data center locations

noramullen1:


layouts/shortcodes/cloud_regions.html
```diff
@@ -13,11 +13,13 @@ <h3 id="{{ $titleID }}">{{ $title }}</h3>
     <table>
       <thead>
         <th align="left">Region</th>
+        <th align="left">Location</th>
         <th align="left">URL</th>
       </thead>
       {{ range .regions }}
         <tr>
           <td {{ if .status }}class="{{ .status }}"{{ end }}>{{ .name }}</td>
+          <td><a href="{{ .location }}">{{ .location }}</a></td>
```
sanderson, line 22: This doesn't need to be a link. Update it to:

```
<td>{{ .location }}</td>
```

data/influxdb_urls.yml
```diff
@@ -12,18 +12,22 @@ cloud:
     - name: Amazon Web Services
       short_name: AWS
       regions:
-        - name: US West (Oregon)
```
sanderson, line 15: US West (Oregon) is the actual name of the AWS region (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-sample-templates-us-west-2.html) and is what we display in the UI. Same goes for all of the region names in this yml file.

noramullen1:
Addresses #1077

PR #1105, opened 2020-06-16
Add steps to download and install CLI

### LINKED ISSUE #1102: Document CLI-only download option
We now have a separate 2.0 download option that is only the `influx` CLI. This is ideal for cloud users since they only need the `influx` CLI to interact with their cloud instance. They do not need `influxd`.

##### Relevant URLs
- https://v2.docs.influxdata.com/v2.0/get-started/
-  https://v2.docs.influxdata.com/v2.0/influxdb-templates/cloud/

kelseiv:
closes #1102

### DIFF
--- a/content/v2.0/get-started.md
+++ b/content/v2.0/get-started.md
@@ -90,14 +90,7 @@ References to the InfluxDB user interface (UI) or localhost:9999 refer to your
 
 ### (Optional) Download and install the influx CLI
 
-You can write, query, and process data through the Cloud user interface (UI) or  `influx` CLI. To download and install the `influx` CLI, complete the following steps:
-
-1. Click the **macOS** or **Linux** button below, and then click the appropriate **influx CLI** button to download the CLI binary.
-2. Do one of the following:
-   - For **macOS**: [Unpackage the binary](#unpackage-the-influx-binary), and then [(optional) place the binary in your $PATH](#optional-place-the-binary-in-your-path)
-   - For **Linux**: [Place the binary in your $PATH](#place-the-binary-in-your-path)
-3. If you're downloading on macOS Catalina, complete the steps in [Run influx CLI on macOS Catalina](/v2.0/get-started/#run-influx-cli-on-macos-catalina).
-4. (Optional) To install `influx` shell completion scripts, see [`influx completion`](/v2.0/reference/cli/influx/completion/#install-completion-scripts).
+To use the `influx` CLI to manage and interact with your InfluxDB Cloud instance, complete the following steps:
 
 {{< tabs-wrapper >}}
 {{% tabs %}}
@@ -155,6 +148,8 @@ To manually authorize the binary:
 We are in the process of updating our build process to ensure released binaries are signed by InfluxData.
 {{% /warn %}}
 
+Now, you're ready to [Use the influx CLI](#use-the-influx-cli).
+
 {{% /tab-content %}}
 <!--------------------------------- END macOS --------------------------------->
 
@@ -186,6 +181,8 @@ sudo cp influxdb_client_2.0.0-beta.12_linux_amd64/influx /usr/local/bin/
 If you rename the binary, all references to `influx` in this documentation refer to the renamed binary.
 {{% /note %}}
 
+Now, you're ready to [Use the influx CLI](#use-the-influx-cli).
+
 {{% /tab-content %}}
 <!--------------------------------- END Linux --------------------------------->
 
@@ -198,6 +195,7 @@ Sign in to [InfluxDB Cloud 2.0](https://cloud2.influxdata.com) using your email 
 <a class="btn" href="https://cloud2.influxdata.com">Sign in to InfluxDB Cloud 2.0 now</a>
 
 ### Start working with your time series data
+
 With {{< cloud-name "short" >}} setup, see [Next steps](#next-steps) for what to do next.
 
 ---
@@ -218,11 +216,13 @@ executables, and running the initial setup process.
 <!-------------------------------- BEGIN macOS -------------------------------->
 {{% tab-content %}}
 ### Download and install InfluxDB v2.0 beta
+
 Download InfluxDB v2.0 beta for macOS.
 
 <a class="btn download" href="https://dl.influxdata.com/influxdb/releases/influxdb_2.0.0-beta.12_darwin_amd64.tar.gz" download>InfluxDB v2.0 beta (macOS)</a>
 
 ### Unpackage the InfluxDB binaries
+
 To unpackage the downloaded archive, **double click the archive file in Finder**
 or run the following command in a macOS command prompt application such
 **Terminal** or **[iTerm2](https://www.iterm2.com/)**:
@@ -233,6 +233,7 @@ tar zxvf ~/Downloads/influxdb_2.0.0-beta.12_darwin_amd64.tar.gz
 ```
 
 #### (Optional) Place the binaries in your $PATH
+
 If you choose, you can place `influx` and `influxd` in your `$PATH` or you can
 prefix the executables with `./` to run then in place.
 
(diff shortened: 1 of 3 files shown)


content/v2.0/influxdb-templates/cloud.md
```diff
@@ -15,12 +15,12 @@ v2.0/tags: [templates]
...
-If using InfluxDB Cloud, download and use the [`influx` command line interface (CLI)](/v2.0/reference/cli/influx/)
+Use the [`influx` command line interface (CLI)](/v2.0/reference/cli/influx/)
 to apply and manage templates in your InfluxDB Cloud account.
 
-InfluxDB OSS 2.0 includes the InfluxDB CLI (`influx`). If you haven’t already, do the following:
+If you haven’t already, do the following:
 
-1. [Download and install InfluxDB 2.0 OSS](/v2.0/get-started/#start-with-influxdb-oss).
+1. [Download and install `influx` CLI](/v2.0/get-started/#optional-download-and-install-the-influx-cli).
 2. [Configure the `influx` CLI](/v2.0/get-started/#set-up-influxdb) to use your
```
kelseiv, line 24: Think step 2 goes away?

content/v2.0/get-started.md
```diff
@@ -88,6 +88,109 @@ References to the InfluxDB user interface (UI) or localhost:9999 refer to your
...
+#### (Optional) Place the binary in your $PATH
+
+If you choose, you can place `influx` in your `$PATH` or you can
+prefix the executable with `./` to run in place.
+
+```sh
+# (Optional) Copy the influx binary to your $PATH
+sudo cp influxdb_client_2.0.0-beta.12_darwin_amd64/influx /usr/local/bin/
+```
+
```
kelseiv, line 137: Curious why steps unpackaging macOS and Linux binaries (and placing binary in $PATH) aren't more parallel. Is it just that Linux binaries must be placed in $PATH and for macOS it's optional?

content/v2.0/influxdb-templates/cloud.md
kelseiv, line 24: does step 2 change with the separate influx install? When I look at this procedure, it seems more for InfluxDB setup than influx CLI setup

content/v2.0/reference/release-notes/influxdb.md
```diff
@@ -12,6 +12,7 @@ weight: 101
 
 ### Features
 
+- Add option for Cloud users to use the `influx` CLI to interact with a Cloud instance. To download the CLI, [open the InfluxData Download page](https://portal.influxdata.com/downloads/), click **v2.0.0-beta**, and then run the command under **your OS (CLI Only)**. For information about using the `influx` CLI, see [influx - InfluxDB command line interface](/v2.0/reference/cli/influx/).
```
kelseiv, line 15: actually, now that I think about it, this should probably link to get-started instructions with for instructions on placing binary in SPATH, the Catalina security issue, and influx completion steps

content/v2.0/get-started.md
```diff
@@ -88,6 +88,109 @@ References to the InfluxDB user interface (UI) or localhost:9999 refer to your
...
+You can write, query, and process data through the Cloud user interface (UI) or  `influx` CLI. To download and install the `influx` CLI, complete the following steps:
+
+1. Click the **macOS** or **Linux** button below, and then click the appropriate **influx CLI** button to download the CLI binary.
+2. Do one of the following:
+   - For **macOS**: [Unpackage the binary](#unpackage-the-influx-binary), and then [(optional) place the binary in your $PATH](#optional-place-the-binary-in-your-path)
+   - For **Linux**: [Place the binary in your $PATH](#place-the-binary-in-your-path)
+3. If you're downloading on macOS Catalina, complete the steps in [Run influx CLI on macOS Catalina](/v2.0/get-started/#run-influx-cli-on-macos-catalina).
+4. (Optional) To install `influx` shell completion scripts, see [`influx completion`](/v2.0/reference/cli/influx/completion/#install-completion-scripts).
+
```
kelseiv, line 101: Do we need to say anything about how to configure the `influx` CLI?

content/v2.0/get-started.md
```diff
@@ -88,6 +88,109 @@ References to the InfluxDB user interface (UI) or localhost:9999 refer to your
 {{< cloud-name >}} UI.
 {{% /cloud %}}
 
+### (Optional) Download and install the influx CLI
+
+You can write, query, and process data through the Cloud user interface (UI) or  `influx` CLI. To download and install the `influx` CLI, complete the following steps:
```
sanderson, line 93: To use the `influx` CLI to manage and interact with your InfluxDB Cloud instance, complete the following steps:

content/v2.0/get-started.md
```diff
@@ -88,6 +88,109 @@ References to the InfluxDB user interface (UI) or localhost:9999 refer to your
...
+
+1. Click the **macOS** or **Linux** button below, and then click the appropriate **influx CLI** button to download the CLI binary.
+2. Do one of the following:
+   - For **macOS**: [Unpackage the binary](#unpackage-the-influx-binary), and then [(optional) place the binary in your $PATH](#optional-place-the-binary-in-your-path)
+   - For **Linux**: [Place the binary in your $PATH](#place-the-binary-in-your-path)
+3. If you're downloading on macOS Catalina, complete the steps in [Run influx CLI on macOS Catalina](/v2.0/get-started/#run-influx-cli-on-macos-catalina).
+4. (Optional) To install `influx` shell completion scripts, see [`influx completion`](/v2.0/reference/cli/influx/completion/#install-completion-scripts).
```
sanderson, line 100: This structure feels weird to me. Should we just nest the OS-specific content under each OS tab? I don't know that they need to be told to click the button. I think that can safely be assumed.

content/v2.0/get-started.md
sanderson: We probably could make these align more closely. Placing them in the $PATH is optional for either OS. It just changes how you execute the CLI. If it's on your $PATH, you can run `influx` from any directory and it will work. If it's not in your $PATH, you have to specify the location of the CLI (`./influx` or `path/to/influx`, etc.). The issue is if they have the 1.x binary on their $PATH, moving the 2.0 to the $PATH will overwrite it since they have the same name.

content/v2.0/get-started.md
```diff
@@ -88,6 +88,109 @@ References to the InfluxDB user interface (UI) or localhost:9999 refer to your
...
+Currently, when you first attempt to run `influx`, macOS prevents it from running.
+To manually authorize the binary:
+
+1. Attempt to run the `influx`.
+2. Open **System Preferences** and click **Security & Privacy**.
+3. Under the **General** tab, there is a message about `influx` being blocked.
+   Click **Open Anyway**.
+
+We are in the process of updating our build process to ensure released binaries are signed by InfluxData.
+{{% /warn %}}
```
sanderson, line 156: I don't know that we need to repeat this content here. It's outlined below. We can just link down to the existing content.

content/v2.0/influxdb-templates/cloud.md
sanderson: Yes, definitely. You nailed it here.

content/v2.0/reference/release-notes/influxdb.md
sanderson: Yeah, I think linking to the getting started page with instructions for downloading the binary would be better here.

content/v2.0/influxdb-templates/cloud.md
kelseiv: @sanderson does that mean that step 2 changes?....as is, I left it alone, but when I looked at the procedure referred to in step 2, I couldn't ascertain which bits were meant for the CLI only

sanderson: Sorry, misread your first question. No, step 2 does not need to change.

content/v2.0/get-started.md
kelseiv: thanks, @sanderson, I get what you're saying about not needing to tell them to click the button. My gut instinct when we start with "complete the following steps" is to include  something upfront so folks don't have to scan down -either numbered steps to indicate the order something needs to be done in or a bulleted list to indicate no specific order is required. That said, this section is pretty short and the corresponding content doesn't have numbered steps, so I took these out.

PR #1130, opened 2020-06-22
Add warning to 'group()' about sort order

### LINKED ISSUE #1129: Add warning to both `group` and `sort` sections that `group()` causes data to return out of order
## Describe the issue here ##

When applying a `group()` function to a query, the data is no longer guaranteed to arrive in order.  This causes a wonkiness in line graphs that look like the below examples.  It requires the use of `sort(columns: ["_time"])` in the query to avoid this issue in visualization.  In the `group()` docs, I feel it would suffice to warn the user of this behavior and show the function application needed to avoid the problem as well as a link to the `sort()` function docs.  In the `sort()` docs, perhaps it would be good to include that this is often used in tandem with `group()` as grouping has behavior that often necessitates it:
Screen Shot 2020-06-16 at 9 29 54 AM

Or:

Screen Shot 2020-06-22 at 11 10 21 AM

### Relevant URLs
https://v2.docs.influxdata.com/v2.0/query-data/flux/group-data/
https://v2.docs.influxdata.com/v2.0/query-data/flux/sort-limit/

### Impact

I believe our users will hit this issue rather often as it's not a common behavior among dashboarding tools.  Some may figure out how to solve it but I think it's important explicitly call out this behavior so it doesn't surprise anyone.

sanderson:
Closes #1129 

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/flux/stdlib/built-in/transformations/group.md
+++ b/content/v2.0/reference/flux/stdlib/built-in/transformations/group.md
@@ -37,6 +37,12 @@ group()
 `group()` does not guarantee the sort order of output records.
 To ensure data is sorted correctly, use [`sort()`](/v2.0/reference/flux/stdlib/built-in/transformations/sort/)
 after `group()`.
+
+```js
+data
+  |> group()
+  |> sort(columns: ["_time"])
+```
 {{% /warn %}}
 
 ## Parameters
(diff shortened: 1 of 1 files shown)


sanderson:
Oh, good call.

I'm merging this into the flux-0.70 branch, so it won't actually go live until that release. Should be some time later this week.

PR #1167, opened 2020-06-30
Add 'http.pathEscape()' function

### LINKED ISSUE #1165: Document `http.pathEscape()` function
PR: https://github.com/influxdata/flux/pull/2945

sanderson:
Closes #1165 

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/flux/stdlib/http/pathescape.md
+++ b/content/v2.0/reference/flux/stdlib/http/pathescape.md
@@ -1,0 +1,44 @@
+---
+title: http.pathEscape() function
+description: >
+  The `http.pathEscape()` function escapes special characters in a string (including `/`)
+  and replaces non-ASCII characters with hexadecimal representations (`%XX`).
+v2_0_ref:
+  name: http.pathEscape
+    parent: HTTP
+weight: 202
+---
+
+The `http.pathEscape()` function escapes special characters in a string (including `/`)
+and replaces non-ASCII characters with hexadecimal representations (`%XX`).
+
+_**Function type:** Transformation_
+
+```js
+import "http"
+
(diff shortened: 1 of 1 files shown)


content/v2.0/reference/flux/stdlib/http/pathescape.md
```diff
@@ -0,0 +1,45 @@
...
+  The `http.pathEscape()` function escapes special characters in a string (including `/`)
+  and replaces non-ASCII characters with hexadecimal representations (`%XX`).
+menu:
+  v2_0_ref:
+    name: http.pathEscape
+    parent: HTTP
+weight: 202
+---
+
+The `http.pathEscape()` function escapes special characters in a string (including `/`)
```
noramullen1, line 13: Maybe this will make sense to people more familiar with the subject matter, but I'm not sure what 'escape' as a transitive verb means?

sanderson: Sorry for referring to wikipedia on this one, but I was trying to find the best way to explain "escaping" in code, and the words weren't coming to me 😄 . 

> An escape character is a character that invokes an alternative interpretation on subsequent characters in a character sequence.

This is common vernacular in the context of programming. For example, if you have a string that contains quotes, you have to "escape" the quotes so they are not interpreted as the end of the string:

```sh
# Internal quotes are escaped with \ and interpreted as part of the string
"This is a string with internal \"quotes\" that I want interpreted as part of the string"

# Internal quotes are not escaped and are parsed as boundaries for strings.
# This gets interpreted as two separate strings with a `quotes` identifier between them.
"This is a string with internal "quotes" that I want interpreted as part of the string"
```

PR #1168, opened 2020-06-30
Add unix timestamp support to 'date' functions

### LINKED ISSUE #1163: Update Timeable Flux functions with integer support
All "Timeable" Flux functions now support time, duration, and integer values.

- time - RFC3339 Timestamp
- duration - Flux duration value (1m, 2h, etc.) relative to `now`
- integer - Nanosecond timestamp (relative to epoch time)

The following functions need to be updated:

- [x] All `date` package functions
- [ ] `range()`

sanderson:
Closes #1163

- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/flux/stdlib/date/nanosecond.md
+++ b/content/v2.0/reference/flux/stdlib/date/nanosecond.md
@@ -29,10 +29,11 @@ date.nanosecond(t: 2019-07-17T12:05:21.012934584Z)
 
 ### t
 The time to operate on.
-Use an absolute time or relative duration.
+Use an absolute time, relative duration, or integer.
 Durations are relative to `now()`.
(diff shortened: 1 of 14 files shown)


content/v2.0/reference/flux/stdlib/date/nanosecond.md
```diff
@@ -55,3 +56,12 @@ date.nanosecond(t: -2111984ns)
...
 // Returns 291422956
 ```
+
+##### Return the nanosecond for a nanosecond Unix timestamp
+```js
+import "date"
+
+date.nanosecond(t: 1581423663293534940)
+
+// Returns 293534940Z
```
kelseiv, line 66: RFC3339 format?

PR #1233, opened 2020-07-22
AWS Lambda content

noramullen1:
- Adapted from [Nate's blog post](https://www.influxdata.com/blog/how-to-expand-data-collection-for-influxdb-with-cloudformation-templates/)

- Creates new developer third party technologies page in Write Data > Developer Tools. (this is the only example on the page so far).


noramullen1:
@sanderson -- I was trying to make it parallel with [the other doc we have under No code solutions](https://v2.docs.influxdata.com/v2.0/write-data/no-code/third-party/), which doesn't separate them out. I'm not sure of any specific plans though to add anything else to this section in the near future. We could also separate out the ones on the other page. What do you think?

sanderson:
@noramullen1 It's probably good for now. We can revisit this in the future when we add more. Same for the other doc.

PR #1260, opened 2020-07-31
Add new Azure region data

pierwill:


sanderson:
The new region will enter beta sometime mid-August and should go GA in September.

kelseiv:
release Sept 1

PR #1263, opened 2020-08-03
Added Rows package and 'rows.map' function

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/v2.0/reference/flux/stdlib/contrib/rows/map.md
+++ b/content/v2.0/reference/flux/stdlib/contrib/rows/map.md
@@ -1,0 +1,223 @@
+---
+title: rows.map() function
+description: >
+  The `rows.map()` function is an alternate implementation of [`map()`](/v2.0/reference/flux/stdlib/built-in/transformations/map/)
+  that is faster, but more limited than `map()`.
+menu:
+  v2_0_ref:
+    name: rows.map
+    parent: Rows
+weight: 202
+v2.0/tags: [functions, package, map]
+related:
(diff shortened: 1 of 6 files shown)


content/v2.0/reference/flux/stdlib/contrib/rows/map.md
```diff
@@ -0,0 +1,223 @@
...
+    parent: Rows
+weight: 202
+v2.0/tags: [functions, package, map]
+related:
+  - /v2.0/reference/flux/stdlib/built-in/transformations/map/
+---
+
+The `rows.map()` function is an alternate implementation of [`map()`](/v2.0/reference/flux/stdlib/built-in/transformations/map/)
+that is faster, but more limited than `map()`.
+`rows.map()` cannot modify table groups key and, therefore, does not need to regroup tables.
```
noramullen1, line 18: table group keys? Unless this is an "attorneys general" situation I'm not understanding haha.

jsternberg:
This was a contrib function so it isn't a formal addition. I don't know how to currently deal with contrib, but if we document them, it needs to be clearly labeled as not part of the stdlib.

jsternberg:
The documentation looks good though.

content/v2.0/reference/flux/stdlib/contrib/rows/map.md
sanderson: I removed "table" and added a link to the "group key" definition in the glossary.

PR #1273, opened 2020-08-06
Beta 16 changelog

noramullen1:

### DIFF
--- a/content/v2.0/reference/release-notes/influxdb.md
+++ b/content/v2.0/reference/release-notes/influxdb.md
@@ -8,6 +8,34 @@ menu:
 weight: 101
 ---
 
+## v2.0.0-beta.16 [2020-08-06]
+
+### Breaking
+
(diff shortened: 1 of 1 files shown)


content/v2.0/reference/release-notes/influxdb.md
```diff
@@ -8,6 +8,34 @@ menu:
 weight: 101
 ---
 
+## v2.0.0-beta.16 [2020-08-06]
+
+### Breaking
```
sanderson, line 13: We also need to add that the `influx repl` command was removed. I'm adding the following warning to the `influx repl` doc:

```
{{% warn %}}
#### Removed in InfluxDB 2.0 beta-16
The `influx repl` command was removed in **InfluxDB 2.0 beta-16**.
To use the Flux REPL, build the REPL from source.
For more information, see the [Flux GitHub repository](https://github.com/influxdata/flux/#readme).
{{% /warn %}}
```

PR #1285, opened 2020-08-10
Add series cardinality note to tags info per DAR #106

noramullen1:
Addresses https://github.com/influxdata/DAR/issues/106

### DIFF
--- a/content/v2.0/reference/key-concepts/data-elements.md
+++ b/content/v2.0/reference/key-concepts/data-elements.md
@@ -107,6 +107,11 @@ location = portland, scientist = mullen
 Because tags are indexed, queries on tags are faster than queries on fields. This makes tags ideal for storing commonly-queried metadata.
 {{% /note %}}
 
+{{% note %}}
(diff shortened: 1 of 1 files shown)


content/v2.0/reference/key-concepts/data-elements.md
```diff
@@ -107,6 +107,11 @@ location = portland, scientist = mullen
 Because tags are indexed, queries on tags are faster than queries on fields. This makes tags ideal for storing commonly-queried metadata.
 {{% /note %}}
 
+{{% note %}}
+Tags containing highly variable information like UUIDs, hashes, and random strings will lead to a large number of series in the database, known as **high series cardinality**. High series cardinality is a primary driver of high memory usage for many database workloads. See [series cardinality](v2.0/reference/glossary/#series-cardinality) for more information.
```
sanderson, line 111: This is splitting hairs, but...

> ...random strings will lead to a large number of **unique** series in the database

abalone23: This comes from [1.x doc](https://docs.influxdata.com/influxdb/v1.8/concepts/schema_and_data_layout/#avoid-too-many-series) so that could be updated as well.

PR #1299, opened 2020-08-12
Updated examples for stacks

kelseiv:
Closes ##1230

- Update example to show string w quotes 
- Add GitOps workflow using stacks


content/v2.0/reference/cli/influx/stacks/init.md
```diff
@@ -35,10 +35,22 @@ influx stacks init [flags]
...
-# Initialize a stack with a name and description
+# Use environment variables
 influx stack init -n $STACK_NAME -d $STACK_DESCRIPTION
 
-# Initialize a stack with a name and urls to associate with stack.
+# Use strings
+influx stack init -n "Example Stack" -d "InfluxDB stack for monitoring some awesome stuff"
+```
+
+### Initialize a stack with a name and urls to associate with the stack
```
pierwill, line 48: URLs

content/v2.0/influxdb-templates/stacks/_index.md
```diff
@@ -13,38 +11,104 @@ related:
...
+- [Review ideal use cases for InfluxDB stacks](#review-ideal-use-cases-for-influxdb-stacks),
+  like automating InfluxDB deployments with GitOps and stacks.
+- [Manage InfluxDB stacks](#manage-influxdb-stacks)
 
-- [Actively develop and extend templates](#actively-develop-and-extend-templates)
+{{% note %}}
+**Key differences between stacks and templates**:
+
+- A template is a file outside of InfluxDB. Once a template is applied to InfluxDB, it becomes a stack.
+- Stacks only add a resource to an instance if the resource doesn't exist.
```
russorat, line 26: In addition to adding a resource, it will modify and delete them as well.

content/v2.0/influxdb-templates/stacks/_index.md
```diff
@@ -13,38 +11,104 @@ related:
...
-Stacks help save time and effort in the following use cases:
+- [Review ideal use cases for InfluxDB stacks](#review-ideal-use-cases-for-influxdb-stacks),
+  like automating InfluxDB deployments with GitOps and stacks.
+- [Manage InfluxDB stacks](#manage-influxdb-stacks)
 
-- [Actively develop and extend templates](#actively-develop-and-extend-templates)
+{{% note %}}
+**Key differences between stacks and templates**:
+
+- A template is a file outside of InfluxDB. Once a template is applied to InfluxDB, it becomes a stack.
```
russorat, line 25: When you apply an InfluxDB template, it automatically creates a stack which allows you to manage the state of that template. I'm not sure if "becomes a stack" is the right language here.

content/v2.0/influxdb-templates/stacks/_index.md
```diff
@@ -13,38 +11,104 @@ related:
...
-Stacks let you modify and update template manifests and apply those changes in
-any stack that uses the template.
+### Automate deployments with GitOps and stacks
+
+GitOps is popular way to configure and automate deployments. Use InfluxDB stacks in a GitOps workflow
+to automatically update distributed instances of InfluxDB OSS or InfluxDB Cloud.
+
+#### Set up a GitHub repository to back your InfluxDB instance
+
+Use our sample [`influxDB-assets` repository](https://github.com/russorat/influxdb-assets) to get started.
```
russorat, line 46: should we move this to the influxdata org?

content/v2.0/influxdb-templates/stacks/_index.md
kelseiv: Thank you, Russ!  Re what it means to "manage the state of a template:"

- If you apply a stack, it knows your environment has a resource. 
- If you apply a template, it doesn't know your environment has a resource. 

To help folks understand, perhaps we can clarify the reason why we made templates not recognize resources. What is the use case?

content/v2.0/influxdb-templates/stacks/_index.md
kelseiv: Excellent suggestion :)

content/v2.0/influxdb-templates/stacks/_index.md
russorat: yeah, a stack has a mapping between the meta.name field of the template file and the id of the created resource. When you "apply" the template again after making changes, it will manipulate the existing resource as long as you keep the meta.name field the same in the file/url/stream.

The reason for the two different "things" (template and stack) is that there are cases where you would want to duplicate a resource (maybe i want to bring in another copy of a dashboard or task and make changes to it) and there are instances where you would want to update an existing resource (deploying changes from a github repo to an environment).

PR #1304, opened 2020-08-13
Verify performance for nodes with more than ~20 GB of data on disk

kelseiv:
Closes #influxdata/DAR#87

This PR is on hold per request to Performance team to evaluate and verify whether the ≈20 GB of data on disk benchmark is valid: influxdata/influxdb#17592


content/enterprise_influxdb/v1.8/administration/anti-entropy.md
```diff
@@ -286,15 +286,17 @@ If a data node suddenly disappears due to a catastrophic hardware failure or for
...
 _View the [Replacing Data Nodes](/enterprise_influxdb/v1.8/guides/replacing-nodes/#replacing-data-nodes-in-an-influxdb-enterprise-cluster) documentation for instructions on replacing data nodes in your InfluxDB Enterprise cluster._
 
-### Replacing a machine that is running a data node
+### Replacing a machine running a data node
 
-Perhaps you are replacing a machine that is being decommissioned, upgrading hardware, or something else entirely.
-The Anti-Entropy service will automatically copy shards to the new machines.
+To replace a machine running a data node, start the Anti-Entropy service to copy shards to the new machines by doing one of the following:
+
+- For nodes with more than ≈20 GB of data on disk, ?
```
kelseiv, line 293: For nodes with more data, what's best practice? Removing and re-adding the data node?

content/enterprise_influxdb/v1.8/guides/replacing-nodes.md
```diff
@@ -69,14 +72,13 @@ The process of replacing both responsive and unresponsive data nodes is the same
...
-all shards on that node are lost.
+If a disk drive fails, all shards on that node are lost. Often in this scenario, you can replace the disk rather than replacing the entire host.
 
-Often in this scenario, rather than replacing the entire host, you just need to replace the disk.
-Host information remains the same, but once started again, the `influxd` process doesn't know
-to communicate with the meta nodes so the AE process can't start the shard-sync process.
+To restart AE and sync shards across the cluster, connect the `influxd` process running on the newly replaced disk to the cluster by doing one of the following:
 
-To resolve this, log in to a meta node and use the [`influxd-ctl update-data`](/enterprise_influxdb/v1.8/administration/cluster-commands/#update-data) command
+- For nodes with more than ≈20 GB of data on disk, ?
```
kelseiv, line 79: what's our best practice here?

content/enterprise_influxdb/v1.8/guides/replacing-nodes.md
```diff
@@ -274,7 +274,10 @@ The process of replacing data nodes is as follows:
 
 ### 2. Replace the old data node with the new data node
 
-Log into any of your cluster's meta nodes and use `influxd-ctl update-data` to replace the old data node with the new data node:
+- For nodes with more than ≈20 GB of data on disk, ?
```
kelseiv, line 277: ditto

kelseiv:
response from @andrew-influx:

My recommendation of 20GB on disk for a datanode is a shot in the dark. Hate to drag this one out but would be good to get a dev resource to way in on it.

@dgnorton Perhaps you can weigh in on good recommendations of when to (or when not to )use the update-data command for a cluster? Mayhap it's more stable then I think?

content/enterprise_influxdb/v1.8/guides/replacing-nodes.md
abalone23: @andrew-influx is it something like this?
· Make file copies of all shards to another location on network just in case
· influxd-ctl add-data
· manual shard rebalancing
· influxd-ctl copy-shard to copy shards to additional node
· influxd-ctl remove-shard to remove shards from node
· remove data node: influxd-ctl remove-data

content/enterprise_influxdb/v1.8/administration/anti-entropy/_index.md
```diff
@@ -287,15 +287,17 @@ If a data node suddenly disappears due to a catastrophic hardware failure or for
...
-Perhaps you are replacing a machine that is being decommissioned, upgrading hardware, or something else entirely.
-The Anti-Entropy service will automatically copy shards to the new machines.
+To replace a machine running a data node, doing one of the following:
 
-Once you have successfully run the `influxd-ctl update-data` command, you are free
-to shut down the retired node without causing any interruption to the cluster.
-The Anti-Entropy process will continue copying the appropriate shards from the
-remaining replicas in the cluster.
+- To start the Anti-Entropy service and copy shards to the new machine, on a meta node in your cluster, run [`influxd-ctl update-data`](/enterprise_influxdb/v1.8/administration/cluster-commands/#update-data), and then shut down the retired data node without causing any interruption to the cluster. The Anti-Entropy process continues to copy the appropriate shards from the remaining replicas in the cluster.
+- For nodes where the data set is too large or shaped in such a way that AE cannot be enabled, complete the following steps:
```
kelseiv, line 295: Thank you, @dgnorton! Re: "too large" and "shaped" in a way the AE cannot be enabled....we typically try to omit terms that are open to interpretation. Are there any specifics that you're comfortable committing to that we can add here? And/or is there any indicator/way to determine that AE cannot be enabled? Do folks try to turn AE on and just doesn't start?

dgnorton: @kelseiv "too large" depends on the shape of their data and computing resources available so there isn't really an exact guideline. They can try enabling AE and if CPU usage spikes too high for the cluster to operate then disable it. If it spikes but the cluster continues to run, it can be left on and see if CPU usage settles back to an acceptable level after a day. There will be a period of time after AE is first enabled where it's generating shard digests and comparing them with other nodes so some increase in CPU and disk I/O is expected.

PR #1305, opened 2020-08-13
Verify whether more than 500 shards per db is accurate metric 

kelseiv:
Closes [#2904
](https://github.com/influxdata/docs.influxdata.com/issues/2904)


content/enterprise_influxdb/v1.8/guides/migration.md
```diff
@@ -14,52 +14,66 @@ menu:
...
-4. [Upgrade the InfluxDB binary on your OSS instance](#upgrade-the-influxdb-oss-instance-to-influxdb-enterprise)
-5. [Add the upgraded OSS instance to the InfluxDB Enterprise cluster](#add-the-new-data-node-to-the-cluster)
-6. [Add existing data nodes back to the cluster](#add-existing-data-nodes-back-to-the-cluster)
-7. [Rebalance the cluster](#rebalance-the-cluster)
+4.  Do one of the following:
+    - [Migrate a data set with zero downtime](#migrate-a-data-set-with-zero-downtime). This method lets you move data between OSS and Enterprise as you're testing the migration.
+    - [Migrate a data set with downtime](#migrate-a-data-set-with-downtime). Note, with this method, you cannot move data from Enterprise back to OSS. This method is useful if you're not able to run a portable backup, for example, for any of the following reasons:
+     - Data set exceeds a certain size (roughly 10 databases or more, or more than 500 shards per database)
+     - Hardware requirements aren't available
+     - Time constraints
```
kelseiv, line 47: @Joe-Influx, I took the number of databases out. Is the following accurate:

Migrate a data set with downtime. 
With this method, you cannot move data from Enterprise back to OSS. This method is useful if you're not able to run a portable backup, for example, for any of the following reasons:
- Data set exceeds a certain size (more than 500 shards per database)
- Hardware requirements aren't available (fewer than 5 instances (3 meta 2 data))
- Time constraints

You mentioned folks might not be able to do a portable backup if they had time constraints...does it make sense to add further detail regarding time?

Joe-Influx:
The only part that gives me pause is `Data set exceeds a certain size (more than 500 shards per database)`. Is this an arbitrary number?

content/enterprise_influxdb/v1.8/guides/migration.md
dgnorton: I'm not sure where the suggested 500 shard limit originated. It probably depends more on the size of the shards, beefiness of the hardware, and maybe other factors combined. We may want to go with something along the lines of "try a portable backup first and use this other method as a fallback for large data sets". I think the time constraint wording is there because there's a known issue with backups being slow on some large data sets. Portable backups are the preferred method because it's the most straight forward but it may not work for all data sets.

kelseiv: Thank you, @dgnorton! Does it make sense to combine the "data set exceeds a certain size" and "time constraints"...or are there reasons other than time that a large data set might be prohibitive?

PR #1307, opened 2020-08-13
move OFFSET and WHERE subquery PR from v1 to v2

kelseiv:
Closes influxdata/DAR#90


content/influxdb/v1.8/query_language/explore-data.md
```diff
@@ -713,7 +713,7 @@ SELECT_clause FROM_clause [WHERE_clause] GROUP BY [* | <tag_key>[,<tag_key]]
 The order of the [tag keys](/influxdb/v1.8/concepts/glossary/#tag-key) is irrelevant.
 
 If the query includes a [`WHERE` clause](#the-where-clause) the `GROUP BY`
-clause must appear after the `WHERE` clause.
+clause must appear after the `WHERE` clause.
```
kelseiv, line 716: @ayang64 @jsteenb2
per the workaround in [EAR 1387](https://github.com/influxdata/EAR/issues/1387)
it appears we can include the WHERE clause after the GROUP BY clause.

Question:
Do we remove this note altogether, or does this note only not apply with subqueries --and is still valid for "main" queries?

kelseiv:
Thank you, @jsteenb2, sorry about that!

content/influxdb/v1.8/query_language/explore-data.md
```diff
@@ -3140,6 +3140,16 @@ Use a subquery to apply a query as a condition in the enclosing query.
 Subqueries offer functionality similar to nested functions and SQL
 [`HAVING` clauses](https://en.wikipedia.org/wiki/Having_%28SQL%29).
 
+{{% warn %}}
+Subqueries **do not support** WHERE or OFFSET. Include **WHERE or OFFSET outside a subquery**. For example:
```
jsternberg, line 3144: This isn't true. Subqueries do support `WHERE` and `OFFSET`, but we recommend **time constraints** be on the outermost.

It might be worth it to just fix this in the query engine as it doesn't make logical sense. But sometimes where clauses need to be in subqueries in particular when filtering based on an intermediary result.

content/influxdb/v1.8/query_language/explore-data.md
```diff
@@ -3324,3 +3334,13 @@ SELECT_clause FROM (SELECT_statement; SELECT_statement) [...]
 ```
 
 The system returns a parsing error if a subquery includes multiple `SELECT` statements.
+
+#### WHERE or OFFSET clause in a subquery
+
+Including the WHERE or OFFSET clause in a subquery can return incorrect results.
+InfluxQL supports the WHERE and OFFSET in main queries only. For example:
```
jsternberg, line 3341: > Including the WHERE or OFFSET clause in a subquery can return incorrect results.
> Time constraints should be placed in the outermost WHERE clause. For example:

ayang64 (changes requested):
please address jsternberg's comments.

but in either case, this is really stale.

PR #1308, opened 2020-08-13
Verify running data node under 50% memory utilization is best practice

kelseiv:
Closes influxdata/DAR#22


content/enterprise_influxdb/v1.7/administration/anti-entropy.md
```diff
@@ -10,7 +10,7 @@ menu:
 ---
 
 {{% warn %}}
-Prior to InfluxDB Enterprise 1.7.2, the Anti-Entropy (AE) service was enabled by default. When shards create digests with lots of time ranges (10s of thousands), some customers have experienced significant performance issues, including CPU usage spikes. If your shards include a small number of time ranges (most have 1 to 10, some have up to several hundreds) and you can benefit from the AE service, enable AE and monitor it closely to see if your performance is adversely impacted.
+Prior to InfluxDB Enterprise 1.7.2, the Anti-Entropy (AE) service was enabled by default. When shards create digests with lots of time ranges (10s of thousands), some customers have experienced significant performance issues, including CPU usage spikes. If your shards include a small number of time ranges (most have 1 to 10, some have up to several hundreds) and you can benefit from the AE service, enable AE and monitor it closely to see if your performance is adversely impacted. As a best practice, we recommend generally running data nodes under 50% memory utilization.
```
dgnorton, line 13: I think we should remove the 50% mem sentence here. I think we have general guidelines elsewhere in the docs about that? If not, I think that's the sort of recommendation that belongs in a general system requirements and sizing document.

kelseiv: Thank you, @dgnorton. @andrew-influx, per David's request, I removed the 50% memory utilization from the AE doc. 

We do allude to this memory requirement in data node configuration under `max-concurrent-compactions`:
_A value of `0` (unlimited compactions) results in 50% of `runtime.GOMAXPROCS(0)` used at runtime,
so when 50% of the CPUs aren't available, compactions are limited._

PR #1344, opened 2020-08-19
Remove REPL from influx CLI docs

pierwill:


pierwill:
This will break a few links, most of which will end up pointing to new content, or to the Flux Github page. I'm thinking it makes sense to merge this to remove the inaccurate information, then fix the links in follow up with new REPL instructions.

PR #1354, opened 2020-08-24
Clarify that window every duration must be positive and nonzero

jsternberg:
Related to influxdata/flux#3085.


content/v2.0/reference/flux/stdlib/built-in/transformations/window.md
```diff
@@ -50,6 +50,7 @@ including **calendar months (`1mo`)** and **years (`1y`)**.
 ### every
 Duration of time between windows.
 Defaults to `period` value.
+The duration must be a positive nonzero duration value.
```
sanderson, line 53: Minor update:

~The duration~ Must be a positive, nonzero duration value.

PR #1372, opened 2020-08-27
Update control-dashboard.md

bckeene:
Pulled out auto-refresh while UI team determines the best approach for reintroduction. @desa


kelseiv:
Added comments and pulling this auto-refresh feature out until the UI team determines the best approach for reintroduction. @desa @bckeene @sanderson

PR #1397, opened 2020-09-04
Pull old storage engine doc

pierwill:


sanderson (commented):
Marking the `_index.md` page as draft won't hide it in the nav. Hugo doesn't recognize/honor the draft status on `Section` pages (`_index.md`). I don't think we need to hide this page for beta 17. It may need a few adjustments, but not many. We just need to compare it with https://docs.influxdata.com/influxdb/v1.8/concepts/storage_engine/#storage-engine.

PR #1423, opened 2020-09-09
New support and feedback layout

sanderson:
image

- [x] Rebased/mergeable


timhallinflux (commented):
do we want to lead folks to the Slack community as well as discourse?  @mhall119

mhall119:
I'd like to point them to both, since some people are more comfortable engaging in one or the other platform.

Also "All InfluxDB users can post in the InfluxData Community." makes it sound like non-users shouldn't, which isn't the case. Anybody can post to either Slack or Discourse, as long as the post is relevant to the InfluxDB community.

sanderson:
@timhallinflux @mhall119 Added the community Slack link and restyled them a bit:

image

timhallinflux (approved):

PR #1447, opened 2020-09-15
Fix quantile() data type

pierwill:
Closes https://github.com/influxdata/DAR/issues/126.

- [x] Rebased/mergeable


sanderson (changes requested):
This change is incorrect. The `column` parameter requires a string value. Not a float.

PR #1454, opened 2020-09-16
Update using_http.md

samhld:
Closes #

__Describe your proposed changes here.__
I think the title of the guide should be more generic to capture anyone trying to use Telegraf to scrape JSON data.  The fact that it uses Citi Bike data should just be a side effect of the guide’s content.  That way the guide will immediately seem more accessible to people.

- [ ] Rebased/mergeable


sanderson (changes requested):
This is a long title that will take up a lot of space (it was before too). What about:

> Using the HTTP input plugin with JSON endpoints

samhld:
@sanderson done!

sanderson (approved):
🎉 Thanks @samhld !

Issue #1280, opened 2020-08-10
Tasks set Now time based on the schedule


lyondhill:
When a task executes a query it adjusts the "now" time that is sent to the query system. This allows for exact query times without having to adjust the query the user is writing. I'm not sure how to convey this to the user. In essence it takes the load off the user to set any specific now time based on the query executing late or early. 

By using the scheduled time we actually allow for the user to use "offset" to delay the execution but without adjusting the "now" time.
Example:
User has a task they want to run on the hour but account for latent data from edge devices. The user can set a Schedule for "@every 1h" and then set a offset for "5m" and the task will execute 5m late but with a "now" set on the query for the exact hour time.
##### Relevant URLs
- https://v2.docs.influxdata.com/v2.0/process-data/get-started/

PR #1538, opened 2020-09-23
add task offset + now time

kelseiv:
Closes https://github.com/influxdata/docs-v2/issues/1280

@lyondhill, I added a note to our tasks docs, but want to be sure I understand the change....the way our docs read, I thought the change you described was how scheduling the task interval, offsets, and `now()` already worked

### DIFF
--- a/content/influxdb/v2.0/process-data/task-options.md
+++ b/content/influxdb/v2.0/process-data/task-options.md
@@ -40,6 +40,7 @@ options task = {
 ```
 
 ## every
+
 The interval at which the task runs.
 
 _**Data type:** Duration_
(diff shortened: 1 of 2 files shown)


content/influxdb/v2.0/process-data/task-options.md
```diff
@@ -70,6 +71,7 @@ options task = {
 ```
 
 ## offset
+
 Delays the execution of the task but preserves the original time range.
 For example, if a task is to run on the hour, a `10m` offset will delay it to 10
 minutes after the hour, but all time ranges defined in the task are relative to
```
kelseiv, line 77: @lyondhill, is this offset definition still accurate?

content/influxdb/v2.0/process-data/task-options.md
```diff
@@ -40,6 +40,7 @@ options task = {
 ```
 
 ## every
+
 The interval at which the task runs.
```
kelseiv, line 44: Is this every definition still accurate?

lyondhill (approved):
Looks great. All the definitions are very clear

PR #1562, opened 2020-09-29
Add page on troubleshooting systemd permissions

### LINKED ISSUE #1550: Add troubleshooting FAQ section for "Unable to open..." and "permission denied" errors
When an InfluxDB package is installed on an instance with systemd, an `influxdb` user and group is automatically created. If the user runs an `influxd` process directly with their login shell (typically `root`), it can generate new series files not accessible by the `influxdb` user. When systemd starts the InfluxDB service (via `sudo systemctl start influxdb`), the InfluxDB process will exit because it cannot access the leftover files owned by the `root` user. The error in the InfluxDB logs (via `journalctl -u influxdb` typically includes text like:

- `error msg="Unable to open series file"`
- `run: open server: open tsdb store: mkdir /var/lib/influxdb/data/_internal/_series/00: permission denied`

The resolution is to set all files in the InfluxDB directories to be owned by the `influxdb` user and group.

```
sudo chown -R influxdb:influxdb /var/lib/influxdb/*
``` 

Or reset the database by removing all files.

##### Relevant URLs
- https://docs.influxdata.com/platform/troubleshoot/
- or similar page, maybe the FAQ

InfluxDB 1.x
InfluxDB Enterprise 1.x

pierwill:
Closes #1550.

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/v1.8/troubleshooting/systemd.md
+++ b/content/influxdb/v1.8/troubleshooting/systemd.md
@@ -1,0 +1,33 @@
+---
+title: Troubleshoot systemd errors
+list_title: systemd permission errors
+description: Troubleshoot errors with InfluxDB and systemd permissions
+menu:
+  platform:
+    name: systemd errors
+    parent: Troubleshoot
+    weight: 1
+---
+
+When running InfluxDB using systemd (Ubuntu, Debian, CentOS), you might encounter errors in the InfluxDB logs (via `journalctl -u influxdb`) like:
+
+- `error msg="Unable to open series file"`
+- `run: open server: open tsdb store: mkdir /var/lib/influxdb/data/_internal/_series/00: permission denied`
+
+When InfluxDB is installed with systemd, an `influxdb` user and group is automatically created.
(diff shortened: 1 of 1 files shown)


content/platform/troubleshoot/systemd.md
```diff
@@ -0,0 +1,33 @@
...
+list_title: systemd permission errors
+description: Troubleshoot errors with InfluxDB and systemd permissions
+menu:
+  platform:
+    name: systemd errors
+    parent: Troubleshoot
+    weight: 1
+---
+
+When running InfluxDB using systemd, you might encounter errors in the InfluxDB logs (via `journalctl -u influxdb`) like:
```
gunnaraasen, line 12: > When running InfluxDB using systemd (Ubuntu, Debian, CentOS),

gunnaraasen (changes requested):
Looks good! Since the platform section doesn't have good discoverability in the new docs layout. Could this be moved to the InfluxDB 1.x [troubleshooting section](https://docs.influxdata.com/influxdb/v1.8/troubleshooting/)?

PR #1633, opened 2020-10-08
Document debug vars `valuesWrittenOK` and `pointsWrittenOK`

### LINKED ISSUE #1608: Update _internal db docs for OSS 1.8.3
Tim noticed an InfluxDB OSS 1.8.3 change (tracking values written via the `/debug/vars HTTP` endpoint)
that impacts the _internal db docs. See the following pages to determine what needs to be updated:

- https://docs.influxdata.com/platform/monitoring/influxdata-platform/tools/measurements-internal/#visualize-influxdb-internal-metrics
- https://docs.influxdata.com/platform/monitoring/influxdata-platform/tools/measurements-internal/#write

pierwill:
Closes #1608 

- [x] Rebased/mergeable

### DIFF
--- a/content/platform/monitoring/influxdata-platform/tools/measurements-internal.md
+++ b/content/platform/monitoring/influxdata-platform/tools/measurements-internal.md
@@ -228,9 +228,11 @@ to visualize InfluxDB `_internal` metrics.
   - [pointReqHH](#pointreqhh-enterprise-only) (Enterprise only)
   - [pointReqLocal](#pointreqlocal-enterprise-only) (Enterprise only)
   - [pointReqRemote](#pointreqremote-enterprise-only) (Enterprise only)
+  - [pointsWrittenOK](#pointsWrittenOK)
   - [req](#req)
   - [subWriteDrop](#subwritedrop)
   - [subWriteOk](#subwriteok)
+  - [valuesWrittenOK](#valuesWrittenOK)
   - [writeDrop](#writedrop)
   - [writeError](#writeerror)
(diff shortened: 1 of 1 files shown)


pierwill:
In the code these are defined here: https://github.com/influxdata/influxdb/blob/master-1.x/services/httpd/service.go#L21-L47.

kelseiv (commented):
> https://github.com/influxdata/influxdb/blob/master-1.x/services/httpd/service.go#L21-L47

Will, it looks like Scott added the `pointsWrittenOK` to the [httpd](http://localhost:1313/platform/monitoring/influxdata-platform/tools/measurements-internal/#httpd) section a couple months ago. Let's check whether `valuesWrittenOK` should also be added to httpd section, whether these fields are stored in both the httpd and write measurements (and if so, let's use the same description in both places).

ayang64:
@pierwill , @kelseiv  -- please correct me if i'm wrong but i think you're asking if the `pointsWrittenOK` and `valuesWrittenOK` stats we see in the `show stats` output is the same as the values we see in the debug vars obtained via the http endpoint.

to that the answer is yes. :)

did i understand the question?

kelseiv:
Thank you, @ayang64 @pierwill, 

My question was this:
Should `pointsWrittenOK` and `valuesWrittenOK` be in both the `httpd` measurement and the `write` measurement?

In this PR, we added both values to the `write` measurement. The`pointsWrittenOK` value was already included in the `httpd` measurement. Do we also add `valuesWrittenOK` to the `httpd` measurement?

PR #1692, opened 2020-10-21
Add information about barrier.idle and window.period

sanderson:
Closes influxdata/DAR#136

- [x] Rebased/mergeable

### DIFF
--- a/content/kapacitor/v1.5/nodes/barrier_node.md
+++ b/content/kapacitor/v1.5/nodes/barrier_node.md
@@ -26,13 +26,13 @@ stream
   |from()
     .measurement('cpu')
   |barrier()
-    .idle(11s)
+    .idle(11s) // Must be longer than the window period
     .delete(TRUE)
   |window()
     .period(10s)
     .every(5s)
   |top(10, 'value')
-  //Post the top 10 results over the last 10s updated every 5s.
+  // Post the top 10 results over the last 10s updated every 5s.
   |httpPost('http://example.com/api/top10')
 ```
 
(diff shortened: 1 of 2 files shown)


danatinflux (commented):
could we add a comment inline on line 29? Maybe something like:

.idle(11s) // one second longer than window period

Does that make it better or more confusing?

content/kapacitor/v1.5/nodes/barrier_node.md
```diff
@@ -35,24 +35,31 @@ stream
...
-| Setters | Description |
-|:---|:---|
-| **[idle](#idle)&nbsp;(&nbsp;`value`&nbsp;`time.Duration`)** | Emit barrier based on idle time since the last received message. Must be greater than zero.  |
-| **[period](#period)&nbsp;(&nbsp;`value`&nbsp;`time.Duration`)** | Emit barrier based on periodic timer.  The timer is based on system clock rather than message time. Must be greater than zero.  |
-| **[delete](#delete)&nbsp;(&nbsp;`value`&nbsp;`Boolean`)** | Delete the group after processing each barrier. |
-| **[quiet](#quiet)&nbsp;(&nbsp;)** | Suppress all error logging events from this node.  |
-
+| Setters                                                         | Description                                                                                                                   |
+|:---                                                             |:---                                                                                                                           |
+| **[idle](#idle)&nbsp;(&nbsp;`value`&nbsp;`time.Duration`)**     | Emit barrier based on idle time since the last received message. Must be greater than zero **and the window `period`**.       |
```
kelseiv, line 53: extra word, but may read better?:
Must be greater than zero **and longer than the window `period`.**

kelseiv (approved):

content/kapacitor/v1.5/nodes/barrier_node.md
danatinflux: I agree, the 'longer than' really helps.

PR #1706, opened 2020-10-25
typo: wrong flag

Eraac:
Using `--read-buckets 03a2bbf46309a000 03ace3a87c269000` instead of `--read-bucket 03a2bbf46309a000 03ace3a87c269000` will allow read on all buckets instead of those specify

### DIFF
--- a/content/influxdb/v2.0/security/tokens/create-token.md
+++ b/content/influxdb/v2.0/security/tokens/create-token.md
@@ -42,7 +42,7 @@ influx auth create -o <org-name> [permission-flags]
 
 # Example
 influx auth create -o my-org \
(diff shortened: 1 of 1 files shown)


pierwill:
According to [this page](https://docs.influxdata.com/influxdb/v2.0/reference/cli/influx/auth/create/), it looks like `--read-buckets` is correct syntax when specifying multiple buckets.

Eraac:
@pierwill 
image

`--read-buckets` allow *all* buckets, no matter parameters used

pierwill:
Ah. Thanks @Eraac. We'll look more into this and get it fixed!

PR #1757, opened 2020-11-04
fix influx to influxd

armadadigital:
Fix typo to influxd backup command.

- [ ] Rebased/mergeable


sanderson:
Thanks for the contribution @armadadigital, but this isn't a typo. The `backup` command is part of the `influx` CLI and should be executed while InfluxDB (`influxd`) is running. The `restore` command is part of the `influxd` service and should only be run while InfluxDB is not running.

PR #1814, opened 2020-11-11
Add missing TLS configuration options

pierwill:
`data-use-tls = true` and `data-insecure-tls = true` allow the meta notes to communicate with the data notes using TLS (with self-signed certificates).

Closes https://github.com/influxdata/DAR/issues/140.

- [x] Rebased/mergeable


danatinflux:
@pierwill LGTM, thanks for the fix!

danatinflux (approved):
LGTM, thanks for the fix!

PR #1822, opened 2020-11-12
Remove -L flag from curl requests for PGP keys

pierwill:
These keys shouldn't move, so we don't want any redirects going on here.

- [x] Rebased/mergeable


sanderson (approved):
LGTM

Issue #1831, opened 2020-11-16
Description of OSS behavior for V1 /write endpoint is out-of-date


danxmoran:
The OSS 2.0 docs say that DBRP mappings are automatically created (if needed) when data is written to the V1 `/write` compatibility endpoint. This behavior was removed between the last RC and GA, to avoid auth problems and confusing behavior.

##### Relevant URLs
- Doc section [here](https://docs.influxdata.com/influxdb/v2.0/reference/api/influxdb-1x/dbrp/#when-writing-data)
- InfluxDB PR that changed the behavior: influxdata/influxdb#19960

##### What products and version are you using?
InfluxDB OSS 2.0.1

Issue #1830, opened 2020-11-16
OSS 2.0.2


kelseiv:
- [x] Nora: Add 2.0.2 release notes from [changelog](https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md) 
- [x] Cross-check release announcement to ensure highlights in release notes and announcement align
- [x] Update versions

**Update docs for the following issues as applicable:**
- [ ] Will: [Improvements to influxd upgrade command](https://github.com/influxdata/influxdb/issues/19985)

- [x] Nora: [Add dbrp management to the CLI #19950](https://github.com/influxdata/influxdb/issues/19950)

- [x] NORA: Description of OSS behavior for V1 /write endpoint is out-of-date https://github.com/influxdata/docs-v2/issues/1831

- [x] Scott: Improvements to `influxd upgrade` command 
https://github.com/influxdata/influxdb/issues/20030

- [x] Incorrect input validation in `influxd upgrade`
https://github.com/influxdata/influxdb/issues/20038

- [x] feat: [Port Pipeline test framework and comprehensive InfluxQL tests](https://github.com/influxdata/influxdb/pull/20004)

- [x] https://github.com/influxdata/influxdb/pull/19878

- [x] Will: https://github.com/influxdata/influxdb/pull/20036 (warn user during upgrade if auth was not enabled for 1.x)

- [x] Scott: provide an option to dump continuous queries during upgrade https://github.com/influxdata/influxdb/pull/20039 
- [ ] Will: include info about what happens if HTTP auth is disabled. https://github.com/influxdata/influxdb/pull/20039 

- [x] Kelly: fix: Don't include duplicates for SHOW DATABASES https://github.com/influxdata/influxdb/pull/20017 (verified this is release notes only)

- [ ] Will: influxd CLI reference docs are missing info on `upgrade` subcommands https://github.com/influxdata/docs-v2/issues/1813

- [x] Scott: influxd CLI reference docs show unavailable subcommands https://github.com/influxdata/docs-v2/issues/1812

PR #1832, opened 2020-11-16
Addresses #1831

noramullen1:
Also addresses #1830

### DIFF
--- a/content/influxdb/v2.0/reference/api/influxdb-1x/dbrp.md
+++ b/content/influxdb/v2.0/reference/api/influxdb-1x/dbrp.md
@@ -36,7 +36,7 @@ the default retention policy for the specified database.
 
 ### When writing data
 
-When writing data to InfluxDB Cloud and InfluxDB OSS 2.0 using the
+When writing data using the
 [`/write` compatibility endpoint](/influxdb/v2.0/reference/api/influxdb-1x/write/),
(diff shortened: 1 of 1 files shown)


content/influxdb/v2.0/reference/api/influxdb-1x/dbrp.md
```diff
@@ -47,9 +47,6 @@ the DBRP mapping service checks for a bucket mapped to the database and retentio
     - **database** exists (without a specified retention policy), the default
       database retention policy is used, a DBRP mapping is added to the bucket,
       and data is written to the bucket.
-- If no matching bucket is found, a new **database/retention-policy** bucket is
```
danxmoran, line 50: (AFAIK) the point above is also outdated, DBRP mappings aren't auto-created in any situation.

This change is unique to OSS, so the top of the section that says `When writing data to InfluxDB Cloud and InfluxDB OSS 2.0` will need to change. I'm not sure how splits are handled in situations like this...

sanderson (approved):

content/influxdb/v2.0/reference/api/influxdb-1x/dbrp.md
noramullen1: Thanks @danxmoran , updated both.

PR #1848, opened 2020-11-17
Add note on auth-enabled settings to v1 upgrade guide

pierwill:
- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/v2.0/upgrade/v1-to-v2/_index.md
+++ b/content/influxdb/v2.0/upgrade/v1-to-v2/_index.md
@@ -130,12 +130,14 @@ to scrape data from the `/metrics` endpoint and store them in a bucket.
 
 ### Secure by default
 
-InfluxDB 2.0 requires authentication and does not support the InfluxDB 1.x
-`auth-enabled = false` configuration option.
-Consider [enabling authentication in your InfluxDB 1.x instance](/influxdb/v1.8/administration/authentication_and_authorization/#set-up-authentication)
-**before** upgrading to InfluxDB 2.0 to ensure the appropriate credentials are in place and
-that the various applications, agents, and visualization tools are able to connect.
+InfluxDB 2.0 requires authentication and does not support the InfluxDB 1.x `auth-enabled = false` configuration option.
 
+Before upgrading to 2.0, [enable authentication in your InfluxDB 1.x instance](/influxdb/v1.8/administration/authentication_and_authorization/#set-up-authentication)
+and test your credentials to ensure your applications, agents, and visualization tools can connect to InfluxDB.
+
+If you upgrade with `auth-enabled = false`, the upgrade may appear complete,
+but client requests to InfluxDB 2.0 may be silently ignored (you won't see a notification the request was denied).
+ 
 ## Perform the upgrade
 
 If you've considered the [guidance above](#before-you-begin-important-considerations)


content/influxdb/v2.0/upgrade/v1-to-v2/_index.md
```diff
@@ -136,6 +136,9 @@ Consider [enabling authentication in your InfluxDB 1.x instance](/influxdb/v1.8/
 **before** upgrading to InfluxDB 2.0 to ensure the appropriate credentials are in place and
 that the various applications, agents, and visualization tools are able to connect.
 
+If `auth-enabled = false` is set in your InfluxDB 1.x configuration, after upgrading to InfluxDB 2.0,
```
kelseiv, line 139: If they set `auth-enabled = true`...does this resolve credentials issue, or are there additional steps to ensure their credentials will work after upgrading?

content/influxdb/v2.0/upgrade/v1-to-v2/_index.md
```diff
@@ -136,6 +136,9 @@ Consider [enabling authentication in your InfluxDB 1.x instance](/influxdb/v1.8/
...
+clients sending requests (for example, Telegraf) will have invalid credentials and requests will be ignored (and return `401 Unauthorized`).
```
timhallinflux, line 140: this isn't what happens.

I would suggest the following:
If you’ve been running with `auth-enabled = false` in your 1.x setup, a warning is now produced as this is not an option in 2.x.  If you see this warning, it is recommended to reconfigure your 1.x instance with `auth-enabled = true` and ensure all of the username/password combinations are working as expected before retrying the upgrade process.

In my situation, I THOUGHT that auth was enabled and was being enforced as I had username and passwords everywhere.  However, when I upgraded, it turns out that one of the passwords that I was using was incorrect -- but since this was NOT being enforced, it only caused problems once the upgrade was completed.  So, the process outlined above is solid guidance to ensure that everything is secure and working before upgrading....i.e. work out the security bits on your known good environment first, then proceed with the upgrade.  Think about it like this... turning auth on is a prerequisite to performing the upgrade.

pierwill: Thanks, @timhallinflux.

content/influxdb/v2.0/upgrade/v1-to-v2/_index.md
```diff
@@ -132,10 +132,14 @@ to scrape data from the `/metrics` endpoint and store them in a bucket.
...
 InfluxDB 2.0 requires authentication and does not support the InfluxDB 1.x
 `auth-enabled = false` configuration option.
-Consider [enabling authentication in your InfluxDB 1.x instance](/influxdb/v1.8/administration/authentication_and_authorization/#set-up-authentication)
+You should [enable authentication in your InfluxDB 1.x instance](/influxdb/v1.8/administration/authentication_and_authorization/#set-up-authentication)
 **before** upgrading to InfluxDB 2.0 to ensure the appropriate credentials are in place and
 that the various applications, agents, and visualization tools are able to connect.
 
+If you attempt to upgrade with `auth-enabled = false` in your 1.x configuration, a warning is produced.
+If you see this warning, it is recommended to reconfigure your 1.x instance with `auth-enabled = true`
+and ensure all of the username and password combinations are working as expected before retrying the upgrade process.
```
kelseiv, line 141: Before upgrading to 2.0, [enable....]() to ensure your applications, agents, and visualization tools have authenticated credentials to connect to InfluxDB. If you upgrade with `auth-enabled = false`, the upgrade may appear complete, but client requests to InfluxDB 2.0 may be silently ignored (you won't see a notification the request was denied).

timhallinflux: yep  ^^.  Perfect.

timhallinflux (changes requested):

kelseiv (approved):

PR #1864, opened 2020-11-18
Update external Telegraf plugins

sanderson:
Closes influxdata/influxdata.com#280

- Provided links for external Telegraf plugins
- Added styles for external flag that appears on each external plugin card.

**Note:** There are existing external plugins that aren't included in Telegraf plugin list. I would've added the, but I wasn't sure if they were bound to a specific Telegraf version.

- [x] Rebased/mergeable

### DIFF
--- a/assets/styles/layouts/article/_telegraf-plugins.scss
+++ b/assets/styles/layouts/article/_telegraf-plugins.scss
@@ -35,6 +35,15 @@
         color: $article-code-accent7;
       }
     }
+    &.external {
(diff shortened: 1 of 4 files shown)


sjwang90:
@sanderson the external plugins aren't bound to any Telegraf release but the prior ones were added to our list around the 1.15 timeframe so if you need to tie them to a release it can be that one.

kelseiv (approved):

PR #1868, opened 2020-11-19
Change wrong ordID in example

Nemecsek:
Param `orgID` in the second part of the example was wrongly set as organization label instead of numerical ID.


sanderson (approved):
Thanks @Nemecsek!

PR #1900, opened 2020-11-30
alter retention policy statement clarify

StoneYunZhao:
The precise statement of create retention policy is:
```
alter_retention_policy_stmt  = "ALTER RETENTION POLICY" policy_name on_clause
                               retention_policy_option
                               [ retention_policy_option ]
                               [ retention_policy_option ]
                               [ retention_policy_option ] .
```
But here is: 
```
ALTER RETENTION POLICY <retention_policy_name> ON <database_name> DURATION <duration> REPLICATION <n> SHARD DURATION <duration> DEFAULT
```

Actually, `ALTER RETENTION POLICY` only need to declare at least one of tokens DURATION, REPLICATION, SHARD DURATION, DEFAULT. But the second description means we should declare all of the token at the exactly same order. 

So I think the correct description is:
```
ALTER RETENTION POLICY <retention_policy_name> ON <database_name> [DURATION <duration>] [REPLICATION <n>] [SHARD DURATION <duration>] [DEFAULT]
```

- [ ] Rebased/mergeable


sanderson (approved):
Thanks @StoneYunZhao

PR #1938, opened 2020-12-04
Update aggregator_processor_plugins.md

Nemecsek:
Fixed a visual problem: changed ASCII schematics that wasn't properly formatted.

Closes #

- [ ] Rebased/mergeable


sanderson:
Thanks @Nemecsek. What OS/browser are you using to view the docs? If you're using the font provided by the docs CSS (Roboto Mono), the schematic should render correctly. With your changes, it not longer renders correctly with Roboto Mono, so I'm curious what you're seeing and why it's different.

PR #1939, opened 2020-12-04
fix(flux/stdlib): clarify that the end index for strings.substring is exclusive

jsternberg:
The end index for `strings.substring` is exclusive so clarifying the documentation to show it as exclusive.

Related to influxdata/flux#3368.


sanderson (approved):
Looks good @jsternberg. Thanks for the clarification.

PR #1959, opened 2020-12-08
Add the Flux 'operator' profiler

### LINKED ISSUE #1944: Flux Profiler Package page is missing info on operator profiler
There is a profiler in addition to the `query` profiler which is documented. It is called the `operator` profiler.

Here's an example of that profiler in action:
Screen Shot 2020-12-04 at 2 56 26 PM

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/profiler/

Cloud 2.0 (and OSS)

sanderson:
Closes #1944 

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/v2.0/reference/flux/stdlib/profiler/_index.md
+++ b/content/influxdb/v2.0/reference/flux/stdlib/profiler/_index.md
@@ -32,12 +32,14 @@ _**Data type:** Array of strings_
 ```js
 import "profiler"
 
-option profiler.enabledProfilers = [""]
+option profiler.enabledProfilers = ["query", "operator"]
 ```
 
-#### Available profilers
+## Available profilers
+- [query](#query)
+- [operator](#operator)
 
-##### query
+### query
 The `query` profiler provides statistics about the execution of an entire Flux script.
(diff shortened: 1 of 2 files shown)


content/influxdb/v2.0/reference/flux/stdlib/profiler/_index.md
```diff
@@ -56,14 +58,18 @@ include a table with the following columns:
...
-
-```js
-import "profiler"
-
-option profiler.enabledProfilers = ["query"]
+### operator
+The `operator` profiler output statistics about each operation in a query.
+[Pushdown operations](/influxdb/v2.0/query-data/optimize-queries/#start-queries-with-pushdown-functions)
+that execute in the data source storage engine are returned as a single operation.
+Operations executed in memory are returned as individual operations.
```
wolffcm, line 65: > Operations executed in memory are returned as individual operations. 

This is a little misleading since the pushed down operations are also often executed in memory, it's just they are executed in the storage tier. You might write "operations executed in the storage tier" instead. (Feel free to disregard if you want to keep consistent terminology with elsewhere in the docs though)

wolffcm (approved):
Looks great, I just had one comment. Thanks!

PR #1965, opened 2020-12-09
docs(flux): add a note to map about invalid column types

jsternberg:
Related to influxdata/flux#3094.


sanderson:
Thanks for this @jsternberg! I updated the format and wording a bit.

sanderson (approved):

PR #1977, opened 2020-12-11
Flux notification fixes

### LINKED ISSUE #1974: Invalid example for monitor.notify
The example code for `monitor.notify()` is invalid for at least two reasons:
* First, the example shows a call to `slack.endpoint()` passing `name` and `channel` parameters, but the only parameters `slack.endpoint()` takes are `url` and `token`.
* Second, the example shows a call to `monitor.notify()` passing only an `endpoint` parameter, but, `monitor.notify()` _requires_ you pass an explicit `data` parameter. The documentation states that "InfluxDB populates notification data", but, it doesn't clarify what that actually _means_; it definitely _doesn't_ mean the parameter can be simply omitted.

##### Relevant URLs
- https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/monitor/notify/
- https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/monitor/notify/

InfluxDB Cloud 2.0

### LINKED ISSUE #1976: slack.endpoint function documentation recommends using non-existent defaults
The documentation for the `slack.endpoint()` function explicitly recommends relying on the default value of `mapFn`:
>You should rarely need to override the default `mapFn` parameter. To see the default `mapFn` value or for insight into possible overrides, view the [`slack.endpoint()` source code](https://github.com/influxdata/flux/blob/master/stdlib/slack/slack.flux).

However, if you follow the link to actually _view_ the `slack.endpoint()` source code, you will observe that `mapFn` _has_ no default, and indeed, trying to call the function returned by `slack.endpoint()` without an explicitly value for `mapFn` results in the error, "missing required argument mapFn".

##### Relevant URLs
https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/slack/endpoint/
https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/slack/endpoint/

InfluxDB Cloud 2.0

sanderson:
Closes #1976
Close #1974 

- [x] Rebased/mergeable


content/influxdb/v2.0/reference/flux/stdlib/monitor/notify.md
```diff
@@ -29,26 +29,51 @@ monitor.notify(
...
+	_notification_endpoint_id: "0000000000000002",
+	_notification_endpoint_name: "example-endpoint-name",
+}
 
 from(bucket: "system")
 	|> range(start: -5m)
-	|> monitor.notify(endpoint: endpoint)
+	|> monitor.notify(
+    endpoint: endpoint,
+    data: notification
```
philomory, line 77: I think this should be `notification_data`, given the assignment statement above.

content/influxdb/v2.0/reference/flux/stdlib/monitor/notify.md
```diff
@@ -29,26 +29,51 @@ monitor.notify(
...
+    text: r._message,
+    color: "danger"
+  }))()
+
+notification_data = {
+	_notification_rule_id: "0000000000000001",
+	_notification_rule_name: "example-rule-name",
+	_notification_endpoint_id: "0000000000000002",
+	_notification_endpoint_name: "example-endpoint-name",
+}
```
philomory, line 71: When you're writing your own custom notification task, can you just pick arbitrary values for these? Is it a problem if you pick the same value some other task is already using? Can you use the same endpoint ID in multiple tasks if you're using the same package, URL, token, etc when calling `<package>.endpoint()`?

content/influxdb/v2.0/reference/flux/stdlib/slack/endpoint.md
```diff
@@ -39,7 +39,7 @@ If using a Slack webhook, you'll receive a Slack webhook URL when you
 _**Data type:** String_
 
 ### token
-The [Slack API token](https://get.slack.help/hc/en-us/articles/215770388-Create-and-regenerate-API-tokens)
+({{< req >}}) The [Slack API token](https://get.slack.help/hc/en-us/articles/215770388-Create-and-regenerate-API-tokens)
```
philomory, line 42: I don't think this is actually required, seeing as it has a default value. If you're using an old-style Webhook URL, you don't need a token.

content/influxdb/v2.0/reference/flux/stdlib/monitor/notify.md
sanderson: Yep, good catch

content/influxdb/v2.0/reference/flux/stdlib/monitor/notify.md
sanderson: I've added the following to this section:

> The InfluxDB monitoring and alerting system uses `monitor.notify()` to store information about sent notifications and automatically assigns these values. If writing a custom notification [task](/influxdb/v2.0/process-data/), use **arbitrary, but unique** values.

You can use existing IDs and names, but it will intermingle logs and other metrics with that probably shouldn't be intermingled.

philomory: That's great, very helpful! Thanks!

content/influxdb/v2.0/reference/flux/stdlib/monitor/notify.md
```diff
@@ -29,26 +29,56 @@ monitor.notify(
...
+
+- _notification_rule_id
+- _notification_rule_name
+- _notification_endpoint_id
+- _notification_endpoint_name
+
+The InfluxDB monitoring and alerting system uses `monitor.notify()` to store
+information about sent notifications and automatically assigns these values.
+If writing a custom notification [task](/influxdb/v2.0/process-data/),
+use **unique arbitrary values**.
```
kelseiv, line 53: for data record fields?

content/influxdb/v2.0/reference/flux/stdlib/monitor/notify.md
kelseiv: We recommend using...for logging purposes?

content/influxdb/v2.0/reference/flux/stdlib/monitor/notify.md
```diff
@@ -29,26 +29,56 @@ monitor.notify(
...
+
 ## Examples
```
kelseiv, line 55: Shall we add examples for pagerduty/http?

content/influxdb/v2.0/reference/flux/stdlib/pagerduty/endpoint.md
```diff
@@ -37,12 +37,12 @@ _**Data type:** String_
...
 
 ### mapFn
-A function that builds the record used to generate the POST request.
+({{< req >}}) A function that builds the record used to generate the POST request.
 Requires an `r` parameter.
 
 _**Data type:** Function_
 
-The returned record must include the following fields:
+The function must return a record that includes the following fields:
```
kelseiv, line 45: `mapFn` accepts a table row (`r`) and returns a record that must include the following fields:

content/influxdb/v2.0/reference/flux/stdlib/slack/endpoint.md
```diff
@@ -54,18 +54,12 @@ _**Data type:** String_
 The output function requires a `mapFn` parameter.
 
 ### mapFn
```
kelseiv, line 56: Are these mapFn updates applicable for the `http.endpoint`?

kelseiv (approved):

content/influxdb/v2.0/reference/flux/stdlib/monitor/notify.md
sanderson: I added a PagerDuty example, but they're all essentially the same. The only thing that would change between them is the endpoint configuration.

content/influxdb/v2.0/reference/flux/stdlib/slack/endpoint.md
sanderson: Yes, updated that as well.

PR #2005, opened 2020-12-18
fix: corrected file name

oori:
File name was wrong:   `influxdb-2.x.x` --> `influxdb2-2.x.x`


sanderson (commented):
Thanks @oori ! Somebody else beat you to this change, but thank you for the contribution!

PR #2017, opened 2020-12-25
Fix minor typo in "Group keys"

haxtibal:
As the description says.
> Every table has a group key – a list of columns ~~which~~ for which every row in the table has the same value.

- [x] Typos do not require signing the CLA
- [x] Rebased/mergeable


kelseiv (approved):

kelseiv:
thank you, @haxtibal

PR #2022, opened 2020-12-28
add redirects

kelseiv:
- Add redirects to resolve 404 errors. 
- For good measure, transferred originally reported [issue](https://github.com/influxdata/ui/issues/467) to fix link in UI (and address any other potentially related links).


pierwill (approved):
👍

PR #2030, opened 2021-01-06
Hide delete field example

### LINKED ISSUE #2025: OSS delete-with-predicate docs mention a feature that doesn't work
The docs say [here](https://docs.influxdata.com/influxdb/v2.0/reference/syntax/delete-predicate/#delete-points-with-a-specific-field) that delete-with-predicate based on `_field` is valid. A user [reported](https://github.com/influxdata/influxdb/issues/20399) that using the field doesn't actually work.

InfluxDB OSS v2.0.3

kelseiv:
Closes #2025

@danxmoran, our Cloud doc is set up to duplicate OSS- does this only affect OSS?

### DIFF
--- a/content/influxdb/v2.0/reference/syntax/delete-predicate.md
+++ b/content/influxdb/v2.0/reference/syntax/delete-predicate.md
@@ -58,13 +58,14 @@ The following will delete points in the `sensorData` measurement:
 _measurement="sensorData"
 ```
 
+<!--
 ### Delete points with a specific field
 The following will delete points with the `temperature` field:
 
 ```sql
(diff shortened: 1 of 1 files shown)


danxmoran:
@kelseiv I'm not sure. From what I've seen / been told this sounds like a core long-standing limitation of the TSM engine, but I haven't tried reproducing outside of OSS. @benbjohnson implemented the OSS version so he might know the state of things on the Cloud side.

danxmoran (approved):
Forgot about this, sorry! I feel safe assuming it doesn't work in Cloud or OSS

Issue #2035, opened 2021-01-06
Add support for HTTP sources in Sideloadnode


kelseiv:
Add support for HTTP sources in Sideloadnode.

https://docs.influxdata.com/kapacitor/v1.5/nodes/sideload_node/#source

add'l detail:

// .source may be one of: a file URI, a URL, or an endpoint name as configured in the Kapacitor configuration file as an httppost object.
//
// If  source is defined as a plain string and not a URI (file://,http://), the source wil interpreted as an endpoint defined in an [[httpost]] section in the Kapacitor configuration.
//
// A source defined as an HTTP URL or [[httppost]] endpoint will be loaded as an HTTP GET and loaded only once when a task is enabled, and then on subsequent calls to the /sideload/reload endpoint.

An HTTP source endpoint should return a JSON document where each property is a key name specified in the order statement and its value is an object with a set of key/value pairs.

// HTTP Source example:
//{
//	"host1" : {
// 	"cpu_threshold":98,
// 	"some_tag": "value",
// 	"disable": "False"
// 	},
// 	"some_tag_value": {
// 		"cpu_threshold": 97
// 		"another_tag": "value"
// 	}
// }

For example

stream
|from()
.database('telegraf')
.retentionPolicy('autogen')
.measurement('cpu')
.groupBy(*)
|sideload()
.source('http://localhost:5000/threshold/')
.order('{{ .host }}')
.field('cpu_threshold', 0.0)
.tag('disable','false')
.tag('custom', 'false')
|where(lambda: !bool("disable"))
|alert()
.details('')
.message('Sideload alert triggered on threshold {{ index .Fields "cpu_threshold" | printf "%0.2f" }} with value {{ index .Fields "usage_idle" | printf "%0.2f" }}')
.crit(lambda: "usage_idle" > float("cpu_threshold"))
.log('/alerts.log')

PR #2043, opened 2021-01-07
Add HTTP source examples to sideload

kelseiv:
Closes https://github.com/influxdata/docs-v2/issues/2035


content/kapacitor/v1.5/nodes/sideload_node.md
```diff
@@ -151,16 +153,67 @@ sideload.quiet()
...
+```js
+|sideload()
+  .source('host1')
+  .order('host/{{.host}}.yml', 'hostgroup/{{.hostgroup}}.yml')
+  .field('cpu_threshold', 0.0)
+  .tag('foo', 'unknown')
+```
 
+{{% note %}}
+A source defined as a **URL** or **endpoint** is loaded once as an HTTP GET when a task is enabled, and then on subsequent calls to the `/sideload/reload` endpoint.
```
sanderson, line 193:
```suggestion
Kapacitor loads **URL** or **endpoint** sources once using an HTTP `GET` requests when a task is enabled, and then on subsequent calls to the `/sideload/reload` endpoint.
```

content/kapacitor/v1.5/nodes/sideload_node.md
```diff
@@ -151,16 +153,67 @@ sideload.quiet()
...
+{{% /note %}}
 
+An HTTP source endpoint should return a JSON where each property is a key name specified in the order statement and its value is an object with a set of key/value pairs.
```
sanderson, line 196:
```suggestion
An HTTP source endpoint should return a JSON object where each property is a key name specified in the order statement and its value is an object with a set of key-value pairs.
```

We probably need to decide how we want to format "key-value." Is it "key/value" or "key-value". Either way, it should be the same in all the docs.

content/kapacitor/v1.5/nodes/sideload_node.md
```diff
@@ -151,16 +153,67 @@ sideload.quiet()
...
+"host2": {
+"cpu_threshold": 85,
+"disable": "true"
+},
+"host3": {
+"cpu_threshold": 56,
+"disable": "false"
+}
+}
+```
```
sanderson, line 214:
~~~suggestion
```js
{
    "host1": {
        "cpu_threshold": 99.9,
        "disable": "false"
    },
    "host2": {
        "cpu_threshold": 85,
        "disable": "true"
    },
    "host3": {
        "cpu_threshold": 56,
        "disable": "false"
    }
}
```
~~~

content/kapacitor/v1.5/nodes/sideload_node.md
```diff
@@ -151,16 +153,67 @@ sideload.quiet()
...
+
+<a class="top" href="javascript:document.getElementsByClassName('article-heading')[0].scrollIntoView();" title="top"><span class="icon arrow-up"></span></a>
```
sanderson, line 216: These don't work anymore and should probably be removed.

sanderson (changes requested):

content/kapacitor/v1.5/nodes/sideload_node.md
kelseiv: agree, key-value works

content/kapacitor/v1.5/nodes/sideload_node.md
kelseiv: Appears we have 3453 occurrences in 69 files

sanderson (approved):

content/kapacitor/v1.5/event_handlers/bigpanda.md
```diff
@@ -37,8 +37,10 @@ Authorization Bearer token for BigPanda REST API.
 
 #### `app-key`
 
+
 BigPanda integration App Key. To get your App Key, log in to your BigPanda account and select **Integrations** in the header toolbar, and then click **New Integration**.
 Select **Alerts REST API**, click **Integrate**, and then **Create an App Key**.
+>>>>>>> master
```
docmerlin, line 43: This looks like someone did a bad merge.

content/kapacitor/v1.5/nodes/sideload_node.md
```diff
@@ -151,16 +153,70 @@ sideload.quiet()
 
 ### Source
 
+<<<<<<< HEAD
+Define the source of the data. The following sources are supported:
+=======
 Source for the data, for example `file://` or `http://`.
+>>>>>>> master
```
docmerlin, line 160: Bad merge.

docmerlin (changes requested):
Looks like there was a bad merge, that wasn't cleaned up.  It needs to be cleaned up.

PR #2046, opened 2021-01-08
Ops genie description text

kelseiv:
Add new details configuration option to OpsGenie v2 event handler.


content/kapacitor/v1.5/event_handlers/opsgenie/v2.md
```diff
@@ -57,6 +58,9 @@ Valid values include:
   * `close` - Close the alert.
   * custom -  Use the `.RecoveryAction()` method to specify the `recovery_action` in a TICK script.
 
+#### `details`
+If `true`, all alerts use their event details as the OpsGenie alert description. If `false`, event details are encoded as description.
```
kelseiv, line 62: Perhaps we can clarify how event details are encoded as description (being that's the default)?

sanderson: Yeah, "event details are encoded as description" isn't clear at all. I'm not even sure what that means.

sanderson (commented):
What is the difference between "[using] event details as the OpsGenie alert description" and "event details are encoded as description"? I don't fully understand the difference in behavior.

content/kapacitor/v1.5/event_handlers/opsgenie/v2.md
docmerlin: @sanderson @kelseiv It means the details field, gets sent as the OpsGenie alert description field.

docmerlin (approved):

content/kapacitor/v1.5/event_handlers/opsgenie/v2.md
kelseiv: Thank you, @docmerlin, to be sure, we're describing this details field that we set to true or false, which in turn sends another details field as the OpsGenie alert description field? 

Want to be sure on the exact name/location of this field--would "event details field" or "alert details field" be more accurate? thanks!

docmerlin: With would be fine.  Alert might make more sense for some people? Shrug.

kelseiv: do we not use a specific term in our product?

PR #2099, opened 2021-01-27
Re-order LDAP instructions

### LINKED ISSUE #2021: wrong order of operation
You cannot do an ldap verify of your toml file before you restart influxdb-meta. The way the doc reads, you edit the conf files, test an auth via curl (is that supposed to be ldap? Because it is not configured yet). Create an ldap toml file, verify and push it, then restart services. You cannot verify it until  you restart At least the meta nodes. 

##### Relevant URLs
https://docs.influxdata.com/enterprise_influxdb/v1.8/administration/ldap/

influxdb enterprise 1.8.

pierwill:
Instruct users to restart the service before verifying LDAP configuration.

Fixes #2021

- [x] Rebased/mergeable


sanderson (approved):
LGTM

PR #2174, opened 2021-02-11
Add information about Telegraf processor/aggregator plugin behavior

sanderson:
Closes influxdata/DAR#174

- [x] Rebased/mergeable

### DIFF
--- a/content/telegraf/v1.16/concepts/aggregator_processor_plugins.md
+++ b/content/telegraf/v1.16/concepts/aggregator_processor_plugins.md
@@ -3,7 +3,7 @@ title: Telegraf aggregator and processor plugins
 description: Aggregator and processor plugins work between the input plugins and output plugins to aggregate and process metrics in Telegraf.
 menu:
   telegraf_1_16:
-    name: Aggregator and processor plugins
+    name: Aggregator & processor plugins
     weight: 20
     parent: Concepts
 ---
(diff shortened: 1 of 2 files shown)


content/telegraf/v1.16/concepts/aggregator_processor_plugins.md
```diff
@@ -43,7 +43,22 @@ Since many users will only care about their aggregates and not every single metr
...
-**NOTE** Since aggregator plugins only aggregate metrics within their periods,
-historical data is not supported. In other words, if your metric timestamp is more
-than `now() - period` in the past, it will not be aggregated. If this is a feature
-that you need, please comment on this [GitHub issue](https://github.com/influxdata/telegraf/issues/1992).
+{{% note %}}
+#### Aggregator plugins do not support historical data
+Since aggregator plugins only aggregate metrics within their periods,
+historical data is not supported.
+In other words, if your metric timestamp is more
+than `now() - period` in the past, it is not aggregated.
```
ssoroka, line 51: this doesn't seem to be the case anymore with the `grace` grace period argument, as it's now configurable

content/telegraf/v1.16/concepts/aggregator_processor_plugins.md
```diff
@@ -43,7 +43,22 @@ Since many users will only care about their aggregates and not every single metr
...
+
+For more information, see [influxdata/telegraf#1992](https://github.com/influxdata/telegraf/issues/1992).
+{{% /note %}}
+
+{{% note %}}
+#### Behavior of processors and aggregators when used together
+When using both aggregator and processor plugins in Telegraf v1.17, processor plugins
+process each batch of data and then pass the processed data to aggregator plugins.
```
ssoroka, line 59: technically processors don't work on batches of data, they process each metric as it comes.

content/telegraf/v1.16/concepts/aggregator_processor_plugins.md
```diff
@@ -43,7 +43,22 @@ Since many users will only care about their aggregates and not every single metr
...
+After aggregating points, processor plugins process the aggregated data again.
+This can have unintended consequences, such as executing mathematical operations twice.
```
ssoroka, line 61: would comment that this is normally not an issue, but users should make sure any custom processor scripts are repeatable without side-effects, and use namepass/namedrop to avoid issues if they can't make the script repeatable (idempotent).

ssoroka (changes requested):
left some comments that apply to both versions.

Thanks!

content/telegraf/v1.16/concepts/aggregator_processor_plugins.md
sanderson: Looks like that grace period was added in 1.17. Should I leave the note in this one (1.16), but take it out of the 1.17 doc?

content/telegraf/v1.16/concepts/aggregator_processor_plugins.md
sanderson: Got it. I'll update it.

PR #2203, opened 2021-02-24
Added note to leave out task options when creating new task

### LINKED ISSUE #2202: Leave out 'options task' assignment when creating a task through the UI
When creating a new task, users should not include the `options task` assignment in their code. The UI automatically adds this when they save the task. If they include it in their code, it will conflict with the injected `options task` assignment and result in an error similar to:

```
Failed to create new task: invalid options; Err: @8:1-13:2: option "task" reassigned: @8:1-13:2: option "task" reassigned
```

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2.0/process-data/manage-tasks/create-task/
- https://docs.influxdata.com/influxdb/cloud/process-data/manage-tasks/create-task/

sanderson:
Closes #2202

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud/process-data/manage-tasks/create-task.md
+++ b/content/influxdb/cloud/process-data/manage-tasks/create-task.md
@@ -53,6 +53,15 @@ The InfluxDB UI provides multiple ways to create a task:
    See [Task options](/influxdb/cloud/process-data/task-options) for detailed information about each option.
 5. Select a token to use from the **Token** dropdown.
 6. In the right panel, enter your task script.
+
+    {{% note %}}
+##### Leave out the options tasks assignment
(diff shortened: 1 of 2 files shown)


pierwill (approved):
Changes look 👍 . One question: is there a case where people might want to add task options upfront? Does leaving this info out result in default options? Seems like a strange UI wrinkle...

sanderson:
I think this is a UI wrinkle. Ideally the UI should detect the options assignment when creating a new task, but if the user includes an option block AND has specified options in the UI fields, which takes precedence? 🤷🏻‍♂️ 

If you create a task through the API or CLI, the task does need the options assignment. It's only when you create a _new_ task that these conflict.

Issue #1947, opened 2020-12-05
[2.0] working with Docker images


timhallinflux:
Folks picking up the Docker images are looking for additional guidance about how to consume the image and effectively work with it.

1) volume mounts -- folks want to persist their data outside the docker image.  Looks like everything resides within `/root/.influxdbv2`.  
2) creating a config file -- assuming the volume mounts are made, should they be able to drop a config.toml into the same directory and have the influxd process pick that up and read it.  Let's show an example.

timhallinflux:
Another request for help outlined here: https://www.orpiske.net/2020/12/using-go-and-influxdb-to-find-the-best-time-to-go-the-gym/

danxmoran:
@kelseiv I took a 1st stab at docs for the new DockerHub build, it's in review at docker-library/docs#1871

pierwill:
Checking in on this.

Wow, @danxmoran! A lot of great info in that PR! I've got the basics working with the `-v` flag in Docker. First thing I'll do to address @timhallinflux's item 1 above is to add a minimal example of persisting data outside the Docker contain in the Docker section of https://docs.influxdata.com/influxdb/v2.0/get-started/.

Next, we can make sure we have easy access to instructions for picking up a config file from the mounted volume. That might be a minimal example plus a link to `docker-library`.

PR #2205, opened 2021-02-24
Document using volume mounts with Docker image

pierwill:
Work toward #1947.

### DIFF
--- a/content/influxdb/v2.0/get-started.md
+++ b/content/influxdb/v2.0/get-started.md
@@ -315,30 +315,7 @@ docker run --name influxdb -p 8086:8086 quay.io/influxdb/influxdb:v2.0.4
 ```
 _To run InfluxDB in [detached mode](https://docs.docker.com/engine/reference/run/#detached-vs-foreground), include the `-d` flag in the `docker run` command._
 
-{{% note %}}
-#### InfluxDB "phone home"
-
-By default, InfluxDB sends telemetry data back to InfluxData.
-The [InfluxData telemetry](https://www.influxdata.com/telemetry) page provides
-information about what data is collected and how it is used.
-
-To opt-out of sending telemetry data back to InfluxData, include the
-`--reporting-disabled` flag when starting the InfluxDB container.
-
-```bash
-docker run -p 8086:8086 quay.io/influxdb/influxdb:v2.0.4 --reporting-disabled
-```
-{{% /note %}}
-
-### Console into the InfluxDB container (optional)
-
-To use the `influx` command line interface, console into the `influxdb` Docker container:
-
-```bash
-docker exec -it influxdb /bin/bash
-```
-
-### Persist data outside the InfluxDB container (optional)
+### Persist data outside the InfluxDB container
 
 1. Create a new directory in which to store your data, and change to that directory.
    ```sh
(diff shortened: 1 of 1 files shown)


content/influxdb/v2.0/get-started.md
```diff
@@ -330,14 +330,31 @@ docker run -p 8086:8086 quay.io/influxdb/influxdb:v2.0.4 --reporting-disabled
...
-### Console into the InfluxDB Container (Optional)
+### Console into the InfluxDB container (optional)
 
 To use the `influx` command line interface, console into the `influxdb` Docker container:
 
 ```bash
 docker exec -it influxdb /bin/bash
 ```
 
+### Persist data outside the InfluxDB container (optional)
```
sanderson, line 341: I don’t know that’d I’d even say this is optional. Mounting a local directory should be the default behavior we suggest. If they don’t specify a local mount, they will lose all their data any time the container restarts.

content/influxdb/v2.0/get-started.md
```diff
@@ -315,6 +315,32 @@ docker run --name influxdb -p 8086:8086 quay.io/influxdb/influxdb:v2.0.4
 ```
 _To run InfluxDB in [detached mode](https://docs.docker.com/engine/reference/run/#detached-vs-foreground), include the `-d` flag in the `docker run` command._
 
+### Persist data outside the InfluxDB container
+
+1. Create a new directory in which to store your data, and change to that directory.
+   ```sh
+   mkdir my-influxdb-docker-data
+   cd my-influxdb-docker-data/
+   ```
```
sanderson, line 324:
```suggestion
1. Create a new directory to store your data in and navigate into the directory.
   
   ```sh
   mkdir path/to/influxdb-docker-volume && cd $_
   ```
```

content/influxdb/v2.0/get-started.md
```diff
@@ -315,6 +315,32 @@ docker run --name influxdb -p 8086:8086 quay.io/influxdb/influxdb:v2.0.4
...
+2. From within your new data directory,
+   run the InfluxDB Docker container with the `-v` flag as in the example below.
+   This will persist the data from `/root/.influxdb2/` _inside_ the container to `my-influxdb-docker-data/` on the host file system.
+   ```sh
+   docker run \
+       --name influxdb \
+       -p 8086:8086 \
+       -v $PWD:/root/.influxdbv2 \
+       quay.io/influxdb/influxdb:v2.0.4 && \
+   ```
```
sanderson, line 334:
~~~suggestion
2. From within your new directory, run the InfluxDB Docker container with the `--volume` flag to
   persist data from `/root/.influxdb2/` _inside_ the container to the current working directory in
   the host file system.
   
   ```sh
   docker run \
       --name influxdb \
       -p 8086:8086 \
       --volume $PWD:/root/.influxdbv2 \
       quay.io/influxdb/influxdb:v2.0.4
   ```
~~~

sanderson (commented):
Some suggestions

content/influxdb/v2.0/get-started.md
pierwill: `path/to/influxdb-docker-volume` sounds to me like something that should already exist... How about `path/to/influxdb-docker-data-volume`?

sanderson (approved):

PR #2216, opened 2021-02-26
Fix missing config setting in Enterprise TLS guide

pierwill:
Closes https://github.com/influxdata/DAR/issues/177.

- [x] Rebased/mergeable


sanderson (approved):
Probably need to port this to 1.7 as well.

PR #2319, opened 2021-03-18
Add BigQuery output

noramullen1:


data/telegraf_plugins.yml
```diff
@@ -1892,6 +1892,13 @@ output:
     introduced: 0.1.7
     tags: [linux, macos, windows, messaging]
 
+  - name: BigQuery Google Cloud
```
sjwang90, line 1895: Just `BigQuery` might be fine. Or `Google BigQuery`

Issue #2286, opened 2021-03-12
Debian installation instructions highly insecure


charlemagnelasse:
The docs specify following instruction how to install influxdb on debian.

```
wget -qO- https://repos.influxdata.com/influxdb.key | sudo apt-key add -
source /etc/lsb-release
echo "deb https://repos.influxdata.com/${DISTRIB_ID,,} ${DISTRIB_CODENAME} stable" | sudo tee /etc/apt/sources.list.d/influxdb.list
```

This is highly insecure as it is adding non-Debian keys to the global trusted keyring - and will stop working in the near future. See

* https://michael-prokop.at/blog/2021/02/16/how-to-properly-use-3rd-party-debian-repository-signing-keys-with-apt/
* https://wiki.debian.org/DebianRepository/UseThirdParty
* https://manpages.debian.org/testing/apt/apt-key.8.en.html

Correct instructions are:

```
sudo mkdir -p /etc/apt/keyrings
wget -qO- https://repos.influxdata.com/influxdb.key | gpg --dearmor | sudo tee /etc/apt/keyrings/influxdb.gpg > /dev/null
source /etc/lsb-release
echo "deb [signed-by=/etc/apt/keyrings/influxdb.gpg] https://repos.influxdata.com/${DISTRIB_ID,,} ${DISTRIB_CODENAME} stable" | sudo tee /etc/apt/sources.list.d/influxdb.list
```

pierwill:
Thanks, @charlemagnelasse! Your instructions look good here. We'll dig into the articles you linked to a bit and get the docs updated.

pierwill:
Many thanks for this, @charlemagnelasse!

PR #2365, opened 2021-03-30
Debian/Ubuntu installation: use separate keyring for InfluxDB key

pierwill:
See #2286.

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/v1.8/introduction/install.md
+++ b/content/influxdb/v1.8/introduction/install.md
@@ -59,7 +59,7 @@ please see the
 Debian and Ubuntu users can install the latest stable version of InfluxDB using the
 `apt-get` package manager.
 
-For Ubuntu users, add the InfluxData repository with the following commands:
+For Ubuntu/Debian users, add the InfluxData repository with the following commands:
 
 {{< code-tabs-wrapper >}}
 {{% code-tabs %}}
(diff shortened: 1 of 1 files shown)


pierwill:
This PR edits the `apt` installation instructions for 1.8 only. If approved, I'll add commits or another PR to address Telegraf and a few previous DB versions.

darinfisher (changes requested):
1. The keyring storage directory should be `/etc/apt/trusted.gpg.d`.
This should be for Ubuntu and most any Debian base release.
[This is for at least `apt` version 1.6 or newer.](http://manpages.ubuntu.com/manpages/bionic/man8/apt-key.8.html)
> Note: Instead of using this command a keyring should be placed directly in the
`/etc/apt/trusted.gpg.d/` directory with a descriptive name and either "gpg" or "asc" as
file extension.

2. `tee` is used while storing the key and the secondary results are redirected to `/dev/null`.  This is not necessary.  `tee` is used when another copy of the results is needed. 
Try using: `wget -qO- https://repos.influxdata.com/influxdb.key | sudo gpg --dearmor > /etc/apt/trusted.gpg.d/influxdb.gpg`

3. Not all Debian releases contain the `/etc/lsb-release` file.  Using the `lsb_release` program should be more consistent.
`` export DISTRIB_ID=`lsb_release -si`;export DISTRIB_CODENAME=`lsb_release -sc` ``

darinfisher (approved):
Looks good.

PR #2441, opened 2021-04-20
Update query optimizations guide

### LINKED ISSUE #1042: Update query optimization doc for window/aggregate pushdowns
In an upcoming version of Flux, windowing and some aggregate operations now get pushed down storage, drastically improving performance. We need to update the [Optimize queries guide](https://v2.docs.influxdata.com/v2.0/query-data/optimize-queries/) with these changes.

- [ ] `window()` is now a pushdown function
- [ ] `aggregateWindow()` benefits from the window pushdown, but performance depends on which aggregate/selector you use to aggregate each window.
- [ ] List which aggregate functions get pushed down.

The operations in this PR are now enabled by default: https://github.com/influxdata/influxdb/pull/18829

##### Relevant URLs
- https://v2.docs.influxdata.com/v2.0/query-data/optimize-queries/

### LINKED ISSUE #2101: Add information about 'set()' vs 'map()' to query optimizations guide
Through performance testing, I found that `set()` and `experimental.set()` have significant performance advantages over `map()`. The takeaway is that if you’re going to modify data in a way that does NOT require existing row data, `set()` or `experimental.set()` are more performant. `map()` passes in the existing row data (`r`) into each row iteration, so you can do things like: `column: r.something + 2.0`. These types of transformations aren't possible with `set()` or `experimental.set()`

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2.0/query-data/optimize-queries/

### LINKED ISSUE #2150: Add information about inline processing to query optimization guide
We need to add information to the query optimization guide about performing inline string/mathematic processing inside of iterative functions, specifically `filter()`. These types of operations prevent `filter()` from being pushed down to storage and can cause significant performance issues. The better approach would be to perform the processing outside of `filter` to produce a scalar value which can then be used inside of `filter`. For an example, see influxdata/flux#3477.

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2.0/query-data/optimize-queries/
- https://docs.influxdata.com/influxdb/cloud/query-data/optimize-queries/

### LINKED ISSUE #2439: Update Flux pushdowns for Cloud and OSS 2.0.5
Our current lists of supported pushdown functions are outdated. We need to update them.

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2.0/query-data/optimize-queries/
- https://docs.influxdata.com/influxdb/cloud/query-data/optimize-queries/

sanderson:
Closes #2439
Closes #2150 
Closes #2101 
Closes #1042

- [x] Rebased/mergeable


content/influxdb/v2.0/query-data/optimize-queries.md
```diff
@@ -71,10 +128,60 @@ Consider their necessity in your data processing before using them:
...
+## Use set() instead of map() when possible
+[`set()`](/influxdb/v2.0/reference/flux/stdlib/built-in/transformations/set/),
+[`experimental.set()`](/influxdb/v2.0/reference/flux/stdlib/experimental/set/),
+and [`map`](/influxdb/v2.0/reference/flux/stdlib/built-in/transformations/map/)
+can each set columns value in data, however **set** functions have performance
+advantages over `map()`.
+
+Use the following guidelines to determine which to use:
+
+- If setting a column value to a predefined, static value, use `set()` or `experimetnal.set()`.
```
kelseiv, line 140:
```suggestion
- If setting a column value to a predefined, static value, use `set()` or `experimental.set()`.
```

content/influxdb/v2.0/query-data/optimize-queries.md
```diff
@@ -71,10 +128,60 @@ Consider their necessity in your data processing before using them:
...
+data
+  |> map(fn: (r) => ({ r with foo: r.bar }))
+```
+
 ## Balance time range and data precision
 To ensure queries are performant, balance the time range and the precision of your data.
 For example, if you query data stored every second and request six months worth of data,
 results would include ≈15.5 million points per series.  Depending on the number of series returned after `filter()`([cardinality](/influxdb/v2.0/reference/glossary/#series-cardinality)), this can quickly become many billions of points.
-Flux must store these points in memory to generate a response.  Use [pushdown functions](#pushdown-functions) to optimize how many points are stored in memory.
+Flux must store these points in memory to generate a response.  Use [pushdowns](#pushdown-functions-and-function-combinations) to optimize how many points are stored in memory.
```
kelseiv, line 165:
```suggestion
Flux must store these points in memory to generate a response. Use [pushdowns](#pushdown-functions-and-function-combinations) to optimize how many points are stored in memory.
```

kelseiv (approved):

PR #2442, opened 2021-04-20
Update April 2021 InfluxDB Cloud release notes

kelseiv:
@bpalani added stuff we talked about today. @russorat, please lmk if we should hold off this InfluxDB Cloud release note update to include annotations (or add later). We do periodic release notes to highlight recent Cloud updates.


bpalani (approved):
This is still missing annotations. Isn't that also GA?

russorat:
annotations are not released yet.

PR #2466, opened 2021-04-30
Contextualize URL selector feature callout text

### LINKED ISSUE #2460: Update “Cloud or OSS?” graphic
In our docs, under the gear symbol in the upper-right corner, we have red graphic that asks “Cloud or OSS?”…does it make sense to change this to say “Set URL” or “Set URL to customize examples”...or something more descriptive?

sanderson:
Closes #2460

This PR changes the text in the URL selector feature callout based on the current product section.

- If in InfluxDB Cloud docs: "Select your region"
- If in InfluxDB OSS docs: "Customize your URL"
- Everywhere else: "Cloud or OSS?"

---

- [x] Rebased/mergeable

### DIFF
--- a/layouts/partials/footer/feature-callout.html
+++ b/layouts/partials/footer/feature-callout.html
@@ -1,3 +1,18 @@
+{{ $scratch := newScratch }}
+{{ $productPathData := findRE "[^/]+.*?" .RelPermalink }}
+{{ $product := index $productPathData 0 }}
+{{ $currentVersion := index $productPathData 1 }}
+{{ $isCloud := eq "influxdb/cloud" (print $product "/" $currentVersion )}}
+{{ $isOSS := in (print $product "/" $currentVersion ) "influxdb/v"}}
+
+{{ $scratch.Set "urlCalloutText" "Cloud or OSS?" }}
+{{ if $isCloud }}
+  {{ $scratch.Set "urlCalloutText" "Select your region"}}
+{{ else if $isOSS }}
+  {{ $scratch.Set "urlCalloutText" "Customize your URL" }}
+{{ end }}
+{{ $urlCalloutText := $scratch.Get "urlCalloutText" }}
+
 <div class="feature-callout start-position" id="callout-url-selector">
-  <p>Cloud or OSS? <a href="#" class="close"><span class="icon-remove"></span></a></p>
+  <p>{{ $urlCalloutText }} <a href="#" class="close"><span class="icon-remove"></span></a></p>
 </div>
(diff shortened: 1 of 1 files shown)


kelseiv:
@sanderson, 
> For everywhere else, does it make sense to hide "Cloud or OSS?"? for other products, seems folks wouldn't customize docs for their Cloud or OSS endpoints?

>For InfluxDB Cloud and OSS docs, would selecting their region/OSS URL do anything other than updating examples throughout the docs with their endpoint? If not, thoughts on rephrasing to "Customize examples" for both?

sanderson:
> For everywhere else, does it make sense to hide "Cloud or OSS?"? for other products, seems folks wouldn't customize docs for their Cloud or OSS endpoints?

All of the other products can be configured to use Cloud or open source. This setting updates examples in the 1.x line, so all the URLs in codeblocks in the Telegraf, Chronograf, and Kapa docs get updated as well.

> For InfluxDB Cloud and OSS docs, would selecting their region/OSS URL do anything other than updating examples throughout the docs with their endpoint? If not, thoughts on rephrasing to "Customize examples" for both?

Right now, it doesn't do anything else. My only thought is "Customize examples" doesn't spur me to action like "Select your region" does. That same argument can't be made for the OSS docs though, since that's essentially what the message currently says 😄 .

kelseiv:
> > All of the other products can be configured to use Cloud or open source. This setting updates examples in the 1.x line, so all the URLs in codeblocks in the Telegraf, Chronograf, and Kapa docs get updated as well.
> 
>  Ah, I see what you mean. Although, do think it's cleaner/potentially less confusing to keep selector in context, given the endpoint is specific to InfluxDB Cloud or InfluxDB OSS. Provide the selector at the top-level in those docs, and then throughout the Telegraf/Chrono/Kapa docs, give the option to select the InfluxDB OSS or Cloud endpoint solely in context of examples (like we already do). I realize bubble goes away after selected, so prob not too invasive, just thought less is better.
>
>> Right now, it doesn't do anything else. My only thought is "Customize examples" doesn't spur me to action like "Select your region" does. That same argument can't be made for the OSS docs though, since that's essentially what the message currently says 😄 .
>
> I see what you mean. How about "Set endpoint for examples?"...trying to describe the purpose of the task and provide consistency in language acorss oss and clout to convey that both "selecting a region" and "customizing your OSS url" has the same end result.

kelseiv (approved):

sanderson:
Updated the following:

- Callout bubble only appears on the home page or while in the OSS or Cloud docs.
- Only two variations of the text:
    - Cloud: "Select your region"
    - Everywhere else: "Set your InfluxDB URL"

Issue #2385, opened 2021-03-25
2.0.4 docker image data folder different from docs.


PaschalisSk:
Unless I am missing something...

__Steps to reproduce:__
1. [Docs mention `/root/.influxdbv2` as data folder.](https://docs.influxdata.com/influxdb/v2.0/get-started/?t=Docker#persist-data-outside-the-influxdb-container)
2. `docker run influxdb:2.0.4` logs `service=storage-engine service=store path=/var/lib/influxdb2/engine/data`  and also `/root/.influxdbv2` is empty.

danxmoran:
`/root/.influxdbv2` was the path used by the `quay.io` builds published for early 2.0.x releases. Looks like the docs haven't been updated since we restored the DockerHub build.

sanderson:
We need to update this in the install/get-started guides. The filesystem layout is different depending on which Docker image you use (Dockerhub vs Quay.io). You can see the differences in the filesystem layout doc: 

https://docs.influxdata.com/influxdb/v2.0/reference/internals/file-system-layout/?t=Docker#file-system-layout

We should provide instructions for both in the install docs since they affect how mounts are configured in the Docker config, but we should also point users to the file system layout doc.

pierwill:
Closed in #2468.

Issue #2324, opened 2021-03-21
Getting started with docker mentions ".influxdbv2" directory but  should be ".influxdb2"


robinrottier:
Docker instructions tell you to map external drive to directory "`.influxdb2`"  but example code mentions  "`.influxdbv2`" for the container directory. My instance used ".influxdb2" so its the "infuxdbv2" thats wrong.

(although my preference woul dbe for the "v2" to indicate its a new version of the database, rather than influxdb2 sounds more like 2nd instance of the database)

https://docs.influxdata.com/influxdb/v2.0/get-started/?t=Docker

From within your new directory, run the InfluxDB Docker container with the --volume flag to persist data from /root/.influxdb2/ inside the container to the current working directory in the host file system.

docker run \
    --name influxdb \
    -p 8086:8086 \
    --volume $PWD:/root/.influxdbv2 \
    influxdb:2.0.4

sanderson:
Related to #2385

pierwill:
Closed in #2468.

PR #2468, opened 2021-04-30
Fix paths to docker volume mounts

### LINKED ISSUE #2463: Wrong volume path on get started with docker page
Hi, 
documentation regarding use of influxdb2.0 with docker/kubernetes is outdated. In [here](https://github.com/influxdata/docs-v2/blob/master/content/influxdb/v2.0/get-started.md#persist-data-outside-the-influxdb-container) it states that volume should be mounted to `/root/.influxdbv2` while in fact it should be mounted to `/var/lib/influxdb2` as is stated in official docker image documentation [here](https://github.com/docker-library/docs/blob/master/influxdb/README.md#upgrading-from-quayio-hosted-influxdb-2x-image). Also the kubernetes [example](https://github.com/influxdata/docs-v2/blob/master/static/downloads/influxdb-k8-minikube.yaml) has wrong volume path too.

I hope it will be fixed soon as it may confuse many people such as myself.

##### Relevant URLs
https://github.com/influxdata/docs-v2/blob/master/content/influxdb/v2.0/get-started.md#persist-data-outside-the-influxdb-container
https://github.com/influxdata/docs-v2/blob/master/static/downloads/influxdb-k8-minikube.yaml

pierwill:
Also bump InfluxDB version in k8s example.

Closes #2463, #2385, #2324.

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/v2.0/get-started.md
+++ b/content/influxdb/v2.0/get-started.md
@@ -419,7 +419,7 @@ _To run InfluxDB in [detached mode](https://docs.docker.com/engine/reference/run
    docker run \
        --name influxdb \
        -p 8086:8086 \
-       --volume $PWD:/root/.influxdbv2 \
+       --volume $PWD:/var/lib/influxdb2 \
        influxdb:2.0.6
    ```
 
--- a/static/downloads/influxdb-k8-minikube.yaml
+++ b/static/downloads/influxdb-k8-minikube.yaml
@@ -23,13 +23,13 @@ spec:
                 app: influxdb
         spec:
             containers:
-              - image: influxdb:2.0.4
+              - image: influxdb:2.0.6
                 name: influxdb
                 ports:
                   - containerPort: 8086
                     name: influxdb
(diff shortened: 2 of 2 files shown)


danxmoran (approved):
Nice

pierwill:
Should these be updated as well?

https://github.com/influxdata/docs-v2/blob/6e06dffbfcaebbb192f63c6d084ed881c47bef04/content/influxdb/v2.0/reference/internals/file-system-layout.md#L202-L211

danxmoran:
> Should these be updated as well?

No, those paths are accurate for the image hosted on `quay.io`.

At some point we should either stop advertising the Quay build, or put in the work to make work the same was as the DockerHub build...

danxmoran (approved):

sanderson:
@pierwill We should put a note in the Docker install instructions that mount paths are different depending on which Docker image you're using. We don't need to spell the differences out there, but we should provide a link to the Docker information in the file system layout doc.

PR #2476, opened 2021-05-04
Install dependencies as project dependencies from NPM repo

### LINKED ISSUE #2474: Set hugo-extended, postcss, etc. as Node devDependencies
Set hugo-extended, postcss, postcss-cli, and autoprefixer as devDependencies to be fetched from the npm repo.
This would:
1. make the install process simpler and more consistent across developers' machines,
2. localize the dependencies to this project instead of relying on global installs.

##### Cons
1. Adds another dependency, hugo-extended, which wraps the hugo binary.
2. ?

##### Relevant URLs
https://www.npmjs.com/package/hugo-extended

jstirnaman:
…endencies. Run npm install or yarn install. (#2474)

Closes #2474 

- [x] Rebased/mergeable


sanderson (changes requested):
`package.json` looks good. There are a few other places we need to update:

- Content in the `README.md` about running the docs locally.
- The deploy script in `circleci/config.yaml` (specifically lines 24-26)

README.md
```diff
@@ -20,37 +23,31 @@ including our GPG key, can be found at https://www.influxdata.com/how-to-report-
...
-    3. Run the following command to install the necessary tools:
+    3. Install dependencies:
 
-      ```sh
-      sudo yarn global add postcss-cli@8.3.0 autoprefixer@9.8.6
-      ```
+          ```
+          $ cd docs-v2
+          $ yarn install
+          ```
```
kelseiv, line 39: not sure if it makes sense to tell folks to delete package.json, but I get the following error (which may be trivial?) when testing:
```kseivert@MBP15-KSEIVERT docs-v2 % yarn install
yarn install v1.22.10
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
```

sanderson: This is trivial. This just means that you likely used npm to generate assets in your local repo. For fresh clones of the docs repo, this won't matter.

README.md
```diff
@@ -20,37 +23,31 @@ including our GPG key, can be found at https://www.influxdata.com/how-to-report-
...
 
-    ```
-    $ cd docs-v2/
-    $ hugo server
-    ```
+    From the same project directory, `docs-v2`, use NodeJS' `npx` command to start the Hugo server:
 
+      ```
+      $ cd docs-v2/
+      $ hugo server
```
sanderson, line 51: With hugo loaded as a node dependency, we should probably change this to use npx:


```suggestion
      $ npx hugo server
```

package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "devDependencies": {
+    "autoprefixer": "^10.2.5",
+    "hugo-extended": "^0.83.1",
+    "postcss": "^8.2.13",
+    "postcss-cli": "^8.3.1"
+  }
+}
```
sanderson, line 8: Since we no longer load dependencies globally in the deploy script, we need to also include `redoc-cli` as a dependency. Version `0.9.13` is the only version I've found to consistently generate the API docs.

```suggestion
{
  "devDependencies": {
    "autoprefixer": "^10.2.5",
    "hugo-extended": "^0.83.1",
    "postcss": "^8.2.13",
    "postcss-cli": "^8.3.1",
    "redoc-cli": "0.9.13"
  }
}
```

sanderson (changes requested):
A few suggestions in the existing changes. We also need to update the API generation script. I'll do that and push a commit.

README.md
```diff
@@ -20,37 +23,31 @@ including our GPG key, can be found at https://www.influxdata.com/how-to-report-
...
 
-    ```
-    $ cd docs-v2/
-    $ hugo server
-    ```
+    In the `docs-v2` directory, start the Hugo server:
 
-    View the docs at [localhost:1313](http://localhost:1313).
+      ```
+      $ cd docs-v2/
```
sanderson, line 50: Since they should already be in the docs-v2 directory, we can drop this command.

sanderson (approved):
One minor suggestion and then I think it's good. We'll need to be sure that the API docs generate correctly after we merge this change. Sometimes the generation breaks, but it doesn't throw an error.

api-docs/generate-api-docs.sh
```diff
@@ -41,7 +41,7 @@ weight: 304
 "
 
   # Use Redoc to generate the v2 API html
-  redoc-cli bundle -t template.hbs \
+  npx redoc-cli bundle -t template.hbs \
```
jstirnaman, line 44: @sanderson Should we give `/api-docs` it's own package.json as well? I also need to update the README with npx.

sanderson: That's not a bad idea. That way those who don't need to generate the API docs won't need to download all the deps to do it. We'll need to update the deploy steps `circleci/config.yml` as well.

jstirnaman: Done

jstirnaman:
> One minor suggestion and then I think it's good. We'll need to be sure that the API docs generate correctly after we merge this change. Sometimes the generation breaks, but it doesn't throw an error.

Please review again. Added the package.json to api-docs and updated the ci script accordingly.

sanderson (approved):
LGTM. Thanks for this @jstirnaman!

PR #2479, opened 2021-05-05
Enterprise: add logging environment variables to config doc

pierwill:
Closes https://github.com/influxdata/DAR/issues/199

- [x] Rebased/mergeable


andrew-influx (approved):
Looks good. Technically the customer states the OSS ENV variable to suppress logging did not work, but with the docs clarified we can take that on as technical test/re-test with an EAR as needed. Thanks!

Issue #2373, opened 2021-03-31
Flux query controls in 1.9.


lesam:
https://github.com/influxdata/influxdb/pull/21108
https://github.com/influxdata/influxdb/pull/21100
https://github.com/influxdata/influxdb/issues/17212

Add memory and concurrency limits to flux controller - see especially https://github.com/influxdata/influxdb/pull/21108#issue-604565893 for details on configuration options and new metrics exposed in `/debug/vars`. Note that since then, the metric name prefix was shortened from `query_control_` to `qc_`

This is for release in 1.9.

PR #2481, opened 2021-05-05
Document query controls in Enterprise 1.9

pierwill:
Document query controls in Enterprise 1.9

Closes https://github.com/influxdata/docs-v2/issues/2373

- [ ] Rebased/mergeable

### DIFF
--- a/content/enterprise_influxdb/v1.9/administration/config-data-nodes.md
+++ b/content/enterprise_influxdb/v1.9/administration/config-data-nodes.md
@@ -1253,13 +1253,22 @@ Environment variable: `INFLUXDB_TLS_MAX_VERSION`
     > higher CPU load as new shard digest files are created.
     > The added load will drop off after shard digests are completed for existing shards. -->
 
-## Flux controller
+## Query controls
 
 ### `[flux-controller]`
 
 #### `query-concurrency = 0`
-#### `query-initial-memory-bytes = 0`
-#### `query-max-memory-bytes = 0`
-#### `total-max-memory-bytes = 0`
-#### `query-queue-size = 0`
+Number of queries allowed to execute concurrently.
 
+#### `query-initial-memory-bytes = 0`
+Initial bytes of memory allocated for a query.
+
+#### `query-max-memory-bytes = 0`
+Maximum total bytes of memory allowed for queries.
+
+#### `total-max-memory-bytes = 0`
+Maximum total bytes of memory allowed for queries.
+
+#### `query-queue-size = 0`
+Maximum number of queries allowed in execution queue.
+When queue limit is reached, new queries are rejected.
--- a/content/enterprise_influxdb/v1.9/administration/config-data-nodes.md
+++ b/content/enterprise_influxdb/v1.9/administration/config-data-nodes.md
@@ -29,6 +29,7 @@ menu:
   * [UDP [udp]](#udp-settings)
   * [Continuous queries [continuous-queries]](#continuous-queries-settings)
   * [TLS [tls]](#tls-settings)
+  * [Flux controller [flux-controller]](#flux-controller)
 
 ## Data node configuration settings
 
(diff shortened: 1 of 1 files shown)


content/enterprise_influxdb/v1.9/administration/config-data-nodes.md
```diff
@@ -1251,3 +1252,23 @@ Environment variable: `INFLUXDB_TLS_MAX_VERSION`
     > If AE is left enabled while lowering your `max-fetch`, you will initially see
     > higher CPU load as new shard digest files are created.
     > The added load will drop off after shard digests are completed for existing shards. -->
+
+## Query controls
```
sanderson, line 1256:
```suggestion
## Flux query settings
```

content/enterprise_influxdb/v1.9/administration/config-data-nodes.md
```diff
@@ -1251,3 +1252,23 @@ Environment variable: `INFLUXDB_TLS_MAX_VERSION`
...
+
+#### `query-max-memory-bytes = 0`
+Maximum total bytes of memory allowed for queries.
+
+#### `total-max-memory-bytes = 0`
+Maximum total bytes of memory allowed for queries.
+
+#### `query-queue-size = 0`
+Maximum number of queries allowed in execution queue.
+When queue limit is reached, new queries are rejected.
```
sanderson, line 1274:
```suggestion
#### query-concurrency
Number of queries allowed to execute concurrently.
Default is `0`.

#### query-initial-memory-bytes
Initial bytes of memory allocated for a query.
Default is `0`.

#### query-max-memory-bytes
Maximum total bytes of memory allowed for queries.
Default is `0`.

#### total-max-memory-bytes
Maximum total bytes of memory allowed for queries.
Default is `0`.

#### query-queue-size
Maximum number of queries allowed in execution queue.
When queue limit is reached, new queries are rejected.
Default is `0`.
```

I know you're just following the convention in from the rest of the config file, but at some point we need to go back and update all these headings. They make the heading anchors unnecessarily complication and are prone to change overtime (when defaults are updated). I think we just update these for now and add the task of reformatting all these configuration headings to the backlog.

Other questions:
- Are these the actual defaults? What kind of behavior does the `0` setting result in for each of these?

sanderson (changes requested):

content/enterprise_influxdb/v1.9/administration/config-data-nodes.md
pierwill: @lesam Can you comment on these values? Are these actual defaults?

> Are these the actual defaults? What kind of behavior does the 0 setting result in for each of these?

lesam: https://github.com/influxdata/influxdb/pull/21108/files#r628465515 - see validation logic here:

query-queue-size and query-concurrency have to either both be zero or neither be zero (it doesn't make sense to have one limited and not the other).

lesam (approved):
lgtm

content/enterprise_influxdb/v1.9/administration/config-data-nodes.md
```diff
@@ -1251,3 +1252,33 @@ Environment variable: `INFLUXDB_TLS_MAX_VERSION`
...
+`0` means unlimited.
+Default is `0`.
+
+#### query-max-memory-bytes
+Maximum total bytes of memory allowed for queries.
+`0` means unlimited.
+Default is `0`.
+
+#### total-max-memory-bytes
+Maximum total bytes of memory allowed for queries.
```
lesam, line 1276: Might be good to clarify - this is the max memory usage across all running flux queries, while `query-max-memory-bytes` is the per-query maximum.

lesam (changes requested):

sanderson (approved):

Issue #2514, opened 2021-05-11
Update Manage InfluxDB Cloud billing page


kelseiv:
On the[ Manage InfluxDB Cloud billing](https://docs.influxdata.com/influxdb/cloud/account-management/data-usage) page, do the following:

- [x]   Explain what a “rate-limited request” is, what requests can be rate limited, and how to resolve a rate limit issue
- [x]  Update the following statement: “All rate-limited requests are rejected; including both read and write requests. Rate-limited requests are not queued.” to reflect that only writes are limited. 
- [x]  Add link to https://docs.influxdata.com/influxdb/cloud/write-data/best-practices/resolve-high-cardinality/
- [x]  Add note for PAYG accounts to contact Support to request raising cardinality limits (use same text that exists elsewhere).
- [x]  Update "Rate-limited HTTP response code" section to identify which requests this retry response works for
- [ ]  Review [Anais's blog](https://www.influxdata.com/blog/tldr-influxdb-tech-tips-monitoring-tasks-and-finding-the-source-of-runaway-cardinality/) and incorporate pertinent content in docs. 
- [ ]  Consider how best to consolidate content from the Account Management page and the Resolve high series cardinality page...does it make sense to use more of a format like Anais's blog post?

Request @rbetts and @bpalani review.

kelseiv:
Balaji approved updates to inaccurate Cloud account and limits content. Closing this priority issue and created [new issue ](https://github.com/influxdata/docs-v2/issues/2637)for incorporating blog content.

PR #2519, opened 2021-05-12
Cloud account updates

kelseiv:
Addresses some issues in https://github.com/influxdata/docs-v2/issues/2514

### DIFF
--- a/content/influxdb/cloud/account-management/billing.md
+++ b/content/influxdb/cloud/account-management/billing.md
@@ -16,6 +16,23 @@ menu:
 products: [cloud]
 ---
 
+To manage InfluxDB Cloud billing, review the following as needed:
+
+- [Upgrade to Usage-Based Plan](#upgrade-to-usage-based-plan)
+- [Access billing details](#access-billing-details):
+  - [Add or update your payment method]()
+  - [Add or update your contact information]()
+  - [Send notifications when usage exceeds an amount]()
+  - [View Usage-based Plan information]()
+  - [View Free Plan information]()
+- [Access billing details]():
+  - [Add or update your payment method]()
+  - [Add or update your contact inforrmation]()
+  - [Send notifications when usage exceeds an amount]()
+- [Review and resolve exceeded plan limits]()
+- [Billing cycle]()
+  - [Declined or late payments]()
+
 ## Upgrade to Usage-Based Plan
 
 1. Click **Upgrade Now** in the lower left corner of the {{< cloud-name "short" >}} user interface (UI).
@@ -83,24 +100,24 @@ On the **Billing page**, view your billing information, including:
 
 On the **Billing page**, view the total limits available for the Free Plan.
 
-## Exceeded rate limits
+## Review and resolve exceeded plan limits
 
-If you exceed your plan's [rate limits](/influxdb/cloud/account-management/pricing-plans/), {{< cloud-name >}} provides a notification in the {{< cloud-name "short" >}} user interface (UI) and adds a rate limit event to your **Usage** page for review.
+If you exceed your plan's [limits](/influxdb/cloud/account-management/pricing-plans/), {{< cloud-name >}} provides a notification in the {{< cloud-name "short" >}} user interface (UI).
 
-All rate-limited requests are rejected; including both read and write requests.
-_Rate-limited requests are **not** queued._
+If you exceed rate limits specifically, including write requests (Data In) or query requests, or exceed the series cardinality limit, InfluxDB adds a rate limit event to your **Usage** page for review, and write requests will be rejected. To start processing write requests again, do the following as needed:
 
-_To remove rate limits, [upgrade to a Usage-based Plan](#upgrade-to-usage-based-plan)._
+- **Usage plan**: To request higher rate limits, contact [InfluxData Support](mailto:support@influxdata.com).
+- **Series cardinality limits**: If you exceed the series cardinality limit, see how to [resolve high series cardinality](https://docs.influxdata.com/influxdb/v2.0/write-data/best-practices/resolve-high-cardinality/).
+- **Free plan**: To remove rate limits, [upgrade to a Usage-based Plan](#upgrade-to-usage-based-plan).
 
-#### Rate-limited HTTP response code
-
-When a request exceeds your plan's rate limit, the InfluxDB API returns the following response:
+<!-- #### Rate-limited HTTP response code
+When a request exceeds your plan's rate limit--either write requests (Data In) or query requests--the InfluxDB API returns the following response:
 
 ```
 HTTP 429 “Too Many Requests”
 Retry-After: xxx (seconds to wait before retrying the request)
 ```
-
+-->
 ## Billing cycle
 
 Billing occurs on the first day of the month for the previous month. For example, if you start the Usage-based Plan on September 15, you're billed on October 1 for your usage from September 15-30.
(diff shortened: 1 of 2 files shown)


content/influxdb/cloud/account-management/billing.md
```diff
@@ -83,24 +96,24 @@ On the **Billing page**, view your billing information, including:
...
-
-If you exceed your plan's [rate limits](/influxdb/cloud/account-management/pricing-plans/), {{< cloud-name >}} provides a notification in the {{< cloud-name "short" >}} user interface (UI) and adds a rate limit event to your **Usage** page for review.
+## Review and resolve plan limit overages
 
-All rate-limited requests are rejected; including both read and write requests.
-_Rate-limited requests are **not** queued._
+If you exceed your plan's [limits](/influxdb/cloud/account-management/pricing-plans/), you'll receive a notification in the {{< cloud-name "short" >}} user interface (UI).
 
-_To remove rate limits, [upgrade to a Usage-based Plan](#upgrade-to-usage-based-plan)._
+For rate limits specifically--including write requests (Data In) or query requests--and in cases where you exceed the series cardinality limit, InfluxDB adds a rate limit event to your **Usage** page for review, and begins to reject write requests. To start processing write requests again, do the following as needed:
```
kelseiv, line 103: Please verify this statement is accurate

content/influxdb/cloud/account-management/billing.md
```diff
@@ -83,24 +96,24 @@ On the **Billing page**, view your billing information, including:
...
 
-#### Rate-limited HTTP response code
+- **Usage-Based plan**: To request higher rate limits, contact [InfluxData Support](mailto:support@influxdata.com).
+- **Series cardinality limits**: If you exceed the series cardinality limit, see how to [resolve high series cardinality](https://docs.influxdata.com/influxdb/v2.0/write-data/best-practices/resolve-high-cardinality/).
+- **Free plan**: To remove rate limits, [upgrade to a Usage-based Plan](#upgrade-to-usage-based-plan).
 
-When a request exceeds your plan's rate limit, the InfluxDB API returns the following response:
+<!-- #### Rate-limited HTTP response code
```
kelseiv, line 109: Hid for now. Think we want to add a note, and then reference API docs ideally?

content/influxdb/cloud/account-management/pricing-plans.md
```diff
@@ -25,18 +25,28 @@ InfluxDB Cloud offers two pricing plans:
...
+### Free Plan limits
+
+The Free Plan includes the rate limits, data limits, and resource limits below.
+
+### Rate limits
+
+- **Data In:** .017MB per second
+- **Query:** 1MB per second
+
+### Data limits
```
kelseiv, line 40: Distinguished between "rate limits," "data limits," and "resource usage"...not married to any of the terminology, just a start to distinguish differences. Also, we were showing data in/query rate limits in spans of 5 minutes, so converted to per second rate. Discussed with @corriganjohn who vaguely recalled this original verbiage may have come from @rkramer9967? Prob best to align terms and verbiage cross our business communications, docs, and UI to the extent possible. Will stay tuned.

rkramer9967: The original verbiage pre-dates my tenure here at Influx. My understanding is the system truly monitors the rate limits in cumulative  5 minute increments as described on our website and so our documentation should continue to align to Data In 5MB/5 minutes and Query 300MB/5 minutes rather than converting these to their per second equivalents, i.e. the system does not limit to .017MB per second of ingest. I can ingest at a higher rate than .017mb/sec for a period of time subject to not exceeding 300MB in a 5 minute period.

kelseiv: Excellent feedback, thank you, @rkramer9967. Will leave as is if that's how the system currently works. Hoping for dev to confirm.

Background: Issue with docs surfaced yesterday as @corriganjohn was testing limits. Chatted with @corriganjohn who noticed potential UX confusion conveying input/out metrics in per/second some places and per 5 minutes here. Understand why we don't want to enforce limit per second.

Thoughts on expanding text to explicitly say: _Although InfluxDB supports tracking data ingest/query per/sec (or more precise level?), InfluxDB only checks ingest/query limits in five minute interrvals, and then enforces the limit if specifed ingest/query data exceeds MB in the previous 5 minutes._

corriganjohn: I don't think this is entirely accurate, based on testing. But not sure how detailed the explanation needs to be for the users. As far as I can tell (this needs development to confirm): 

- The system takes the assigned rate limit value (measured in per second) and multiplies that (by the number of seconds in 5 minutes) to arrive at a 'limit value'.
- Every 5 minutes the 'tracking' is cleared and restarted
- If the rate limit is hit at any point within the time window (eg it might only take 1 minute of intense processing) then all processes are stopped. No NEW processes can be triggered until the 'clock' is reset (after 5 minutes).
- Once the clock is reset, new in/out processes can be triggered (processes that were running when limit was hit do NOT restart. Note: this could potentially affect any Tasks that were set to run during the 'time out' window)

While we set the limit metric using 'per second', i think it makes sense to be consistent and only use the 'per 5 minute' calculated value for the users

content/influxdb/cloud/account-management/data-usage.md
```diff
@@ -20,9 +20,9 @@ right corner of your {{< cloud-name "short" >}} user interface (UI) and select *
 Find data usage information for the time frame specified in the drop-down at the top of the Usage page.
 
 - **Data In:** Total data in MB written to your {{< cloud-name "short" >}} instance.
-- **Data Out:** Total data in MB sent as responses to queries from your {{< cloud-name "short" >}} instance.
+- **Storage:** Total disk usage in gigabytes.
 - **Query Count:** Total number of individual query operations, which include queries, tasks (alerts, notifications) and Data Explorer activity.
-- **Storage Usage:** Total disk usage in gigabytes.
+- **Data Out:** Total data in MB sent as responses to queries from your {{< cloud-name "short" >}} instance.
 - **API Request Count:** The total number of query and write API requests received during the specified time frame.
```
bpalani, line 26: We do not have API Request Count anymore on our usage page.
image

content/influxdb/cloud/account-management/pricing-plans.md
```diff
@@ -53,38 +54,43 @@ The Usage-based Plan offers more flexibility and ensures you only pay for what y
...
+_To request higher rate limits, contact [InfluxData Support](mailto:support@influxdata.com)._
 
 - **Data In:** 300MB every 5 minutes
+- **Query:** 3000MB every 5 minutes
 - **Ingest batch size:** 50MB
-- **Queries:** 3000MB every 5 minutes
+- **Series cardinality:** 1,000,000 initial limit (higher limits available; [contact InfluxData Support](mailto:support@influxdata.com))
 - **Storage:** Unlimited retention
 {{% note %}}
 Increase the retention period up to 1 year by [updating a bucket’s retention period in the InfluxDB UI](/influxdb/cloud/organizations/buckets/update-bucket/#update-a-buckets-retention-period-in-the-influxdb-ui), or [set a custom retention period](/influxdb/cloud/organizations/buckets/update-bucket/#update-a-buckets-retention-period) using the [`influx` CLI](influxdb/cloud/reference/cli/influx/).
```
bpalani, line 78: We can increase bucket retention period to Unlimited or up to 1 year by updating the bucket's retention in the InfluxDB UI. For custom retention periods use CLI.

bpalani (approved):

PR #2539, opened 2021-05-13
update: mention that max-select-point frequency only checks every second.

### LINKED ISSUE #1498: mention that max-select-point only checks every 1 second
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v1.7/troubleshooting/query_management/#max-select-point

According to this pr from a few years ago, this only checks every second: https://github.com/influxdata/influxdb/pull/8480/files

We should mention that in the docs.

jstirnaman:
Closes #1498 

The code comment isn't clear, but I'm assuming here that the count, not the config, is checked every second. I haven't verified it.

### DIFF
--- a/content/influxdb/v1.4/troubleshooting/query_management.md
+++ b/content/influxdb/v1.4/troubleshooting/query_management.md
@@ -141,6 +141,8 @@ the following error:
 ```
 ERR: max number of points reached
 ```
+InfluxDB checks the point count every second (so queries exceeding the maximum aren’t immediately aborted).
+
 
 ### max-select-series
 
--- a/content/influxdb/v1.5/troubleshooting/query_management.md
+++ b/content/influxdb/v1.5/troubleshooting/query_management.md
@@ -142,6 +142,8 @@ the following error:
 ```
 ERR: max number of points reached
 ```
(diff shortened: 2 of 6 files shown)


content/influxdb/v1.4/troubleshooting/query_management.md
```diff
@@ -141,6 +141,8 @@ the following error:
 ```
 ERR: max number of points reached
 ```
+The point count will be checked every second, so queries will not be aborted immediately upon hitting the limit.
```
kelseiv, line 144: Thoughts on streamlining and weaving this line into text above? Something like:

The maximum number of points a SELECT statement can process (default 0 allows unlimited points). 

InfluxDB checks the point count every second (so queries aren’t immediately aborted). If your query exceeds `max-select-point`, InfluxDB stops the query and outputs the following error: `ERR: max number of points reached`

kelseiv (approved):

sanderson (approved):

content/influxdb/v1.4/troubleshooting/query_management.md
jstirnaman: @kelseiv updated and rebased

PR #2544, opened 2021-05-14
Update create.md

dahlinPL:
fixed syntax in example


kelseiv:
Thank you, @dahlinPL!

PR #2553, opened 2021-05-17
Flux group keys demo

sanderson:
This PR introduces in new group key demo that lets doc users visually get sense of how Flux group keys work. This is part of the new getting started with Flux docs in the flux-restructure line of work.

group-key-demo

- [x] Rebased/mergeable


content/flux/v0.x/get-started.md
```diff
@@ -0,0 +1,25 @@
...
+Flux is a **functional data scripting** language designed to unify the processes
+of querying, processing, analyzing, and acting on data into a single syntax.
+
+## Data model
+### Stream of tables
+### Table
+### Column
+### Row
+### Group key
+
```
kelseiv, line 20: Maybe Table grouping example would work best under the Table header? 

May be helpful to introduce/highlight example data and purpose along the lines of: 

This example includes the following data:
- `example` measurement
- `loc` tag with two values
-  sensorID` tag with two value
- `templ` and `hum` fields with multiples values. 

By default, this example demonstrates how data is modeled if no group key is applied. To filter data by group key, select one or more of the check boxes (keys?) to see resulting table records/rows._

Any reason not to limit sample two values per field? ....shorter to scan and consume.

content/flux/v0.x/get-started.md
```diff
@@ -0,0 +1,25 @@
...
+{{< flux/group-key-demo >}}
+
+## Data types
+### Basic types
+### Composite types
```
kelseiv, line 25: same as above

kelseiv (approved):

content/flux/v0.x/get-started.md
kelseiv: Any reason not to limit sample two values per field? ....shorter to scan and consume.

content/flux/v0.x/get-started.md
```diff
@@ -0,0 +1,25 @@
...
+
+Flux is a **functional data scripting** language designed to unify the processes
+of querying, processing, analyzing, and acting on data into a single syntax.
+
+## Data model
+### Stream of tables
+### Table
+### Column
+### Row
+### Group key
```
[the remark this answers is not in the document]
sanderson: Yeah, these are just placeholders.

content/flux/v0.x/get-started.md
sanderson: > Any reason not to limit sample data to two values per field? ....shorter to scan and consume.

That's the reason 😄

content/flux/v0.x/get-started.md
sanderson: Just place holders. These may end up being broken into multiple docs.

jstirnaman:
Nice visual! General question: we're not using any javascript processing/transpilation pipeline, right?

jstirnaman (approved):
Nice visual! General question: we're not using any javascript processing/transpilation pipeline, right?

sanderson:
@jstirnaman We're using Hugo's js/asset pipeline which essentially combines and minifies the js files. It's fairly rudimentary and was introduced in an earlier version of Hugo. I know they've made some updates to js processing, but I haven't taken the time to dig into the new functionality.

PR #2562, opened 2021-05-18
Update OSS and Cloud with latest-patch shortcode

sanderson:
- [x] Rebased/mergeable


pierwill (approved):
Works like a charm! 👍

PR #2583, opened 2021-05-24
Written demo data instructions

mnorriswalker:
Right now, the instructions to add a demo data set are only in the YouTube video. I added the steps, a note about how they impact free account rate limits, and about the types of resources instantiated.


noramullen1:
Hi @mnorriswalker,  demo data instructions live here: https://docs.influxdata.com/influxdb/cloud/write-data/#add-a-demo-data-bucket. They're not in the reference documentation because we keep procedures in the task-based docs, not reference.

mnorriswalker:
Ok I couldn't find that when searching. NBD. Closing now.

PR #2586, opened 2021-05-25
Recommend TLS 1.3 in Enterprise 1.9 docs

pierwill:
- [ ] Rebased/mergeable


sanderson (approved):
Just double check for links that point to the old anchor (#min-version--tls12).

_This is another reason why I hate including the default values in the headings 🤦🏻 . I don't know why this was ever started._

PR #2591, opened 2021-05-25
Add missing MQTT option parameter

### LINKED ISSUE #2538: MQTT retained option parameter missing
Have found that retain() option, as documented is missing the boolean flag. If just .retained() is used in a TICK script it gives an error about the number of parameters which is confusing

invalid TICKscript: line 21 char 10: too many arguments to set property Retained on *pipeline.MQTTHandler

Need to add TRUE so the statement is .retained(TRUE)

##### Relevant URLs
- _Provide relevant URLs (the doc page in question, project issues, community threads, etc.)_

https://docs.influxdata.com/kapacitor/v1.5/event_handlers/mqtt/

In the example TICK script

|alert()
  // ...
  .mqtt('topic-name')
    .brokerName('name')
    .qos(1)
    .retained()

The boolean flag is missing for retained - the last line should be 

   .retained(TRUE)

pierwill:
Closes #2538

- [x] Rebased/mergeable

### DIFF
--- a/content/kapacitor/v1.5/event_handlers/mqtt.md
+++ b/content/kapacitor/v1.5/event_handlers/mqtt.md
@@ -104,7 +104,7 @@ options:
   .mqtt('topic-name')
     .brokerName('name')
     .qos(1)
-    .retained(true)
+    .retained(TRUE)
 ```
 
 ## Using the MQTT event handler
--- a/content/kapacitor/v1.5/event_handlers/mqtt.md
+++ b/content/kapacitor/v1.5/event_handlers/mqtt.md
@@ -193,6 +193,7 @@ options:
   broker-name: 'alerts-broker'
   topic: 'alerts'
   qos: 2
+  retained: true
 ```
 
 Add the handler:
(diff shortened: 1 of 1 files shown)


sanderson (commented):
I might be wrong, but I know with other Kapacitor boolean params, the value has to be all caps (`TRUE`).

@docmerlin, can you confirm?

content/kapacitor/v1.5/event_handlers/mqtt.md
```diff
@@ -104,7 +104,7 @@ options:
   .mqtt('topic-name')
     .brokerName('name')
     .qos(1)
-    .retained()
+    .retained(true)
```
pierwill, line 107:
```suggestion
    .retained(TRUE)
```

pierwill:
Thanks Scott. That was the original suggestion in the issue. Updated.

sanderson (changes requested):
One last quick change and it should be good to go. You should also include the `retained` param in the handler file example beginning on line 188:

```yaml
# ...
options: 
  broker-name: 'alerts-broker' 
  topic: 'alerts' 
  qos: 2
  retained: true
```

sanderson (approved):
LGTM

PR #2604, opened 2021-05-26
Verify checksum for hugo

pierwill:
- [ ] Rebased/mergeable


sanderson (commented):
We actually won't need this script any more. With @jstirnaman's change, Hugo is loaded as a node dependency. Once that PR merges, this install script will be obsolete.

Issue #2392, opened 2021-04-07
Add 'tables' parameter to all Flux transformations


sanderson:
We need to go through and include the `tables` parameter in all Flux transformations. We excluded this parameter because it's assumed with transformations. However, it can be used to specify input data other than piped-forward data. We can just add the following to each transformation function:

~~~md
### tables
Input data.
Default is piped-forward data (`<-`).
~~~

sanderson:
I'll do this on the `flux-restructure` branch.

PR #2653, opened 2021-06-02
fix: minor updates from reviewing flux functions

russorat:
Closes #

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/v2.0/reference/flux/stdlib/experimental/last.md
+++ b/content/influxdb/v2.0/reference/flux/stdlib/experimental/last.md
@@ -31,12 +31,6 @@ experimental.last()
 `experimental.last()` drops empty tables.
 {{% /warn %}}
 
-## Parameters
-
-### tables
-Input data.
-Default is pipe-forwarded data.
-
 ## Examples
 
 #### Return the last non-null value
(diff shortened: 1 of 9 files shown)


content/influxdb/v2.0/reference/flux/stdlib/experimental/last.md
```diff
@@ -31,12 +31,6 @@ experimental.last()
 `experimental.last()` drops empty tables.
 {{% /warn %}}
 
-## Parameters
-
-### tables
-Input data.
-Default is pipe-forwarded data.
-
```
sanderson, line 39: These `tables` parameters need to remain. In fact, these will be added to all functions that take piped-forward data as input (see https://github.com/influxdata/docs-v2/issues/2392). There has been some confusion over this. Until recently, we've left this parameter undocumented, but have introduced it back in on newly documented functions.

The following scripts are functionally the same:

```javascript
import "array"
import "experimental"

data = array.from(rows: [{_value: 3.1},{_value:2,8},{_value:4.6}])

data
  |> experimental.last()
```

``` javascript
import "array"
import "experimental"

data = array.from(rows: [{_value: 3.1},{_value:2,8},{_value:4.6}])

experimental.last(tables: data)
```

sanderson (changes requested):
One update (see comment about `tables` parameter).

PR #2667, opened 2021-06-03
Fix typo in optimize-queries.md with `set()` function

Eric-Arellano:
The `set()` function uses the kwarg `value`, not `as`. It's also a normal function and doesn't use braces.


sanderson (approved):
Thanks @Eric-Arellano!

Eric-Arellano:
Thanks for the top-notch docs! This is an amazing project

PR #2654, opened 2021-06-03
Publish InfluxDB 2.0.7 docs

pierwill:
- [x] Tests pass (no build errors)
- [x] Rebased/mergeable

PR #2669, opened 2021-06-04
fix: update usage for `influx restore`

wbaker85:
Found this while responding to a community question about how to do a restore - the docs show a flag of `--input`, but that's not actually a flag (anymore?) and is now actually the argument.


pierwill:
Thanks, @wbaker85! These changes are (I think) already in https://github.com/influxdata/docs-v2/pull/2654. Would love your review there!

pierwill:
Closing for now.

wbaker85:
Ah, great minds think alike I guess 😄 Looks like it is taken care of in #2654, great!

PR #2702, opened 2021-06-14
Fix "Manage clusters" doc

### LINKED ISSUE #2648: InfluxDB Enterprise 1.9: Manage Clusters
enterprise_influxdb/v1.9/administration/cluster-commands/

Contents here are incorrect:

Current text:
```
Use the following tools to manage and interact with your InfluxDB Enterprise clusters:

To manage your clusters, use the influxd-ctl cluster management utility
To manage nodes, back up and restore data, and rebalance clusters, use the influx command line interface (CLI)
```

However, managing nodes, back up and restore, and rebalancing all require access to the meta nodes....and that is what `influxd-ctl` does!  So, these activities should be associated with `influxd-ctl`.

The `influx` CLI tool provides an interactive shell for the HTTP API associated with `influxd`. Use influx to write data (manually or from a file), query data interactively, and view query output in different formats.

pierwill:
Closes #2648

- [x] Rebased/mergeable

### DIFF
--- a/content/enterprise_influxdb/v1.9/administration/cluster-commands.md
+++ b/content/enterprise_influxdb/v1.9/administration/cluster-commands.md
@@ -14,8 +14,8 @@ menu:
 
 Use the following tools to manage and interact with your InfluxDB Enterprise clusters:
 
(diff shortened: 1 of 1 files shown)


timhallinflux (commented):
LGTM

PR #2763, opened 2021-06-21
Add range to NOAA examples (addresses #2720 )

### LINKED ISSUE #2720: The examples are not working any longer
The examples on the website no longer work.  
We should receive some results but currently, we are getting errors, i.e. for:  
```
from(bucket: "noaa")
  |> group()
  |> keep(columns: ["location"])
  |> unique(column: "location")
```
we are getting an error:
```
error in building plan while starting program: cannot submit unbounded read to "noaa"; try bounding 'from' with a call to 'range'
```
and when we provide a range
```
keep: schema collision: cannot group float and string types together
```

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2.0/query-data/common-queries/count_unique_values_for_column/

noramullen1:
Closes #2720 

@sanderson , a couple of the examples already had a range defined so I didn't add it to those, but did add it to the three without a range. Let me know if I got that wrong. Thanks!

### DIFF
--- a/content/influxdb/v2.0/query-data/common-queries/calculate_new_column.md
+++ b/content/influxdb/v2.0/query-data/common-queries/calculate_new_column.md
@@ -24,6 +24,7 @@ The following query:
 ```js
 from(bucket: "noaa")
   |> filter(fn: (r) => r._measurement == "average_temperature")
+  |> range(start: -30d)
   |> map(fn: (r) => ({r with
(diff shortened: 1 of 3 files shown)


sanderson (approved):

noramullen1:
@sanderson , this is the one where I was hoping you could take another look at the linked issue. Thanks!

sanderson (approved):

PR #2786, opened 2021-06-26
Fix hint box for `Why indexing matters`

aberkecz:
The hint box on the docs side is broken up into multiple pieces which decreases the readability.

- [ ] Rebased/mergeable


sanderson (changes requested):
Thanks @aberkecz, but we've moved away from using markdown block quotes to format these note blocks. Going through and updating all notes like these is in our backlog of to-dos. Could you wrap the block in [`note` shortcode](https://github.com/influxdata/docs-v2/blob/master/CONTRIBUTING.md#notes-and-warnings)? For example:

```md
{{% note %}}
... Content for the note here ...
{{% /note %}}

sanderson:
Closing in favor of #2960

PR #2792, opened 2021-06-29
values of query parameter `precision` for write api.

StoneYunZhao:
Reproduce:
```bash
curl -X POST "http://localhost:8086/write?db=1&precision=aa"
```

Got:
```json
{"error":"invalid precision \"aa\" (use n, u, ms, s, m or h)"}
```

So, available values for precision are "n, u, ms, s, m, h".

But the code `serveWriteV1` support both "ns" and "n". I think the code is not preciseness. Another evidence is from `serveWriteV2` which turn "ns" into "n".

- `serveWriteV1`: [handler.go, line 866](https://github.com/influxdata/influxdb/blob/master-1.x/services/httpd/handler.go#L866)
- `serveWriteV2`: [handler.go, line 843](https://github.com/influxdata/influxdb/blob/master-1.x/services/httpd/handler.go#L843)

- [ ] Rebased/mergeable


sanderson (approved):
Thanks @StoneYunZhao

PR #2800, opened 2021-06-30
Use hardcoded checksum for `s3deploy`

pierwill:
Also remove unused install script.


sanderson (changes requested):
I'm not sure I want to move in this direction. If we hardcode the hash, we have to manually update it anytime we want to bump the version of s3deploy. Is there a significant security risk if we dynamically fetch the hash from a trust source?

If we don't want to dynamically fetch the hash, let's define it where the s3deploy version is defined (in `circleci/config.yml`), then load it here as an environment variable.

pierwill:
We will have to update the hash manually, but unless the cost of doing so is prohibitive, there is a real security gain when reducing the number of trusted connections that have to be made dynamically.

Your suggestion of putting the checksum in the CircleCI yaml is great! That way there's only one file to change whenever we update this dependency, and it shouldn't be too painful.

sanderson (approved):

PR #2808, opened 2021-07-06
fix: example for updating a bucket retention period

wbaker85:
The example at https://docs.influxdata.com/influxdb/v2.0/organizations/buckets/update-bucket/#update-a-buckets-retention-period doesn't work quite right; for this command to work, a duration needs to be provided, and it isn't necessarily nanoseconds. I updated the documentation so that the example will work better.


pierwill (approved):

pierwill:
Thanks, @wbaker85!

PR #2813, opened 2021-07-07
Correct debug API call

### LINKED ISSUE #2812: Wrong profile API call for Enterprise 1.8.4 forward
This PR changed the method for taking profiles:

https://github.com/influxdata/influxdb/pull/19655

Previously correct method was documented at https://docs.influxdata.com/influxdb/v1.8/tools/api/#debugpprofall-http-endpoint

`curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all?cpu=true"`

The new correct method is to give the actual duration:

`curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all?cpu=30s"`

pierwill:
Closes #2812

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/v1.8/tools/api.md
+++ b/content/influxdb/v1.8/tools/api.md
@@ -234,9 +234,9 @@ After 30 seconds of data has been collected, the results are output to a file.
 100  237k    0  237k    0     0   8025      0 --:--:--  0:00:30 --:--:-- 79588
 ```
 
-{{% note }}
+{{% note %}}
 If using InfluxDB 1.8.3 or earlier, use `/debug/pprof/all?cpu=true`.
-{{% /note }}
+{{% /note %}}
 ### `/debug/requests` HTTP endpoint
 
 Use this endpoint to track HTTP client requests to the `/write` and `/query` endpoints.
--- a/content/influxdb/v1.8/tools/api.md
+++ b/content/influxdb/v1.8/tools/api.md
@@ -214,21 +214,21 @@ For more information about the Go `/net/http/pprof` package and the interactive 
 
 #### `/debug/pprof/all` HTTP endpoint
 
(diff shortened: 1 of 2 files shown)


content/influxdb/v1.8/tools/api.md
```diff
@@ -214,21 +214,21 @@ For more information about the Go `/net/http/pprof` package and the interactive
...
-curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all?cpu=true"
+curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all?cpu=30s"
 ```
 
->**Note:** When the `cpu=true` option is included, a CPU profile is generated for 30+ seconds.
-> If you're concerned about running a CPU profile (which only has a small, temporary impact on performance), then you can set `?cpu=false` or omit `?cpu=true` altogether.
-
-As the following example shows, the cURL output includes "Time Spent," the time elapsed (in  seconds). After 30 seconds of data has been collected, the results are output to a file.
+As the following example shows, the cURL output includes "Time Spent," the time elapsed (in  seconds).
+After 30 seconds of data has been collected, the results are output to a file.
```
lesam, line 228: Might want to add a note here that for 1.8.2 and earlier the correct command is the old one (`cpu=true`)

pierwill: 1.8.2 and earlier, or 1.8.3 and earlier?

lesam: I believe 1.8.3 and earlier - though nobody should be using 1.8.3 .

lesam (approved):

content/influxdb/v1.8/tools/api.md
pierwill: Ah, okay 👍

PR #2881, opened 2021-07-21
Fix handler yaml example

galuque:
The type of `arg` should be "list of string" and the current examples uses just a string. If used as is the following error will occur:

```
"failed to decode options into *alert.ExecHandlerConfig: 1 error(s) decoding:\n\n* 'args': source data must be an array or slice, got string"
```

Changed the handler examples from:

```yaml
id: handler-id
topic: topic-name
kind: exec
options:
  prog: /path/to/executable
  args: 'executable arguments'
```

to

```yaml
id: handler-id
topic: topic-name
kind: exec
options:
  prog: /path/to/executable
  args:
    - 'executable arguments'
```
In both 1.5 and 1.6 docs

- [ ] Rebased/mergeable

### DIFF
--- a/content/kapacitor/v1.5/event_handlers/exec.md
+++ b/content/kapacitor/v1.5/event_handlers/exec.md
@@ -29,7 +29,8 @@ topic: topic-name
 kind: exec
 options:
   prog: /path/to/executable
-  args: 'executable arguments'
+  args:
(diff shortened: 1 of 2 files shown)


sanderson (approved):
Thanks @galuque!

PR #2925, opened 2021-07-27
Correctly update InfluxDB URLs that don't have a trailing slash or URL path

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/assets/js/influxdb-url.js
+++ b/assets/js/influxdb-url.js
@@ -191,8 +191,8 @@ function updateUrls(prevUrls, newUrls) {
   .forEach(function (o) {
      if (o.replace != o.with) {
        $(elementSelector).each(function() {
-         // Lookbehind matches if o.replace is not preceded by [/.].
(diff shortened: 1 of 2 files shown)


sanderson:
@jstirnaman I know you're unavailable at the moment, but take a look at this when you're back. Any placholder URL that didn't end in a trailing slash or didn't have a path wasn't being updated. This PR fixes that, but it may have other unintended consequences as well.

I just made the lookbehind regex a little more specific to identify a protocol syntax pattern.

jstirnaman:
> @jstirnaman I know you're unavailable at the moment, but take a look at this when you're back. Any placholder URL that didn't end in a trailing slash or didn't have a path wasn't being updated. This PR fixes that, but it may have other unintended consequences as well.
> 
> I just made the lookbehind regex a little more specific to identify a protocol syntax pattern.

Thanks for fixing that; I was afraid my pattern might not be specific enough.

Issue #2803, opened 2021-07-01
Review Enterprise 1.9 auth docs


pierwill:
> @lesam @pierwill 
> Does it makes sense (and does it cover all bases) to replace the entire "User management" section here with a link to [Set up users with fine-grained authorization](https://docs.influxdata.com/enterprise_influxdb/v1.9/guides/fine-grained-authorization/)?
If so, let's do that, and then publish. If that doesn't cover all bases, let's create another doc issue to track this change specifically, and push out this PR, which addresses many broken links and starts folks on the auth path.

_Originally posted by @kelseiv in https://github.com/influxdata/docs-v2/pull/2778#discussion_r662517992_

timhallinflux:
I think we need to re-organize the security sections of the docs under Administration.  There are a number of submenu items that I believe we need to collect together.  @kelseiv and I were looking at this earlier today.

timhallinflux:
no... it does not cover all the bases....

If I break this down a bit more --- looking at what you can do with OSS and what you can do with Enterprise:

1. OSS supports the ability to setup users (READ/WRITE/ALL/ADMIN) permissions.  If you do nothing more, you have that permission for all DBs within the instance.

2. OSS ALSO allows for setting database level grants.  Meaning, within an instance, you can grant/deny access to the various databases within that instance (i.e. not a blanket grant....like in no. 1)

This is generally covered here :https://docs.influxdata.com/influxdb/v1.8/administration/authentication_and_authorization/#user-types-and-privileges

It appears that the only way to maintain these today is via InfluxQL commands.  This can be issued via the CLI (interactive session) or the Query API or via Chronograf.  However, Chronograf has specific UI pages which supports No. 1, but if you want to do No 2....you will need to submit individual InfluxQL queries through the Data Explorer (in the same way that you would use the CLI or Query API).

3. Enterprise shifts the number of user permission from 4 to 16.   These are covered here: https://docs.influxdata.com/enterprise_influxdb/v1.9/features/users/

These permissions can be set via the API and there are specific API calls to allow for User Management within InfluxDB Enterprise.  These are documented here: https://archive.docs.influxdata.com/enterprise_kapacitor/v1.5/administration/auth/#user-and-privilege-management-over-the-influxd-meta-api

Chronograf can again be used to set the individual user permissions when connected to InfluxDB Enterprise.

[...]

PR #2936, opened 2021-07-29
Add note about managing Enterprise users in Chronograf

sanderson:
Closes influxdata/DAR#217

- [x] Rebased/mergeable

### DIFF
--- a/content/enterprise_influxdb/v1.9/administration/authentication_and_authorization.md
+++ b/content/enterprise_influxdb/v1.9/administration/authentication_and_authorization.md
@@ -7,6 +7,9 @@ menu:
     name: Manage authentication and authorization
     weight: 30
     parent: Administration
+related:
+  - /enterprise_influxdb/v1.9/guides/fine-grained-authorization/
+  - /{{< latest "chronograf" >}}/administration/managing-influxdb-users/
 ---
 
 This document covers setting up and managing authentication and authorization in InfluxDB Enterprise.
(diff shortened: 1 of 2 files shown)


timhallinflux (commented):
So there are a few things that I think we are silent on that probably need to surface and this is likely to result in a number of other changes.

Specifically this page: https://docs.influxdata.com/enterprise_influxdb/v1.9/administration/authentication_and_authorization/#user-types-and-privileges  is actually geared towards OSS...and not Enterprise.

The Enterprise page: https://docs.influxdata.com/enterprise_influxdb/v1.9/features/users/
enumerates the broader user roles...but is mostly silent on the best practices for accessing/setting these roles.  You cannot use the CLI!  The CLI only understand READ/WRITE/ALL/ADMIN....

Unfortunately, there are some challenges here in terms of how best to document this and that you simply cannot achieve each and every use case through a single tool.

It may be worth documenting the use case/intention and then pointing at the appropriate tools/APIs that are required to fulfill the intent.

We also should layer in a scenario -- which is about user roles/permissions when you "upgrade" from OSS 1.x to Enterprise 1.x.

Examples
Use Case: CRUD for users and permissions with InfluxDB Enterprise
The Enterprise API should be used -- against the meta nodes.
Chronograf may also be used.

What you can't you do?  You cannot specify database level permissions(grants) for users via Chronograf.

Where issues potentially arise? Using the CLI to change the user's role results in?? needs some testing/validation

Use Case: Apply FGA 
API only

etc.

sanderson:
Holding off and will revamp this addition on my next support rotation (unless somebody else on support rotation wants to pick it up).

pierwill:
Closing this in favor of ongoing work on Enterprise auth docs. See for instance: https://github.com/influxdata/docs-v2/issues/2803

PR #2937, opened 2021-07-29
Add file system layout to 1.8 and clarify shard index location

sanderson:
Closes influxdata/DAR#151

- [x] Rebased/mergeable


kelseiv (approved):

timhallinflux:
@sanderson is there a corresponding InfluxDB Enterprise version of this doc that we should create for 1.9.x?

sanderson:
@timhallinflux We can certainly add it for InfluxDB Enterprise. I assume the only OS we would need to document this for is Linux?

timhallinflux:
yep.  No Windows support in Enterprise.

PR #2975, opened 2021-08-04
Work with booleans

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/flux/v0.x/data-types/basic/boolean.md
+++ b/content/flux/v0.x/data-types/basic/boolean.md
@@ -21,7 +21,7 @@ A **boolean** type represents a truth value (`true` or `false`).
 - [Negate boolean values](#negate-boolean-values)
 
 ## Boolean syntax
-Boolean values are represented by the following:
+Boolean literals are represented by the following:
 
 ```js
 true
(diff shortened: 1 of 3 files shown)


pierwill (approved):

content/flux/v0.x/data-types/basic/boolean.md
```diff
@@ -2,11 +2,131 @@
...
+- **uint**: value must be `0` (false) or `1` (true).
+
+```js
+bool(v: "true")
+// Returns true
+
+bool(v: 0.0)
+// Returns false
+
+bool(v: uint(v: 1))
```
kelseiv, line 50: could this also be `bool(v: 1)`? If not, we may want to expand/clarify on the bool() funtion page

kelseiv (approved):

content/flux/v0.x/data-types/basic/boolean.md
sanderson: It can be, there is just no way to express a uint in raw syntax. I'll add an example for just straight ints.

sanderson, line 50:
```suggestion
bool(v: 0.0)
// Returns false

bool(v: 0)
// Returns false

bool(v: uint(v: 1))
```

PR #2982, opened 2021-08-06
Work with records

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/flux/v0.x/data-types/composite/record.md
+++ b/content/flux/v0.x/data-types/composite/record.md
@@ -26,8 +26,8 @@ Each property can have a different value type.
 A **record** literal contains a set of key-value pairs (properties) enclosed in curly brackets (`{}`).
 Properties are comma-delimitted.
 **Property keys must be strings** and can optionally be enclosed in double quotes (`"`).
-If a property key contains only numeric characters or any whitespace characters,
-it must be enclosed in double quotes.
+If a property key contains whitespace characters or only numeric characters,
+you must enclose the property key in double quotes.
 Property keys are associated to values by a colon (`:`).
 **Values can be any type**.
 
--- a/content/flux/v0.x/data-types/composite/record.md
+++ b/content/flux/v0.x/data-types/composite/record.md
@@ -11,3 +11,151 @@ menu:
 weight: 201
 flux/v0.x/tags: ["composite types", "data types"]
 ---
+
+A **record** type is a set of key-value pairs (also known as properties).
+Keys (also known as labels) are strings.
+Values can be any data type.
+Each property can have a different value type.
+
+###### On this page:
+- [Record syntax](#record-syntax)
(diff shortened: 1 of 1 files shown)


content/flux/v0.x/data-types/composite/record.md
```diff
@@ -10,4 +10,168 @@ menu:
     parent: Composite types
 weight: 201
 flux/v0.x/tags: ["composite types", "data types"]
----
\ No newline at end of file
+---
+
+A **record** type is a set of key-value pairs (also known as properties).
```
kelseiv, line 15: Should we define properties in our glossary (are we using this term throughout the docs in the same way?). 
Noticed our definition of field refers to "the key-value pair in InfluxDB’s data structure that records metadata and the actual data value." 

Would it be accurate to say a record is a set of fields (or tags?) also known as properties? If not, maybe helpful to clarify this distinction as appropriate throughout.

content/flux/v0.x/data-types/composite/record.md
```diff
@@ -10,4 +10,168 @@ menu:
...
+- [Record syntax](#record-syntax)
+- [Reference values in a record](#reference-values-in-a-record)
+- [Operate on records](#operate-on-records)
+
+## Record syntax
+A **record** literal contains a set of key-value pairs (properties) enclosed in curly brackets (`{}`).
+Properties are comma-delimitted.
+**Property keys must be strings** and can optionally be enclosed in double quotes (`"`).
+If a property key contains only numeric characters or any whitespace characters,
+it must be enclosed in double quotes.
```
kelseiv, line 30:
```suggestion
If a property key contains only numeric characters and whitespace characters,
you must enclose the property key in double quotes.
```

content/flux/v0.x/data-types/composite/record.md
```diff
@@ -10,4 +10,168 @@ menu:
...
+  "Company Name": "ACME",
+  "Street Address": "123 Main St.",
+  id: 1123445
+}
+
+c["Company Name"]
+// Returns ACME
+
+c["id"]
+// Returns 1123445
```
kelseiv, line 87: did you mean to use Street Address as and example....or was this intentional to show you can get the id with bracket or dot notation?

kelseiv (approved):

content/flux/v0.x/data-types/composite/record.md
sanderson: No, the term "properties" is specific to records. The terms "fields" or "tags" don't really have meaning within the Flux language. They are concepts tied to the InfluxDB storage engine.

content/flux/v0.x/data-types/composite/record.md
sanderson: Both :). It was to show how to format a key that has a space it it. You have to wrap it in quotes.

PR #2994, opened 2021-08-09
Add OSS-Enterprise migration guide to 1.9

### LINKED ISSUE #2993: Migrate from OSS to InfluxDB Enterprise is missing in 1.9 doc set
Looks like we lost this guide in the shift to 1.9

##### Relevant URLs
https://docs.influxdata.com/enterprise_influxdb/v1.8/guides/migration/

It is referenced here: https://docs.influxdata.com/enterprise_influxdb/v1.9/features/clustering-features/#oss-conversion
but results in a 404.

pierwill:
Closes #2993.

This adds the migration guide present in Enterprise 1.8 docs to 1.9.

However, I'm not sure this guide can work as a drop in replacement. This might be the reason it wasn't copied over during the 1.9 release. For instance, the requirement that "Your **OSS and Enterprise version is the same**, for example, InfluxDB 1.8 and InfluxDB Enterprise 1.8" on line 20. Does this still hold?

- [x] Rebased/mergeable


content/enterprise_influxdb/v1.9/guides/migration.md
```diff
@@ -0,0 +1,330 @@
...
+    parent: Guides
+---
+
+Migrate a running instance of InfluxDB open source (OSS) to an InfluxDB Enterprise cluster.
+
+## Prerequisites
+
+- An InfluxDB OSS instance running **InfluxDB 1.7.10 or later**.
+- An InfluxDB Enterprise cluster running **InfluxDB Enterprise 1.7.10 or later**
+- Your **OSS and Enterprise version is the same**, for example, InfluxDB 1.8 and InfluxDB Enterprise 1.8.
```
timhallinflux, line 20: I'm unclear why An InfluxDB Enterprise cluster running 1.7.10 or later is a pre-requisite.  I believe the instructions have the user install and setup the InfluxDB Enterprise meta nodes.  However, I think these instructions are kind of confusing.  I believe we should have two separate guides.... one for upgrading OSS to Enterprise (where the Enterprise setup is "fresh") and another for adding an OSS node to an existing InfluxDB Enterprise installation (some what more rare...but there are other considerations when doing this -- like rebalancing).  

The version numbers between OSS and Enterprise will no longer be the same.  So, we can suggest that the user ensure that they are running on at least 1.8.x and preferably 1.8.latest.
InfluxDB Enterprise will be 1.9 (which is the current release).

Also, after the upgrade --- I think the rebalancing steps are potentially confusing?  different steps for a 2 data node cluster vs. a 4...   (there isn't much rebalancing to do for a 2 node cluster other than -> 1. Use the [ALTER RETENTION POLICY](/influxdb/v1.9/query_language/manage-database/#modify-retention-policies-with-alter-retention-policy) statement to increase the [replication factor](/enterprise_influxdb/v1.9/concepts/glossary/#replication-factor) on all existing retention polices to the number of data nodes in your cluster._

timhallinflux (changes requested):

pierwill:
Doc has been edited: cleaned up formatting, fixed links, etc.

Changed the instructions saying version numbers must match to say that users should upgrade to latest versions.

content/enterprise_influxdb/v1.9/guides/migration.md
```diff
@@ -0,0 +1,270 @@
...
+   For more information, see [`-export`](/enterprise_influxdb/v1.9/tools/influx_inspect#export).
+
+7. [Import data into Enterprise](/enterprise_influxdb/v1.9/administration/backup-and-restore/#importing-data).
+
+8. Verify data is successfully migrated to your Enterprise cluster. See:
+   - [Query data with the InfluxDB API](/enterprise_influxdb/v1.9/guides/query_data/)
+   - [View data in Chronograf](/{{< latest "chronograf" >}}/)
+
+9. Follow [Stop writes and remove OSS](#stop-writes-and-remove-oss) below.
+
```
kelseiv, line 72: @pierwill lets pull up content from "Stop writes and remove OSS" out of the hidden section

content/enterprise_influxdb/v1.9/guides/migration.md
```diff
@@ -0,0 +1,318 @@
...
+```
+    {{% /code-tab-content %}}
+    {{% code-tab-content %}}
+```bash
+sudo systemctl stop influxdb
+```
+    {{% /code-tab-content %}}
+    {{< /code-tabs-wrapper >}}
+
+    Double check that the service is stopped.
```
kelseiv, line 95:
```suggestion
    3. Double check that the service is stopped.
```

content/enterprise_influxdb/v1.9/guides/migration.md
```diff
@@ -0,0 +1,318 @@
...
+    The following command should return nothing:
+
+    ```bash
+    ps ax | grep influxd
+    ```
+
+3. Remove the InfluxDB OSS package.
```
kelseiv, line 102:
```suggestion
4. Remove the InfluxDB OSS package.
```

kelseiv (approved):

PR #3051, opened 2021-08-22
updated the email docs

smzgit:
single & double quotes in tickscript hold different meanings. In cpu_alert.tick tickscript 'usage_idle' must be enclosed within double quotes,ie., "usage_idle", otherwise kapacitor throws error -> enabling task cpu_alert: Failed to compile stateful expression for crit: mismatched type to binary operator. got string < int. see bool(), int(), float(), string(), duration()

Closes #

- [ ] Rebased/mergeable

### DIFF
--- a/content/kapacitor/v1.6/event_handlers/email.md
+++ b/content/kapacitor/v1.6/event_handlers/email.md
@@ -156,7 +156,7 @@ stream
   |from()
     .measurement('cpu')
   |alert()
-    .crit(lambda: "usage_idle" < 10)
+    .crit(lambda: int("usage_idle") < 10)
     .message('Hey, check your CPU')
     .topic('cpu')
 ```
--- a/content/kapacitor/v1.6/event_handlers/email.md
+++ b/content/kapacitor/v1.6/event_handlers/email.md
@@ -134,7 +134,7 @@ stream
   |from()
     .measurement('cpu')
   |alert()
-    .crit(lambda: 'usage_idle' < 10)
+    .crit(lambda: "usage_idle" < 10)
     .message('Hey, check your CPU')
     .email()
       .to('oncall1@mydomain.com')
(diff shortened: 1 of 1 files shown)


smzgit:
@jstirnaman can you please review this pull request? Thanks!

smzgit:
@jstirnaman  yeah exactly! Actually when i tried the same example without casting then it didn't work & throws error : mismatched type to binary operator. got string < int. 
And in another example - i found that same field was casted & it worked, i thought its some documentation mistake thats why i made the PR.

content/kapacitor/v1.6/event_handlers/email.md
```diff
@@ -156,7 +156,7 @@ stream
   |from()
     .measurement('cpu')
   |alert()
-    .crit(lambda: "usage_idle" < 10)
+    .crit(lambda: int("usage_idle") < 10)
```
sanderson, line 159:
```suggestion
    .crit(lambda: "usage_idle" < 10)
```

sanderson (changes requested):
Thanks for this @smzgit. I think we should remove the `int` casting from this example. It isn't necessary. In TICKscript, single quotes enclose string literals and double quotes are used to access field and tag values. The `mismatched type to binary operator. got string < int` error you got must have come from `'usage_idle` > 10` because that's comparing a string literal to an integer. You shouldn't get that error from `"usage_idle" < 10` unless the `usage_idle` field contains string values, which it definitely shouldn't.

smzgit:
Thanks @sanderson  for the suggestion, `int()` casting was required there, applied the suggested change.

jstirnaman (approved):

jstirnaman:
LGTM!

sanderson (approved):

PR #3058, opened 2021-08-25
Add `max-body-size` to Enterprise data node docs

### LINKED ISSUE #2548: max-body-size is not well-documented for 1.X Enterprise product.
A customer has asked if the `max-body-size` parameter in the OSS Influxdb also applies in the Enterprise version. It does, as the Enterprise version inherits that code from OSS, and the configuration to set the value, as well as the default value, are the same.

The customer has suggested adding to the Enterprise documentation to make this clear.

pierwill:
Add to Enterprise 1.8 and 1.9.

Closes #2548.

- [x] Rebased/mergeable


davidby-influx (approved):
LGTM

PR #3137, opened 2021-09-16
Mention cache maximum errors in config docs

### LINKED ISSUE #3094: add documentation of the relationship between cache-max-memory-size and cache-snapshot-memory-size
This issue asks about the connection between cache-max-memory-size and cache-snapshot-memory-size.

https://community.influxdata.com/t/is-there-a-any-recommendation-for-tuning-cache-memory-size/16102

We should add this to the page about optimizing writes (or maybe somewhere else). 

```
Perhaps your data rate is faster than you can snapshot, given the difference between snapshot memory size and cache maximum size?  As I understand the doc, the cache is written out out when it hits storage-cache-snapshot-memory-size .  So, you need a sufficient size difference between the max and the snapshot size, so that you don't overflow the cache while snapshotting.
```

##### Relevant URLs
https://docs.influxdata.com/influxdb/v2.0/write-data/best-practices/optimize-writes/

pierwill:
Part of #3094.

- [x] Rebased/mergeable


sanderson (commented):
I think we need a little more context for this error message. Increasing the this setting won't always solve the problem. We should add the answer provided in the community thread to https://docs.influxdata.com/influxdb/v1.8/troubleshooting/errors/

pierwill:
Merging this, but leaving the related issue open for further docs.

PR #3157, opened 2021-09-21
Add 'experimental/record' package

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/flux/v0.x/stdlib/experimental/record/_index.md
+++ b/content/flux/v0.x/stdlib/experimental/record/_index.md
@@ -39,8 +39,8 @@ record.any = {} // polymorphic record with unconstrained property types
 
 ### record.any
 Currently, default function parameter values constrain the parameter's type.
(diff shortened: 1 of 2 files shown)


content/flux/v0.x/stdlib/experimental/record/_index.md
```diff
@@ -0,0 +1,54 @@
...
+The `experimental/record` package provides the following constants:
+
+```js
+record.any = {} // polymorphic record with unconstrained property types
+```
+
+### record.any
+Currently, default function parameter values constrain the parameter's type.
+`record.any` is a polymorphic record value that can be used in
+as a default record value when input record property types are not known.
```
sanderson, line 43:
```suggestion
Currently, default function parameter values constrain the parameter's type.
`record.any` is a polymorphic record value that can be used as a default
record value when input record property types are not known.
```

Issue #3169, opened 2021-09-23
Add InfluxDB 2.0.9 release notes and new flux-log-enabled option


kelseiv:
https://github.com/influxdata/influxdb/blob/2.0/CHANGELOG_OLD.md#v209-unreleased

PR #3170, opened 2021-09-23
Add InfluxDB 2.0.9 release notes and new influxd flux-log-enabled option

kelseiv:
Closes https://github.com/influxdata/docs-v2/issues/3169

### DIFF
--- a/content/influxdb/v2.0/reference/config-options.md
+++ b/content/influxdb/v2.0/reference/config-options.md
@@ -87,7 +87,6 @@ tls-key = "/path/to/influxdb.key"
 Only non-default settings need to be defined in the configuration file.
 {{% /note %}}
 
-
 ## Configuration options
 
 To configure InfluxDB, use the following configuration options when starting the
@@ -97,6 +96,7 @@ To configure InfluxDB, use the following configuration options when starting the
 - [bolt-path](#bolt-path)
 - [e2e-testing](#e2e-testing)
 - [engine-path](#engine-path)
+- [flux-log-enabled](#flux-log-enabled)
 - [http-bind-address](#http-bind-address)
 - [http-idle-timeout](#http-idle-timeout)
 - [http-read-header-timeout](#http-read-header-timeout)
@@ -338,6 +338,59 @@ engine-path = "/users/user/.influxdbv2/engine"
 ```json
 {
   "engine-path": "/users/user/.influxdbv2/engine"
+}
+```
+{{% /code-tab-content %}}
+{{< /code-tabs-wrapper >}}
+
+---
+
+### flux-log-enabled
+
+Include option to show detailed logs for Flux queries, including the following:
+
+- query response size (`response_size`)
+- query errors (`err`)
+- query durations (`stat_total_duration`, `stat_compile_duration`, `stat_execute_duration`)
+- memory allocation (`stat_max_allocated`, `stat_total_allocated`)
+
+**Default:** `false`  
+
+| influxd flag         | Environment variable       | Configuration key |
+|:------------         |:--------------------       |:----------------- |
+| `--flux-log-enabled` | `INFLUXD_FLUX_LOG_ENABLED` | `flux-log-enabled`|
+
+###### influxd flag
+```sh
+influxd --flux-log-enabled=true
+```
+
+###### Environment variable
+```sh
+export INFLUXD_FLUX_LOG_ENABLED=true
+```
+
+###### Configuration file
+{{< code-tabs-wrapper >}}
+{{% code-tabs %}}
+[YAML](#)
+[TOML](#)
+[JSON](#)
+{{% /code-tabs %}}
+{{% code-tab-content %}}
+```yml
+flux-log-enabled: true
+```
+{{% /code-tab-content %}}
+{{% code-tab-content %}}
+```toml
+flux-log-enabled = "true"
+```
+{{% /code-tab-content %}}
+{{% code-tab-content %}}
+```json
+{
+  "flux-log-enabled": "true"
 }
 ```
 {{% /code-tab-content %}}
--- a/content/influxdb/v2.0/reference/release-notes/influxdb.md
+++ b/content/influxdb/v2.0/reference/release-notes/influxdb.md
@@ -8,6 +8,21 @@ menu:
 weight: 101
 ---
 
+## v2.0.9 [2021-09-23]
+
+### Features
+
+Add `--flux-log-enabled` option to [`influxd`](/influxdb/v2.0/reference/config-options/) to show detailed logs for Flux queries.
+
+### Bug fixes
+
+- Updating an inactive task no longer schedules it. Thanks @raffs!
+- Preserve comments in Flux queries when saving task definitions.
+- Resolve the compaction queue stats flutter.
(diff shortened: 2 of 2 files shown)


content/influxdb/v2.0/reference/config-options.md
```diff
@@ -345,6 +346,59 @@ engine-path = "/users/user/.influxdbv2/engine"
...
 ---
 
+### flux-log-enabled
+
+Include option to show detailed logs for Flux queries, including the following information:
+
+- query response size (`response_size`)
+- query errors (`err`)
+- query durations (`stat_total_duration`, `stat_compile_duration`, `stat_execute_duration`)
+- memory allocation (`stat_max_allocated`, `stat_total_allocated`)
```
kelseiv, line 356: @wbaker85 @russorat is these accurate and/or can we provide detail to clarify what's being logged? For example, a comprehensive list of errors? Which bits the duration and allocated are measuring?

content/influxdb/v2.0/reference/release-notes/influxdb.md
```diff
@@ -8,6 +8,21 @@ menu:
...
+### Features
+
+Add `--flux-log-enabled` option to [`influxd`](/influxdb/v2.0/reference/config-options/) to show detailed logs for Flux queries.
+
+### Bug fixes
+
+- Updating an inactive task no longer schedules it. Thanks @raffs!
+- Preserve comments in Flux queries when saving task definitions.
+- Resolve the compaction queue stats flutter.
+- Include `curl` as a dependency in `influxdb2` packages.
```
kelseiv, line 22: does it make sense to move line 22 under features?

wbaker85: This was more of a bug fix thing related to systemd so I think Bug Fixes is the right place for it...@codyshepherd might have thoughts as well.

content/influxdb/v2.0/reference/config-options.md
wbaker85: Here's what all the log fields mean:
- `compiler_type`: Compiler used for processing the query (will always be Flux)
- `response_size`: Size of the response, in bytes
- `query`: The textual representation of the query
- `err`: Any errors encountered while processing the query
- `stat_total_duration`: Total duration to process the query
- `stat_compile_duration`: Duration to compile the query
- `stat_execute_duration`: Duration to execute the query
- `stat_max_allocated`: Maximum amount of memory allocated while processing the query, in bytes
- `stat_total_allocated`: Total amount of memory allocated while processing the query, in bytes. This includes memory that was freed and then used again.

The list of potential errors would be vast I'd imagine, so I don't think we'd want to list them all here. I think it will be sufficient to describe it as just "any errors encountered" as I've got listed above.

wbaker85:
Is this the full set of release notes for 2.0.9? There are some other things to include if so.

We stopped updating the `CHANGELOG.md` file in VCS recently, and instead have the changelog automatically generated. There should be a link to download the full changelog here: https://github.com/influxdata/influxdb/blob/2.0/CHANGELOG.md. There are a number of other changes that will be in this release beyond what is in https://github.com/influxdata/influxdb/blob/2.0/CHANGELOG_OLD.md - the `CHANGELOG_OLD` file is just the last version before we started using the automated changelog.

kelseiv:
@wbaker85 Congrats on automating the changelog!

I initially landed [here](https://github.com/influxdata/influxdb/blob/2.0/CHANGELOG_OLD.md#v209-unreleased), so worked w that...if all of 2.0.9 changes aren't listed here, seems like we may want to update this list?  cc @dgnorton

I did notice the "old" and "new" change logs, I think w the new graphic format, the old format prob works best for docs...is there a way to retain that format moving forward?

content/influxdb/v2.0/reference/config-options.md
kelseiv: thank you, @wbaker85

kelseiv: @wbaker85 are the durations in seconds?

wbaker85:
I think we can probably come up with something for the old format moving forward, but it doesn't look like we have that right now. Are you able to get the full list of changes from https://dl.influxdata.com/platform/nightlies/2.0/CHANGELOG.md? That should have everything. We don't plan on updating the `CHANGELOG.md` file in the repo anymore.

kelseiv:
OK, that'd be super helpful, thank you @wbaker85.
Yes, got the full list of changes. 

To be sure, please verify changes look accurate (found some overlap between old and new changes phrased slightly different).

content/flux/v0.x/stdlib/influxdata/influxdb/cardinality.md
```diff
@@ -15,11 +15,7 @@ related:
 introduced: 0.92.0
 ---
 
-The `influxdb.cardinality()` function returns the [series cardinality](/{{< latest "influxdb" "v2" >}}/reference/glossary#series-cardinality) of data stored in InfluxDB Cloud.
-
-{{% cloud %}}
-**InfluxDB Cloud** supports the `influxdb.cardinality()` function, but **InfluxDB OSS does not**.
-{{% /cloud %}}
```
kelseiv, line 22: cc @sanderson

content/influxdb/v2.0/reference/release-notes/influxdb.md
```diff
@@ -8,18 +8,54 @@ menu:
...
+- Correctly filter requests to `/api/v2/authorizations` by `org` and `user` parameters.
+- Fix `X-Influxdb-Build` and `X-Influxdb-Version` response header at `/ping`.
+- Repair bad port dropping return value names.
+- Discard excessive errors over `DefaultMaxSavedErrors (100)` to prevent out-of-memory crashes.
+- Show message to set the Flux `content-type` when a query fails to parse as JSON.
+- Use consistent path separator in permission string representation.
+- Upgrade `golang.org/x/sys` to avoid panics on macs.
+- Ensure TSI index compacts log files that are old or too large.
```
kelseiv, line 42: do we have specifics re old and too large?

content/influxdb/v2.0/reference/release-notes/influxdb.md
```diff
@@ -8,6 +8,62 @@ menu:
...
+	- `ScraperResourceType`
+  - `SecretsResourceType`
+	- `LabelsResourceType`
+	- `ViewsResourceType`
+	- `DocumentsResourceType`
+	- `NotificationRuleResourceType`
+	- `NotificationEndpointResourceType`
+	- `ChecksResourceType`
+	- `DBRPResourceType`
+
```
kelseiv, line 46: @glinton @lesam @russorat, do we want to say anything more about why/how this allows the ui (and cli?) to generate an all-access token in a more uniform, and less brittle, way between cloud and oss?

glinton: it might be a good place to point out that an all-access token, in it's current form, is a snapshot of the resources at that point in time and is not evolutionary. meaning that if a resource type (generic bucket permission covers new instances of the same type) is added after creation, that "all-access" token will not include permissions to interact with that new resource.

PR #3188, opened 2021-09-29
feat: add custom processing of swagger specs before bundling to HTML

### LINKED ISSUE #3187: Fix server URLs for paths outside (above) the /api/v2 path
e.g. routes, ping, health have incorrect paths when rendered by Redoc. In the /openapi specs, some routes are explicitly set to `''` for consumption by downstream projects. Setting these to `/` works in Redoc, but breaks other consumers.

jstirnaman:
Closes #3187
Adds OpenAPI CLI and custom decorators for "top-level" API paths.
No substantive changes to v1 compatibility specs, just pulled the latest from /openapi repo.
Resolves conflicts with 2.0.9 branch.


api-docs/package.json
```diff
@@ -5,7 +5,10 @@
   "description": "InfluxDB API documentation",
   "license": "MIT",
   "dependencies": {
+    "@redocly/openapi-cli": "^1.0.0-beta.61",
```
sanderson, line 8: Where this is still in beta, we'll want to make sure that the generate process fails on error. I vaguely remember disabling exist on error because, at one point, redoc only exited with non-zero exit codes. I don't know if that's still the case. Either way, just something to watch out for.

sanderson (approved):
No changes. Comments are really just artifacts for me 😄  Solidify them in my brain.

api-docs/.redocly.yaml
```diff
@@ -0,0 +1,16 @@
+apiDefinitions:
+  cloud: ./cloud/swagger.yml
+  cloudV1Compat: ./cloud/swaggerV1Compat.yml
+  'v2.0': ./v2.0/swagger.yml
+  'v2.0V1Compat': ./v2.0/swaggerV1Compat.yml
```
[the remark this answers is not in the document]
jstirnaman: These aren't actually used in the generate script. I'm using the file path instead:
`npx openapi bundle $version/swagger.yml -o $version/swagger.yml`

api-docs/cloud/swaggerV1Compat.yml
```diff
@@ -21,15 +20,17 @@ paths:
...
+        - $ref: '#/components/parameters/TraceSpan'
+        - $ref: '#/components/parameters/AuthUserV1'
+        - $ref: '#/components/parameters/AuthPassV1'
         - in: query
           name: db
           schema:
             type: string
           required: true
-          description: The bucket to write to. If none exist a bucket will be created with a default 3 day retention policy.
+          description: >-
```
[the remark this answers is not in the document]
jstirnaman: [Apparently not](https://stackoverflow.com/a/21699210/1211900), but this was the first time I'd seen at as well.

api-docs/package.json
jstirnaman: Crossed my mind and forgot to test.

> redoc only exited with non-zero exit codes
I'm not sure I follow. You mean even when there wasn't an error?

I'm testing `set -e`. `openapi` fails and exits when given a bad path or missing reference in the JS. `redoc-cli` fails and exits when given a bad path in options.

sanderson: Yeah, if I remember correctly, success or failure, it would exit with a non-zero code. But this was a looong time ago.

PR #3226, opened 2021-10-07
Fix api docs dependencies

jstirnaman:
Addresses vulnerability report.
Removes deprecated npx dependency (the source of the vulnerabilities).
Simplifies installation and updating for docs maintainers by replacing redoc-cli in package file with redoc-cli version specifier in the npx command.

### DIFF
--- a/api-docs/README.md
+++ b/api-docs/README.md
@@ -33,7 +33,7 @@ In your terminal, run:
 npx --version
 ```
 
-If `npx` returns errors, [download](https://nodejs.org/en/) and run a recent version of the Node.JS installer for your OS.
+If `npx` returns errors, [download](https://nodejs.org/en/) and run a recent version of the Node.js installer for your OS.
 
 In your terminal, from the root of the docs repo, run:
 


api-docs/README.md
```diff
@@ -26,17 +26,20 @@ api-docs/
 Because the API documentation HTML is gitignored, you must manually generate it
 to view the API docs locally.
 
-In your terminal, from the root of the docs repo, run:
+Verify that you have a working `npx` (it's included with Node.JS).
```
sanderson, line 29: The registered trademark has `js` in lowercase.

```suggestion
Verify that you have a working `npx` (it's included with Node.js).
```

api-docs/README.md
```diff
@@ -26,17 +26,20 @@ api-docs/
...
+In your terminal, run:
 
 ```sh
+npx --version
+```
 
+If `npx` returns errors, [download](https://nodejs.org/en/) and run a recent version of the Node.JS installer for your OS.
```
sanderson, line 36:
```suggestion
If `npx` returns errors, [download](https://nodejs.org/en/) and run a recent version of the Node.js installer for your OS.
```

sanderson (approved):
Two minor changes and it's good to go. Thanks!

PR #3248, opened 2021-10-13
Edit InfluxDB Enterprise features page

### LINKED ISSUE #3080: Re-organize and expand cluster features overview with outline of security related features and clustering
https://docs.influxdata.com/enterprise_influxdb/v1.9/features/clustering-features/

Add section near the top of the page to outline and provide highlights for security related features which are unique to InfluxDB Enterprise.  These are not currently described within this section (which presumably is the section capturing an overview of everything that is unique to Enterprise or links to how things work within Enterprise (when the feature exists already within OSS).  Clustering is generally described elsewhere, but this is the key feature and it doesn't call out things like hinted-handoff and anti-entropy services. These is an Enterprise only feature and should be highlighted - with pointers to the details.

Perhaps we should also point out here that InfluxDB Enterprise can run with a single meta node and data node -- which is non-clustered and not HA, but delivers on all of the other features/capabilities described.

##### Relevant URLs
https://docs.influxdata.com/enterprise_influxdb/v1.9/features/clustering-features/

Suggested additions:

Clustering
- How it works: https://docs.influxdata.com/enterprise_influxdb/v1.9/concepts/clustering/
- Supporting services which enable clustering: hinted-handoff and anti-entropy
- HH overview: https://docs.influxdata.com/enterprise_influxdb/v1.9/concepts/clustering/#hinted-handoff
- Anti-entropy: https://docs.influxdata.com/enterprise_influxdb/v1.9/administration/anti-entropy/ (activation of this service is optional and it exists to eliminate edge cases...but does require additional resources)
- The current section about shard movement and tools for Enterprise clusters should be tucked into this section - currently here: https://docs.influxdata.com/enterprise_influxdb/v1.9/features/clustering-features/#shard-movement
- Highlighting the differences in the various OSS features probably belong here as well: query routing, query management, continuous queries and subscriptions which all appear in their own sections already.

Security
- Expansion of core user permissions and roles.  Pointer to more details here: https://docs.influxdata.com/enterprise_influxdb/v1.9/features/users/
- Fine Grained Authorization.  Describe...and pointer to: https://docs.influxdata.com/enterprise_influxdb/v1.9/guides/fine-grained-authorization/
- Kapacitor Security integration (now with Kapacitor OSS)
- LDAP configuration option
- Password hashing alogrithim option

So the new outline for the page should look something like:

Clustering
Entitlements
Security
Backup and Restore
Migrating from OSS
Troubleshooting 
- via PProf and pointers to the platform monitoring docs

pierwill:
Closes #3080.

- [x] Rebased/mergeable


lwandzura (commented):
looks good

PR #3255, opened 2021-10-16
clarify run add-meta cmd on existing node

kelseiv:
Closes https://github.com/influxdata/EAR/issues/2694


danatinflux:
I don't think that's the right EAR? 2694 is for the token information for Chronograf.

PR #3257, opened 2021-10-17
Simplify 'send email' code examples

slai:
* instead of defining the HTTP POST payload as a string, define it as a record and use [`json.encode`](https://docs.influxdata.com/flux/v0.x/stdlib/json/encode/) to encode into bytes
* use [implicit string conversion](https://docs.influxdata.com/flux/v0.x/spec/lexical-elements/#string-literals) when interpolating values into the payload

- [x] Rebased/mergeable


slai:
@sanderson note removed. Thanks for the review!

PR #3324, opened 2021-10-28
Update terminology for blocked CIDR range

pierwill:
Merge after https://github.com/influxdata/kapacitor/pull/2649.

- [ ] Rebased/mergeable


pierwill:
If the changes suggested in https://github.com/influxdata/kapacitor/pull/2649 are ever made, we'll want to update the docs with changes like the ones here. But we can close this for now.

PR #3337, opened 2021-11-01
API Ref docs processing

jstirnaman:
API Reference Docs pipeline overhaul. Pulls `contracts/ref` files and applies custom content (tags, x-tagGroups, security, info, etc.) before writing results to cloud/ref.yml and v2.0/ref.yml.
- Renamed source files from swagger.yml to ref.yml to avoid confusion and in case we need to pull additional specs in the future.
- Uses OpenAPI CLI plugins with custom decorators.
- YAML and JS all the way down.


jstirnaman:
Still need to add documentation for the team.

sanderson (commented):
👌🏻 Looks really good. Thanks @jstirnaman

sanderson (approved):

PR #3342, opened 2021-11-03
fix: Authorization, not Authentication, header

jstirnaman:
(closes https://github.com/influxdata/DAR/issues/241)

### DIFF
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,0 +1,4 @@
+trailingComma: "es5"
+tabWidth: 2 
+semi: true 
+singleQuote: true
--- a/api-docs/cloud/ref.yml
+++ b/api-docs/cloud/ref.yml
@@ -5949,7 +5949,7 @@ components:
         within the organization.
 
 
-        Include your API token in an `Authentication: Token YOUR_API_TOKEN` HTTP
+        Include your API token in an `Authorization: Token YOUR_API_TOKEN` HTTP
         header with each request.
 
 
@@ -5957,7 +5957,7 @@ components:
 
 
         `curl https://us-east-1-1.aws.cloud2.influxdata.com/
-              --header "Authentication: Token YOUR_API_TOKEN"`
+              --header "Authorization: Token YOUR_API_TOKEN"`
 
         For more information and examples, see the following:
           - [Use tokens in API requests](https://docs.influxdata.com/influxdb/cloud/api-guide/api_intro/#authentication).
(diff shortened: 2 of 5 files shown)


kelseiv (approved):

kelseiv:
@jstirnaman to be sure, does it make sense to update any of the following occurrences?:

 ```
  QuerystringAuthentication:
      type: apiKey
      in: query
      name: u=&p=
      description: >
        ### Querystring authentication scheme

```

`BasicAuthentication:`
or
`2. Configure InfluxDB authentication:`

jstirnaman:
> @jstirnaman to be sure, does it make sense to update any of the following occurrences?:
> 
> ```
>  QuerystringAuthentication:
>      type: apiKey
>      in: query
>      name: u=&p=
>      description: >
>        ### Querystring authentication scheme
> ```
> 
> `BasicAuthentication:` or `2. Configure InfluxDB authentication:`

No, it's confusing, but "Authentication" is the "let me do this" process. "Authorization" is the HTTP header that tells the server what kind of creds you're sending. That's not be confused with, `/authorizations` in InfluxDB which manages tokens and their association to permissions.

kelseiv:
> No, it's confusing, but "Authentication" is the "let me do this" process. "Authorization" is the HTTP header that tells the server what kind of creds you're sending. That's not be confused with, `/authorizations` in InfluxDB which manages tokens and their association to permissions.

Thanks, @jstirnaman! OK, whatever we can say throughout to make this less confusing (and consistent) throughout the v2/Cloud docs the better...fwiw, here's where we landed with authentication/authorization in the enterprise docs cc @pierwill. I imagine we want to be consistent across those as well (unless any surrounding functionality has intentionally changed).

pierwill:
We should be careful with this. In English, I think we'd say that this process of passing a token constitutes *authentication*. But `Authorization: TOKEN` standard scheme, yes? @jstirnaman @sanderson. Even if the docs are *about authentication*, if the API expects `Authorization: TOKEN`, that's what the example should say.

sanderson:
`Authorization` is a [standard HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) used for authenticating with a service. It supports multiple schemes (Token, Bearer, Basic, etc.).

In regards to language, this is how I view the two terms:

- Authentication: the _process_ of authenticating with a service with defined credentials (username/password, token, etc.) to prove you are who you say you are.
- Authorization: a _set of permissions_ granted to the authenticated client to perform actions within the service.

PR #3363, opened 2021-11-05
Fix/api ref 2.1

jstirnaman:
Update OSS API Ref docs to v2.1.
Update examples to match guides.
Fix descriptions.


jstirnaman:
I've yet to apply some earlier suggestions from @kelseiv. This is mostly to get 2.1 out the door.

sanderson (approved):
👌🏻

PR #3386, opened 2021-11-10
Hotfix icons

kelseiv:
@sanderson looks like "data" works but not "disks" - did a global search/replace & updated contrib guidelines

I should say, for good measure, I replaced with "load-data" being that's what I saw you used in the monitor & alert PR, but verified that "data" works too


sanderson (approved):
Nice catch.

PR #3392, opened 2021-11-11
Update line-protocol.md

rickspencer3:

Adds information to the section about sending zipped protocol to explain the uncompressed limits and how the server reacts.

- [X ] Signed the [InfluxData CLA](https://www.influxdata.com/legal/cla/)
- [ X] Rebased/mergeable


sanderson (commented):
@rickspencer3 Does this limit apply to both OSS and Cloud, or is this unique to Cloud?

kelseiv:
thank you, @rickspencer3 @sanderson, shall we push this one out for Cloud?

....checking with [edge team]( https://influxdata.slack.com/archives/CPYJNHM4K/p1637696347302900) to verify whether implemented in OSS.

content/influxdb/cloud/write-data/developer-tools/line-protocol.md
```diff
@@ -120,6 +120,8 @@ influx write \
   --compression gzip
 ```
 
+There is a limit to the size of uncompressed data. As the data is uncompressed, if the uncompressed payload reaches that limit, the lines that have been uncompressed to that point will be written, but subsequent lines will not be. In such cases, the caller will receive a 413 from the server, and should retry with a smaller uncompressed payload. The current uncompressed limit is 250 MB, but this is subject to change.
```
kelseiv, line 123:
```suggestion
InfluxDB Cloud limits data in an uncompressed write request payload to 250 MB. After 250 MB is uncompressed and written, the remaining lines are not written, and the caller receives a 413 response. 
We recommend the caller retry the request with a smaller uncompressed payload.
```

sanderson: Once this suggestion is applied, I'm good to merge this.

sanderson (approved):

PR #3432, opened 2021-11-22
Update difference and increase functions

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/flux/v0.x/stdlib/universe/difference.md
+++ b/content/flux/v0.x/stdlib/universe/difference.md
@@ -24,9 +24,10 @@ _**Output data type:** Float_
 
 ```js
 difference(
-  nonNegative: false,
-  columns: ["_value"],
-  keepFirst: false
+    nonNegative: false,
+    columns: ["_value"],
+    keepFirst: false,
+    initialZero: false,
 )
 ```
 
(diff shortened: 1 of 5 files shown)


content/flux/v0.x/stdlib/universe/difference.md
```diff
@@ -24,17 +24,19 @@ _**Output data type:** Float_
...
+    initialZero: false,
 )
 ```
 
 ## Parameters
 
 ### nonNegative {data-type="bool"}
 Indicates if the difference is allowed to be negative.
-When set to `true`, if a value is less than the previous value, it is assumed the previous value should have been a zero.
+When set to `true`, if a value is less than the previous value, the function
```
kelseiv, line 38: _When set to `true`, if a value is less than the previous value, the function assumes the previous value should have been a zero._

to be sure, `nonNegatives = true` means diff is allowed to be negative? 

Curious why if the first value is 3 and second value is 2, why would the function assume 3 should be zero?

content/flux/v0.x/stdlib/universe/difference.md
```diff
@@ -24,17 +24,19 @@ _**Output data type:** Float_
 
 ```js
 difference(
-  nonNegative: false,
-  columns: ["_value"],
-  keepFirst: false
+    nonNegative: false,
+    columns: ["_value"],
+    keepFirst: false,
+    initialZero: false,
```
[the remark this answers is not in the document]
sanderson: Yeah, it's what the new `flux fmt` formatter does.

content/flux/v0.x/stdlib/universe/difference.md
sanderson: > to be sure, `nonNegatives = true` means diff is allowed to be negative?

No, when set to true, it returns a zero-sum change from the previous. This is a special application specifically designed so `increase()` correctly supports counter data.

PR #3471, opened 2021-12-08
Fix external plugin readme links

### LINKED ISSUE #3410: Broken link for Oracle plugin
When clicking on "View" for the Oracle plugin, the link is broken, you will get a 404 error as shown in the following screenshots.

image

image

I believe the correct location in GitHub should be the following:

https://github.com/bonitoo-io/telegraf-input-oracle/blob/main/README.md

However, this is not in the same repo where most other inputs plugin info is housed, which is normally here:

https://github.com/influxdata/telegraf/tree/release-1.20/plugins/inputs

##### Relevant URLs

https://docs.influxdata.com/telegraf/v1.20/plugins/#oracle

https://github.com/bonitoo-io/telegraf-input-oracle/blob/main/README.md

noramullen1:
Closes #3410 -- @lwandzura noticed an issue with one broken readme link for an external plugin, which turned out to apply to many other external plugins.

### DIFF
--- a/data/telegraf_plugins.yml
+++ b/data/telegraf_plugins.yml
@@ -42,7 +42,7 @@ input:
     description: |
       The Amazon CloudWatch Alarms input plugin pulls alarm statistics from Amazon CloudWatch.
     introduced: 1.16.0
-    link: https://github.com/vipinvkmenon/awsalarms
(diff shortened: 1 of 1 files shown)


sanderson (approved):
LGTM

Issue #3201, opened 2021-10-04
SHOW MEASUREMENTS ON *.*


lesam:
https://github.com/influxdata/influxdb/pull/22388
https://github.com/influxdata/influxdb/pull/22396
https://github.com/influxdata/plutonium/pull/3632

_Describe the new feature here._

See issue comments:

https://github.com/influxdata/influxdb/issues/3318#issuecomment-719841659 
https://github.com/influxdata/influxdb/issues/3318#issuecomment-721117864
https://github.com/influxdata/influxdb/issues/3318#issuecomment-912809271

This will be in Enterprise 1.10, OSS 2.1, and a future release of Cloud (date undecided).

davidby-influx:
This needs to be **_removed_** from the 1.8 Docs: https://docs.influxdata.com/influxdb/v1.8/query_language/explore-schema/#show-measurements

It is not in 1.8, and will not be in 1.8 or 1.9

@pierwill

PR #3483, opened 2021-12-13
Remove wildcards from 1.8 InfluxQL schema doc

pierwill:
See https://github.com/influxdata/docs-v2/issues/3201#issuecomment-992927665.

### DIFF
--- a/content/influxdb/v1.8/query_language/explore-schema.md
+++ b/content/influxdb/v1.8/query_language/explore-schema.md
@@ -438,11 +438,6 @@ See the Data Exploration page for documentation on the
 [`OFFSET` clause](/influxdb/v1.8/query_language/explore-data/#the-offset-clause),
 and on [Regular expressions in queries](/influxdb/v1.8/query_language/explore-data/#regular-expressions).
 
-The following queries with wildcards characters (`*`) are also allowed:
-
-- `SHOW MEASUREMENTS ON *.*` to show all databases and all RPs
-- `SHOW MEASUREMENTS ON <db>.*` to show a specific database and all its RPs
-
 ### Examples
 
 #### Run a `SHOW MEASUREMENTS` query with the `ON` clause


content/influxdb/v1.8/query_language/explore-schema.md
```diff
@@ -438,10 +438,7 @@ See the Data Exploration page for documentation on the
 [`OFFSET` clause](/influxdb/v1.8/query_language/explore-data/#the-offset-clause),
 and on [Regular expressions in queries](/influxdb/v1.8/query_language/explore-data/#regular-expressions).
 
-The following queries with wildcards characters (`*`) are also allowed:
-
-- `SHOW MEASUREMENTS ON *.*` to show all databases and all RPs
-- `SHOW MEASUREMENTS ON <db>.*` to show a specific database and all its RPs
+Run `SHOW MEASUREMENTS ON <db>.*` to show a specific database and all its RPs.
```
davidby-influx, line 441: I think this line has to go as well.  @lesam ?

davidby-influx (changes requested):
I believe all the wildcards have to be removed.  @lesam ?  Please confirm.

sanderson (commented):
@davidby-influx @lesam Is this feature supported in Enterprise 1.9? What about in the current 2.x and Cloud InfluxQL implementations?

davidby-influx:
@sanderson - I don't think it is in 1.9 Enterprise. It is slated for 1.10 Enterprise.

No idea about cloud. 

This is in 2.1.0 OSS

lesam:
@sanderson not yet supported in cloud, the issue is here: https://github.com/influxdata/idpe/issues/11806

content/influxdb/v1.8/query_language/explore-schema.md
lesam: Yes. Previous behaviour, preserved in all versions:

```
show measurements # only valid from CLI after use database, same as:
show measurements on <database> # shows all measurements from all retention policies, not broken out by policy
```

Additional behaviour in OSS 2.1.0 and Enterprise 1.10:
```
show measurements on *.* # all databases and all retention policies, broken out by retention policy
show measurements on <db>.* # specific database and all RPs, broken out by retention policy (different than non-wildcard which is not broken by retention policy, for backward-compatibility)
show measurements on <db>.<rp> # specific database and specific RP
```

Error everywhere:
```
show measurements .*
show measurements on .*
```

sanderson:
@lwandzura Although we're removing this wildcard support from the 1.8 InfluxQL docs, we need to make sure it is added to the InfluxQL docs in OSS 2.x and Cloud.

sanderson (approved):

PR #3506, opened 2021-12-17
Update go.md

samhld:
Closes #

- [ ] Rebased/mergeable


sanderson (commented):
@samhld When I check the suggested link change, I get a 404.

samhld:
Woops should have only been the `go get` statement.  The correct URL for that includes the `/v2` according to: https://github.com/influxdata/influxdb-client-go

I also tried both and the `/v2` version is the one that works.

sanderson (approved):
Thanks @samhld!

PR #3510, opened 2021-12-20
Update 2.x linux installation instructions

### LINKED ISSUE #3509: 2.x linux installation instructions might result in duplicated influxd processes
The 2.x linux installation instructions at https://docs.influxdata.com/influxdb/v2.1/install/?t=Linux do not explicitly state to do one of the following between download the binary and running that directly vs. running a systemd service. If somebody follows the instructions from beginning to end, they will start the binary, and then run into problems when they start the systemd service. I've seen this happen from community reports.

Also the first section refers to `Download the InfluxDB package.` - these section is not downloading a package, but a binary. So I think this might contribute to some confusion

wbaker85:
Closes #3509

Clarifies that only one of the two installation steps is necessary (install systemd package OR download binary), also clarifies wording related to the downloaded binary.

I moved the systemd option to come before manually downloading the binary to be symmetrical with the mac instructions.


content/influxdb/v2.1/install.md
```diff
@@ -205,9 +210,71 @@ For information about installing the `influx` CLI, see
...
+    ● influxdb.service - InfluxDB is an open-source, distributed, time series database
+      Loaded: loaded (/lib/systemd/system/influxdb.service; enabled; vendor preset: enable>
+      Active: active (running)
+    ```
+
+When installed as a service, InfluxDB stores data in the following locations:
+
+- **Time series data:** `/var/lib/influxdb/engine/`
+- **Key-value data:** `/var/lib/influxdb/influxd.bolt`.
+- **influx CLI configurations:** `~/.influxdbv2/configs` _(see [`influx config`](/influxdb/v2.1/reference/cli/influx/config/) for more information)_ .
```
sanderson, line 252:
```suggestion
For information about where InfluxDB stores data on disk when running as a service,
see [File system layout](/influxdb/v2.1/reference/internals/file-system-layout/?t=Linux#installed-as-a-package).
```

content/influxdb/v2.1/install.md
```diff
@@ -205,9 +210,71 @@ For information about installing the `influx` CLI, see
...
+   ```
+
+2. Edit the `/lib/systemd/system/influxdb.service` file as follows:
+
+   ```sh
+   ExecStart=/usr/bin/influxd $ARG1 $ARG2
+   ```
 
-    Download the InfluxDB package [from your browser](#download-from-your-browser)
+### Manually Download and Install Linux Binary
```
sanderson, line 273:
```suggestion
### Manually download and install the influxd binary
```

content/influxdb/v2.1/install.md
```diff
@@ -196,6 +196,11 @@ influxd --reporting-disabled
 {{% tab-content %}}
 ## Download and install InfluxDB v{{< current-version >}}
 
+Do one of the following:
+
+- [Install InfluxDB as a service with systemd](#install-influxdb-as-a-service-with-systemd)
+- [Manually download and install linux binary](#manually-download-and-install-linux-binary)
```
sanderson, line 202:
```suggestion
- [Manually download and install the influxd binary](#manually-download-and-install-the-influxd-binary)
```

content/influxdb/v2.1/install.md
```diff
@@ -286,7 +353,7 @@ If `gpg` is not available, see the [GnuPG homepage](https://gnupg.org/download/)
 
 ## Start InfluxDB
 
-Start InfluxDB by running the `influxd` daemon:
+If InfluxDB was installed as a systemd service, systemd will manage starting and stopping the `influxd` daemon and no further action is required. If the binary was manually downloaded and added to the system `$PATH`, the `influxd` daemon can be started using the following command:
```
sanderson, line 356:
```suggestion
If InfluxDB was installed as a systemd service, systemd manages the `influxd` daemon and no further action is required.
If the binary was manually downloaded and added to the system `$PATH`, start the `influxd` daemon with the following command:
```

sanderson (changes requested):
Added some suggestions.

wbaker85:
@sanderson thanks for the suggestions! I incorporated them all. Let me know if there's anything else.

sanderson (approved):

Issue #3426, opened 2021-11-19
Update "Set up FGA" instructions for Influx Enterprise v.1.9


lwandzura:
Remove "Optional" from Step 4 in "Use fine-grained authorization in InfluxDB Enterprise" doc.  

![image](https://user-images.githubusercontent.com/51929958/142691075-0d6dcbe0-0da2-438f-8554-6bab1230d315.png)

This is incorrect.  The way to run FGA should be:

* Enable the user with roles or InfluxQL GRANT to have coarse read (and if any writes are expected, write) on a database
* Restrict users on the database with FGA (acl/restrictions) endpoint)
* Use grants to allow some users to see some parts of the database (acl/grants)

You need to use the coarse permissions FIRST, before restricting to a fine grain.

##### Relevant URLs

URL for Using FGA for version 1.9:
https://docs.influxdata.com/enterprise_influxdb/v1.9/guides/fine-grained-authorization/#specify-role-permissions

URL for v1.6 explains it better:
https://docs.influxdata.com/enterprise_influxdb/v1.6/guides/fine-grained-authorization/

lwandzura:
Make sure to add Sam Arnold as reviewer

PR #3513, opened 2021-12-21
FGA update

lwandzura:
Closes #https://github.com/influxdata/docs-v2/issues/3426

Removed Optional from Step 3, added language to clarify starting with coarse read/write permissions first, then moving to fine grained permissions via roles or grants.


content/enterprise_influxdb/v1.9/guides/fine-grained-authorization.md
```diff
@@ -48,8 +48,18 @@ In a typical cluster configuration, the HTTP ports for data nodes
 You may need to work with your network administrator to gain access to the meta node HTTP ports.
     {{% /note %}}
 
-4. _(Optional)_ [Create roles](#manage-roles).
-   Roles let you grant permissions to groups of users assigned to each role.
+4. [Create roles](#manage-roles). Roles let you grant permissions to groups of users assigned to each role.
+   As administrator, create a user with a standard grant, as shown in the example below, and then restrict by role or via grants. 
+   You **must** set up coarse read/write permissions **FIRST**, then restrict to a fine grain.
```
kelseiv, line 53: - Let's add steps to indicate required order:

4.
   ..a.   
   ..b.

- Let's use replace "set up coarse" to "grant all" (more explicit)

kelseiv: Re:
create a user with a standard grant, as shown in the example below, and then restrict by role or via grants.

The example below grants users `east` and `west` all permissions (insert reference link to all permissions) on the `datacenters` database. 

And then, let's add a fine grain permission to one of the users for clarity (in example and text).

kelseiv, line 53:
```suggestion
4. [Create roles](#manage-roles) to grant permissions to users assigned to a role. Do the following:
   a.  As administrator, grant users all permissions. The example below grants users `east` and `west` all permissions on the `datacenters` database.  
   b. Add a fine grain permission to users as needed.
```

kelseiv: - Does this example create a role...or show how to create a user and grant all permissions (before the role is created)? Let's clarify accordingly.
- Let's add an example of how to specify a fine grain permission (to either user) here, clarifying in text and example.

pierwill (approved):
I agree with @kelseiv that step 4 could use some clarification.

Perhaps ` You **must** set up coarse read/write permissions **FIRST**, then restrict to a fine grain.` should be in a separate note?

kelseiv:
> I agree with @kelseiv that step 4 could use some clarification. Perhaps You **must** set up coarse read/write permissions **FIRST**, then restrict to a fine grain. should be in a separate note?

@pierwill @lwandzura, I did intentionally remove this statement/note above in my edit. My intention was to  1.) use numbers to convey the order things need to happen and 2.) replace the term "coarse read/write permissions" with "grant all permissions" (given that's what the example shows and given the term "coarse" isn't super explicit).

I see @lwandzura moved the user/permissions to Step 4 and made Step 5 creating roles--looks good!

Let's also move lines 55-63 up after line 52. As it's written now could be read as the example shows how to add a fine grain permission.....ideally, we'd also add an example of "how to add a fine grain permission" (for either the `east` or `west` user) after line 53 to clarify in context...but as discussed, some of this content maybe further down in the doc, and given we don't want to duplicate, we'll need to read through the whole doc to determine what are the necessary steps for setting up FGA (and only the necessary steps), and then removing additional conceptual info not related to the task at hand into another doc

PR #3519, opened 2021-12-22
Add more examples to usage.limits

### LINKED ISSUE #3502: Expand usage limits examples to make the output more usable for dashboards and tasks
Currently within InfluxDB Cloud, users can retrieve their usage limits using the [usage.limits() function.](https://docs.influxdata.com/flux/v0.x/stdlib/experimental/usage/limits/)

However, the output of that function is a JSON object and something that can directly be leveraged with other functions.

To successfully leverage the output, we require that JSON object to be parsed and mapped into something that could be used in dashboards cells or tasks using the [json.parse()](https://docs.influxdata.com/flux/v0.x/stdlib/experimental/json/parse/) function.

This proposal is to expand the example on the usage limits page to show how to convert the output into something which can be used (potentially via findRecord) within Dashboard cells (to display single stats for the individual rate limits) or tasks (in order to allow for an early warning check/alert).

sanderson:
Closes #3502

- [x] Rebased/mergeable


gavincabbage (approved):
Awesome!

PR #3537, opened 2022-01-04
Use influxdb_v2 plugin in telegraf getting-started.md

sspaink:
Updated the Telegraf getting started documentation to recommend using the newer `influxdb_v2` plugin when generating the telegraf configuration. I also felt this part could use more explanation, so I added a more simple and complicated version of the command and a link to the other possible flags. Also it seems `--sample-config` and `-sample-config` are both valid but the Telegraf docs only mention `--sample-config`, so updated the examples to have the extra dash.

This came from a discussion with @appletreeisyellow, what do you think of these changes?

- [x] Rebased/mergeable


sspaink:
!signed-cla

noramullen1 (approved):

PR #3672, opened 2022-01-12
Remove verbose output for Enterprise API curl examples

pierwill:
It seems to me that the only useful information in these is the returned status codes. The TLS handshake information is outside the scope of this doc, I'd say, which is focused on managing authorizations.


pierwill:
@sanderson approved in another channel.

PR #3678, opened 2022-01-13
fix: link to reference for field data type. (#3056)

### LINKED ISSUE #3056: Add the valid data types to the measurement schema page
https://docs.influxdata.com/influxdb/cloud/organizations/buckets/bucket-schema/

I was wondering what the various types were for the dataTypes. according to the API spec, they are:

"integer" "float" "boolean" "string" "unsigned"

We should add them somewhere in the docs. 

https://docs.influxdata.com/influxdb/cloud/api/#operation/createMeasurementSchema

jstirnaman:
Closes #3056 

Provide the reference for field data type values.

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud/organizations/buckets/bucket-schema/_index.md
+++ b/content/influxdb/cloud/organizations/buckets/bucket-schema/_index.md
@@ -85,7 +85,7 @@ Use the `influx` CLI to set the schema-type and measurement schemas for your buc
       ```json
       [
        {"name":"time","type":"timestamp"},
-       {"name":"fsWrite","type":"field","dataType":"float"}
+       {"name":"fsWrite","type":"field"}
       ]
       ```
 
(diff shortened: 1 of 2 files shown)


content/influxdb/cloud/organizations/buckets/bucket-schema/_index.md
```diff
@@ -97,6 +97,18 @@ Use the `influx` CLI to set the schema-type and measurement schemas for your buc
       ]
       ```
 
+   The default field data type is `string`.
+   To set the data type of a field column, provide the `dataType` property and a valid
+   [field data type](/influxdb/cloud/reference/key-concepts/data-elements/#field-value),
```
kelseiv, line 102: thoughts on:
- explicitly stating here "strings, floats, integers, or booleans" so folks don't have to click
- and/or rephrasing valid data type for the [field value]

kelseiv (approved):

content/influxdb/cloud/organizations/buckets/bucket-schema/_index.md
kelseiv: also looks like we may need to add "timestamp" to the list in the "Field value" section?
per https://docs.influxdata.com/influxdb/cloud/reference/glossary/#data-type
also, would it make sense to clarify the distinction between integer and unsigned integer?

PR #3691, opened 2022-01-18
Add `-start` and `-end` flags

gwossum:
Add `-start` and `-end` flags to `influxd-ctl backup` documentation. These flags are implemented by https://github.com/influxdata/plutonium/pull/3764 and will be available in the enterprise 1.10 release.


timhallinflux:
We should add an example which shows the use of the -start and -end flags.  Docs team member to add this as part of validating with 1.10 Enterprise release.

samhld:
@lwandzura yes it is.

sanderson (changes requested):
@gwossum These changes should be made on the `enterprise-1.10` branch and to the `content/enterprise_influxdb/v1.9/administration/backup-and-restore.md` file. We should also include a time bound backup example in the collection of examples.

samhld (approved):

kelseiv:
need to update the following doc as well:

https://docs.influxdata.com/enterprise_influxdb/v1.10/tools/influxd-ctl/#backup

PR #3729, opened 2022-01-28
Update `-shard` option for PR 22998 for InfluxDB OSS 1.10

gwossum:
If InfluxDB OSS 1.10 is released, then then `-shard` backup documentation needs to be updated to match the improvements in https://github.com/influxdata/influxdb/pull/22998/

Closes #

- [ ] Rebased/mergeable


kelseiv:
@gwossum --want to loop back around on this one --give we never released OSS 1.10, and this option was added to the Enterprise docs--- to be sure, this option was not back ported to OSS 1.8? thanks!

PR #3743, opened 2022-02-02
Use bytes in certain Enterprise config examples

### LINKED ISSUE #3649: max-index-log-file-size parsed differently between config and environment
See https://github.com/influxdata/influxdb/issues/22974.

pierwill:
Closes #3649.

### DIFF
--- a/content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
+++ b/content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
@@ -262,13 +262,13 @@ Environment variable: `INFLUXDB_DATA_INGRESS_METRIC_BY_LOGIN_ENABLED`
 
 #### `cache-max-memory-size = "1000000000"  # 1g`
 
-The maximum size a shard cache can reach before it starts rejecting writes.
+The maximum size in bytes that a shard cache can reach before it starts rejecting writes.
 
 Environment variable: `INFLUXDB_DATA_CACHE_MAX_MEMORY_SIZE`
 
 #### `cache-snapshot-memory-size = "26214400"  # 25m`
 
-The size at which the TSM engine will snapshot the cache and write it to a TSM file, freeing up memory.
+The size in bytes at which the TSM engine will snapshot the cache and write it to a TSM file, freeing up memory.
 
 Environment variable: `INFLUXDB_DATA_CACHE_SNAPSHOT_MEMORY_SIZE`
 
--- a/content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
+++ b/content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
@@ -260,13 +260,13 @@ Environment variable: `INFLUXDB_DATA_INGRESS_METRIC_BY_LOGIN_ENABLED`
 
 ### Data settings for the TSM engine
 
-#### `cache-max-memory-size = "1000000000"`
+#### `cache-max-memory-size = 1000000000`
 
 The maximum size in bytes that a shard cache can reach before it starts rejecting writes.
 
 Environment variable: `INFLUXDB_DATA_CACHE_MAX_MEMORY_SIZE`
 
-#### `cache-snapshot-memory-size = "26214400"`
+#### `cache-snapshot-memory-size = 26214400`
 
 The size in bytes at which the TSM engine will snapshot the cache and write it to a TSM file, freeing up memory.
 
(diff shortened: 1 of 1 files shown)


content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
```diff
@@ -356,7 +356,7 @@ Environment variable: `INFLUXDB_DATA_MAX_VALUES_PER_TAG`
 
 ### TSI (`tsi1`) index settings
 
-#### `max-index-log-file-size = "1m"`
+#### `max-index-log-file-size = "131072"  # 1m`
```
lesam, line 359: The code that parses this is here:

```
	// Parse unit of measure
	suffix := text[len(sizeText)-1]
	if !unicode.IsDigit(rune(suffix)) {
		switch suffix {
		case 'k', 'K':
			mult = 1 << 10 // KiB
		case 'm', 'M':
			mult = 1 << 20 // MiB
		case 'g', 'G':
			mult = 1 << 30 // GiB
		default:
			return fmt.Errorf("unknown size suffix: %c (expected k, m, or g)", suffix)
		}
		sizeText = sizeText[:len(sizeText)-1]
	}
```

These are standard SI prefixes, so this should be 

```
#### `max-index-log-file-size = "1000000" # 1m`
```

pierwill, line 359:
```suggestion
#### `max-index-log-file-size = "1048576"  # 1m`
```

sanderson (commented):
- Should these values (without units) be strings or integers?
- We need to specifty the unit of the values we're recommending in the option description. For example: "The size at which the TSM engine will ... in bytes.
- I REALLY want to get away from including default values in configuration headings. I know this is just following the established pattern, but the established pattern isn't good. Any time we have to update the default value, we have to find and replace links to that anchor.
The default value should be part of the description. I don't know that this needs to be done as part of this PR, but we should plan on this in the future.

pierwill:
@lesam Are these values okay as strings, or should the be integers?

content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
```diff
@@ -260,15 +260,15 @@ Environment variable: `INFLUXDB_DATA_INGRESS_METRIC_BY_LOGIN_ENABLED`
 
 ### Data settings for the TSM engine
 
-#### `cache-max-memory-size = "1g"`
+#### `cache-max-memory-size = "1000000000"  # 1g`
```
sanderson, line 263:
```suggestion
#### `cache-max-memory-size = "1000000000"`
```

content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
```diff
@@ -260,15 +260,15 @@ Environment variable: `INFLUXDB_DATA_INGRESS_METRIC_BY_LOGIN_ENABLED`
...
 
-The maximum size a shard cache can reach before it starts rejecting writes.
+The maximum size in bytes that a shard cache can reach before it starts rejecting writes.
 
 Environment variable: `INFLUXDB_DATA_CACHE_MAX_MEMORY_SIZE`
 
-#### `cache-snapshot-memory-size = "25m"`
+#### `cache-snapshot-memory-size = "26214400"  # 25m`
```
sanderson, line 269:
```suggestion
#### `cache-snapshot-memory-size = "26214400"`
```

content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
```diff
@@ -288,14 +288,14 @@ This setting does not apply to cache snapshotting.
 
 Environmental variable: `INFLUXDB_DATA_CACHE_MAX_CONCURRENT_COMPACTIONS`
 
-#### `compact-throughput = "48m"`
+#### `compact-throughput = "50331648"  # 48m`
```
sanderson, line 291:
```suggestion
#### `compact-throughput = "50331648"`
```

content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
```diff
@@ -288,14 +288,14 @@ This setting does not apply to cache snapshotting.
...
 
 The maximum number of bytes per seconds TSM compactions write to disk. Default is `"48m"` (48 million).
 Note that short bursts are allowed to happen at a possibly larger value, set by `compact-throughput-burst`.
 
 Environment variable: `INFLUXDB_DATA_COMPACT_THROUGHPUT`  
 
-#### `compact-throughput-burst = "48m"`
+#### `compact-throughput-burst = "50331648"  # 48m`
```
sanderson, line 298:
```suggestion
#### `compact-throughput-burst = "50331648"`
```

content/enterprise_influxdb/v1.9/administration/configure/config-data-nodes.md
```diff
@@ -356,7 +356,7 @@ Environment variable: `INFLUXDB_DATA_MAX_VALUES_PER_TAG`
 
 ### TSI (`tsi1`) index settings
 
-#### `max-index-log-file-size = "1m"`
+#### `max-index-log-file-size = "1048576"  # 1m`
```
sanderson, line 359:
```suggestion
#### `max-index-log-file-size = "1048576"`
```

sanderson (approved):
I added suggestions to remove the "comments" from each heading.

lesam:
> @lesam Are these values okay as strings, or should they be integers?

In the toml, they can be strings, strings with SI suffixes, or integers - all are accepted (though I'd recommend integers).

When passed as environment variables they have to be strings (because env variables can't be ints) and we don't support the suffixes.

PR #3757, opened 2022-02-07
Update regular-expressions.md

sjwang90:
i found the double negative here confusing. i think because of the `=~` my brain shifted to what should be _included_ not excluded.

feel free to reject if prefer it how it is


sanderson (approved):
Much better. Thanks @sjwang90.

PR #3809, opened 2022-02-24
Fix precision definition in Telegraf configuration

sjwang90:
https://community.influxdata.com/t/telegraf-precision-parameter-parsing-in-agent-configuration/24002/7
https://github.com/influxdata/telegraf/issues/10734


Franky1:
Description of `precision` is much better now! I am fine with the new description.

powersj (approved):
Agreed that the changes make this much more clear to the user and as written are not valid configs.

noramullen1 (approved):

PR #3873, opened 2022-03-18
Add hardening-enabled configuration option

### LINKED ISSUE #3866: Document new InfluxDB OSS --hardening-enabled feature
**PR:** https://github.com/influxdata/influxdb/pull/23207

A new `--hardening-enabled` option is added to influxd turn on IP address verification and limit both flux and template fetching HTTP requests to non-private addresses. In the future, we plan to enable new security features that aren't suitable for the default install with this option. Put another way, this new option is intended to be used to make it easy to turn on all security options when running in production environments.

It's enabled by starting InfluxDB with the `--hardening-enabled` command line flag.

It's useful because certain flux functions (eg, `to()`, `from()`, etc) and template fetching cause InfluxDB to make HTTP requests over the network. When private IP validation is enabled, InfluxDB will first verify that the IP address of the URL is not a private IP address.

Milestoning to InfluxDB OSS 2.2, but we plan to backport this to 2.1 as well.

sanderson:
Closes #3867
Closes #3866 

Ports the changes made by @jdstrand in #3867 to the 2.2 release branch.

### DIFF
--- a/content/influxdb/v2.2/reference/config-options.md
+++ b/content/influxdb/v2.2/reference/config-options.md
@@ -125,6 +125,7 @@ To configure InfluxDB, use the following configuration options when starting the
 - [engine-path](#engine-path)
 - [feature-flags](#feature-flags)
 - [flux-log-enabled](#flux-log-enabled)
+- [hardening-enabled](#hardening-enabled)
 - [http-bind-address](#http-bind-address)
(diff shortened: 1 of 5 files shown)


jdstrand (approved):
I didn't run this under hugo but it reads great and includes the changes from the original PR. Thanks!

timhallinflux (approved):

Issue #3871, opened 2022-03-18
Add example for templating email addresses in email alert.


timhallinflux:
**PR:** 
[Templating for address in email alert](https://github.com/influxdata/kapacitor/pull/2650)

_Describe the new feature here._
This adds templating email addresses for email alerts.

timhallinflux:
@docmerlin @samhld -- might need some additional inputs for docs.

danatinflux:
This is for the `ToTemplates()` method, AFAIK.

docmerlin:
The email handler has a method .toTemplates(template string)

Example:
```flux
stream
	|from()
		.measurement('cpu')
		.where(lambda: "host" == 'serverA')
		.groupBy('host')
	|window()
		.period(10s)
		.every(10s)
	|count('value')
	|default()
		.field('extraemail','bob@example.com')
		.tag('tagemail','bob2@example.com')
	|alert()
		.id('kapacitor.{{ .Name }}.{{ index .Tags "host" }}')
		.details('''
<b>{{.Message}}</b>
Value: {{ index .Fields "count" }}
<a href="http://graphs.example.com/host/{{index .Tags "host"}}">Details</a>
''')
		.info(lambda: "count" > 6.0)
		.warn(lambda: "count" > 7.0)
		.crit(lambda: "count" > 8.0)
		.email()
			.to('user1@example.com', 'user2@example.com')
			.toTemplates('{{ index .Fields "extraemail" }}')
```

Like the .to method, .toTemplates can be used more than once.

danatinflux:
@docmerlin The original case from Infor wanted to create e-mail addresses on the fly by leveraging tag/field values. I don't believe this addresses this use case?

Here's what they posted:

"We have a team that is attempting to build an email address out of a tag.  This doesn’t seem to be a thing, though.  Whenever the email address is being built – it doesn’t replace the token with the actual value.  It’s been tried in several different ways.

An example below.

The ‘var farm’ – gets set properly and is fine.  Except if it is used to build an email address.  It reverts back to making {{ index.Tags “farm” }} as a literal.

Trying to build the email with the token directly – ends up with no replacement being done, and the token is taken as a literal.  Doesn’t matter where it is done (at top, or in the data section or in the alert section).

Is there something special that is done when dealing with strings that look like email addresses, and this isn’t possible?"


```

// *********************************************
// Swap Alerts
// *********************************************
// Alert if average threshold exeeded, or std deviation exceeded (2/2.5/3)
[...]

Issue #3872, opened 2022-03-18
[Retroactive docs change] 1.6.3 added toTemplates method for email handler


samhld:
@docmerlin Can you add here what needs changing? I assume this needs to change at least:
  - 1.6.3 / 1.6.4 (retro) release notes
  - [email (smtp) handler page](https://docs.influxdata.com/kapacitor/v1.6/event_handlers/email/)
  - smtp portions of the config docs?

docmerlin:
Yes those will need changes.
The email handler has a method `.toTemplates(template string)` 

Example:

```tick
stream
					|from()
						.measurement('cpu')
						.where(lambda: "host" == 'serverA')
						.groupBy('host')
					|window()
						.period(10s)
						.every(10s)
					|count('value')
					|default()
						.field('extraemail','bob@example.com')
						.tag('tagemail','bob2@example.com')
					|alert()
						.id('kapacitor.{{ .Name }}.{{ index .Tags "host" }}')
						.details('''
<b>{{.Message}}</b>
Value: {{ index .Fields "count" }}
<a href="http://graphs.example.com/host/{{index .Tags "host"}}">Details</a>
''')
						.info(lambda: "count" > 6.0)
						.warn(lambda: "count" > 7.0)
						.crit(lambda: "count" > 8.0)
						.email()
							.to('user1@example.com', 'user2@example.com')
							.toTemplates('{{ index .Fields "extraemail" }}')
```

Like the `.to` method, `.toTemplates` can be used more than once.

PR #3886, opened 2022-03-24
Add to template

lwandzura:
Closes https://github.com/influxdata/docs-v2/issues/3871 and https://github.com/influxdata/docs-v2/issues/3872

Add toTemplate option to Email event handler doc with description and example


samhld (approved):
LGTM!

PR #3901, opened 2022-03-28
fix: update RPM/DEB repo URLS

powersj:
We now have stable, codename agnostic repo URLs for installing via a DEB or RPM. These URLs prevent the need to update constantly when new versions and/or releases occur. It also ensures that as new distros are released we do not need to add them.

- [x] Rebased/mergeable


sanderson:
Thanks @powersj! Are the old repo URLs still valid or should we backport these changes to the previous Telegraf versions docs?

powersj:
> Thanks @powersj! Are the old repo URLs still valid or should we backport these changes to the previous Telegraf versions docs?

Great question, old URLs are still valid and will remain backward compatible, but will not be getting any new releases and/or codenames added to the

PR #3938, opened 2022-04-07
fix: telegraf release notes version

sjwang90:
Should be 1.22.1 not 1.22.0


kelseiv:
looks like this has been resolved-thank you @sjwang90

PR #3939, opened 2022-04-07
Document `--write-variables` and `--write-annotations` in `influx auth create`

chrblabla:
Seems like this has not found it's way into the documentation, yet. Here, I added the info to use `--write-variables` which can be granted to users via the `influx auth create` command.


chrblabla:
!signed-cla

sanderson:
@chrblabla Thanks for this PR. Looks like there are some upstream conflicts with some changes that were recently merged into the docs. Could you resolve the conflicts and then we'll go ahead and merge your changes.

kelseiv (approved):

PR #3944, opened 2022-04-08
chore: durability updates for TSM snapshots

Joe-Blount:
Closes # https://github.com/influxdata/idpe/issues/13699

We're adding an occasional backup of TSM files.  The purpose is to allow us to satisfy our backup requirements without needing to keep the OOB backups forever (as we do today).  This will allow us to make commitments on how quickly deleted data is purged from our systems - today we cannot make any such commitment.

This PR adds some minor description of the new TSM backup.  More or less detail could be provided - I tried to mimic the level of detail in the other sections.

- [ ] Rebased/mergeable


content/influxdb/cloud/reference/internals/durability.md
```diff
@@ -65,17 +66,23 @@ When each compaction cycle completes, InfluxDB Cloud stores compressed
 [TSM](/influxdb/cloud/reference/glossary/#tsm-time-structured-merge-tree) files
 in object storage.
 
+### Periodic TSM snapshots
+Periodically the TSM files uploaded to object storage are snapshoted.
+Each TSM snapshot is a complete copy of all non-deleted data, as of the time the TSM snapshot is created.
+These snapshots are created weekly and preserved for 100 days, providing multiple recovery points.
```
kelseiv, line 72:
```suggestion
To provide multiple data recovery points, InfluxDB Cloud takes weekly snapshots of TSM files uploaded to object storage. The TSM snapshot includes a copy of all (non-deleted) data when the snapshot is created.
These snapshots are preserved for 100 days.
```

content/influxdb/cloud/reference/internals/durability.md
```diff
@@ -65,17 +66,23 @@ When each compaction cycle completes, InfluxDB Cloud stores compressed
...
-- **Historic backup:** compressed TSM files
+- **Compaction backup:** TSM files
+- **TSM Snapshots:** Weekly snapshots of TSM files in objectstore
 
 The Recovery Point Objective (RPO) is any accepted write.
 The Recovery Time Objective (RTO) is harder to definitively predict as potential failure modes can vary.
 While most common failure modes can be resolved within minutes or hours,
 critical failure modes may take longer.
-For example, if we need to rebuild all data from the message queue backup,
+For example, if we need to rebuild all data from the TSM Snapshots and message queue backup,
```
kelseiv, line 85:
```suggestion
For example, if we need to rebuild all data from the TSM snapshots and message queue backup,
```

kelseiv (approved):

Joe-Blount:
LGTM.

I don't think this is worth elaborating on in the documentation, but I'll provide some more detail here and you can include it if you think its appropriate.  Regarding the `100 days`:  We are required to have 90 days of backups, which I'm interpreting to mean we must always have a backup available to protect data 90d old.

So with weekly backups kept for 100d, the oldest backup at any time will be 93-100d old.  That's because when a backup reaches 100d old and we delete it, the next oldest backup is 93d old (giving us 3d of buffer before violating our commitments).
All we want to commit to is 90d protection, as stated in our TOS.  That is accomplished by keeping weekly backups for 100d.

I think this documentation is fine as it is, because this doc is explaining what we do, not necessarily what we commit to.

kelseiv:
Thank you, @Joe-Blount, I agree this is good for our docs site context.

This add'l context like this maybe appropriate for our Sales Enablement on Disaster Recovery brochure @lwandzura, @bpalani?


cc @noramullen1 @garylfowler @gunnaraasen - re Monthly Cloud updates

content/influxdb/cloud/reference/internals/durability.md
```diff
@@ -65,17 +66,22 @@ When each compaction cycle completes, InfluxDB Cloud stores compressed
...
+To provide multiple data recovery points, InfluxDB Cloud takes weekly snapshots of TSM files uploaded to object storage. The TSM snapshot includes a copy of all (non-deleted) data when the snapshot is created.
+These snapshots are preserved for 100 days.
+
 ## Recovery
 InfluxDB Cloud uses the following out-of-band backups stored in object storage to recover data:
 
 - **Message queue backup:** line protocol from inbound write requests within the last 96 hours
-- **Historic backup:** compressed TSM files
+- **Compaction backup:** TSM files
+- **TSM Snapshots:** Weekly snapshots of TSM files in objectstore
```
kelseiv, line 78:
```suggestion
- **TSM snapshots:** Weekly snapshots of TSM files in objectstore
```

PR #3966, opened 2022-04-20
fix: add missing `remotes`/`replications` `auth create` flags

DStrand1:
Adds missing CLI flags for `influx auth create` for InfluxDB 2.2 (`--read-remotes`, `--write-remotes`, `--read-replications`, `--write-replications`)

These flags were (and currently still are) missing from the help text in `influx auth create -h` in the latest InfluxCLI release. However, they currently will work if used, but are just not displayed anywhere. The CLI PR adding them to the `-h` text: https://github.com/influxdata/influx-cli/pull/374


sanderson (approved):
Thanks @DStrand1!

Issue #4034, opened 2022-05-20
Manipulate timestamps page empty


lwandzura:
The Manipulate timestamps page in 2.2 is empty:

https://docs.influxdata.com/influxdb/cloud/query-data/flux/manipulate-timestamps/

Need to include everything from here:

/influxdb/v2.1/query-data/guides/manipulate-timestamps/

lwandzura:
@sanderson  looks like the shortcode should pull data from another doc??

<img width="893" alt="image" src="https://user-images.githubusercontent.com/51929958/169579665-d6cb3229-0e39-40ca-9307-977793c695ee.png">


The page actually looks like this:

![image](https://user-images.githubusercontent.com/51929958/169579825-bdb6e878-46a5-4dc6-a9a8-fa2f0dae63f2.png)

I'll copy the info from 2.1 over

PR #4036, opened 2022-05-20
Fix manipulate timestamps

lwandzura:
Closes https://github.com/influxdata/docs-v2/issues/4034

@sanderson please make sure I changed the front matter correctly


sanderson (approved):
👌🏻

PR #4055, opened 2022-05-23
Expand on NFS restrictions in Enterprise documentation

sanderson:
Closes  influxdata/DAR#297

Call out specific cloud-based NFS-mountable file system services that users cannot use with InfluxDB Enterprise.

- [x] Rebased/mergeable


aaronkoerner (approved):
Looks good

PR #4104, opened 2022-06-08
Draft for DAR 166

peterreg:
Closes  /influxdata/DAR/issues/166

### DIFF
--- a/content/influxdb/cloud/organizations/users.md
+++ b/content/influxdb/cloud/organizations/users.md
@@ -24,7 +24,10 @@ By default, each user has full permissions on resources in your organization.
   - [Resend an invitation](#resend-an-invitation)
   - [Withdraw an invitation](#withdraw-an-invitation)
 - [Remove a user from your organization](#remove-a-user-from-your-organization)
-  - [Remove yourself from an organization](#remove-yourself-from-an-organization)
+
+{{% note %}}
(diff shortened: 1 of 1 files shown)


content/influxdb/cloud/organizations/users.md
```diff
@@ -68,6 +68,10 @@ Accounts can have up to 50 pending invitations at one time.
 
 ## Remove a user from your organization
 
+{{% warn %}}
+If you delete an InfluxDB Cloud user, the user cannot be reactivated. The user needs to create a new account with a new token.
```
peterreg, line 72: @kelseiv Does an admin need to tell a user that they need to create a new account, or does the user need to create their own account? It's not quite clear the distinction between user and account.

kelseiv: @peterreg, for this warning, let's say:
For security purposes, once an InfluxDB user account is deleted, the user (and their token) cannot be reactivated. 

A couple more things we can do on this doc:

- At the top of the page, add a tip to the top of the page: If you have multiple users, we recommend creating a system account for administrative purposes. For example, if you have applications writing to InfluxDB, ensure the applications use the system account credentials.
- Remove the broken link "Remove yourself from an organization" from the bullet list at top of doc --we removed this section a cpl months back - thank you!

kelseiv (approved):

PR #4116, opened 2022-06-14
fix: adding possible resource types

sunbryely-influxdata:
Closes #https://github.com/influxdata/DAR/issues/305

Add list of all possible resources to export

image

- [ ] Rebased/mergeable


content/influxdb/v2.2/reference/cli/influx/export/all.md
```diff
@@ -45,6 +45,26 @@ influx export all [flags]
...
+- NotificationEndpointHTTP     
+- NotificationEndpointPagerDuty
+- NotificationEndpointSlack    
+- NotificationRule             
+- Task                         
+- Telegraf                     
+- Variable
+
+For additional information on each resource, please see 
+[template resources](/influxdb/v2.2/influxdb-templates/#template-resources).
```
kelseiv, line 67: we may want to make expand/clarify groupings being we don't have 1:1 corollary

For example, here: https://docs.influxdata.com/influxdb/v2.2/influxdb-templates/#template-resources
we see HTTP, Slack, and PagerDuty endpoint--what diff is plain `NotificationEndpoint`vs `NotificationEndpointHTTP`?

kelseiv (approved):
LGTM- one comment/question-thanks!

sanderson (changes requested):

content/influxdb/v2.2/reference/cli/influx/export/all.md
sanderson: I take back my comment here. Those resource types are import to know when using the CLI. @kelseiv The list you linked to is more of a high level list. `NotificationEndpointHTTP`, `NotificationEndpointPagerDuty`, and `NotificationEndpointSlack` are implied in "notification endpoints."

sanderson (approved):

Issue #4120, opened 2022-06-14
Add alert about known bug in Replications in v2.2


samhld:
_Describe the issue here._

There is a known bug having to do with replications not persisting after remote connection is lost. This should be called out on release notes and the replications procedure page, I would think.

@kelseiv @sanderson @lwandzura Thoughts?

kelseiv:
We could add a warning in the InfluxDB 2.2 release notes, under [this heading](https://docs.influxdata.com/influxdb/v2.2/reference/release-notes/influxdb/#technical-preview-replicate-data-remotely).

Known issue: 
Occasionally, `remote-write` failures cause data in the replication queue to get stuck. 
<insert how they will know this is happening--will certain errors appear?>
To resolve this issue, restart the edge node <restart InfluxDB service?> to ensure remote writes are successfully processed.

samhld:
@kelseiv thanks!

I believe the only way they'll know it's happening is buy checking the presence of data in the target. You could also get a "hint" if the queue size is growing while your local bucket ingest is not....but again that's just a hint. 

Something like:
"""
#### Known issue:
On rare occasions, remote-write failures will cause the data in the replication queue to get stuck. You'll be able to identify this is happening by checking the presence of data you expect in the replication target bucket. A less exact indication that this may be happening would be a buildup of the queue size. You can view this size with `influx replication list[|ls\`.

Preceding the fix to this issue, you can remediate the issue by restarting your node. We recommend doing this earlier rather than later as the queue can build up, resulting in dropped data.
"""

PR #4123, opened 2022-06-14
draft-questions

kelseiv:
Closes https://github.com/influxdata/docs-v2/issues/4120


content/influxdb/v2.2/reference/release-notes/influxdb.md
```diff
@@ -25,12 +25,19 @@ This release includes the following new [features](#features) and several [bug f
...
+Add the option to [replicate data from InfluxDB OSS to InfluxDB Cloud](/influxdb/v2.2/write-data/replication).
 
-  This [**technical preview**](/influxdb/v2.2/reference/glossary/#technical-preview) feature lets you create a durable subscription on a per bucket basis for the purposes of replicating data on-write from an InfluxDB OSS instance to InfluxDB Cloud.
-  - Lets you store, analyze, and aggregate data locally while also forwarding newly arriving data to a centralized InfluxDB Cloud account.
-  - If InfluxDB OSS loses connectivity to InfluxDB Cloud, the data gets sent when network connectivity is reestablished.
-  - Configuration options include a maximum buffer size and data age restrictions to restrict the amount of data stored on disk.
+{{% warn %}}
+**Known issue:**
+On rare occasions, remote write failures cause data in the replication queue to get stuck. You'll be able to identify this is happening by checking the presence of data you expect in the replication target bucket. A less exact indication that this may be happening would be a buildup of the queue size. You can view this size with `influx replication list`.
```
kelseiv, line 32: if this only happens on rare occasions, are we recommending they periodically check the replication target bucket programmatically? (Ideally, we want to be as explicit about how to do this proactively)....rather than say, _occurs rarely, but you'll need to figure out when to check_

For example:
On rare occasions, remote write failures cause data in the replication queue to get stuck. To prevent this issue from occurring, we recommend that you do the following <insert recommended cadence, depending on the amount of data being written?…> :
- Verify the expected data appears in the replication target bucket.
- View this queue size (of the replication target bucket?) using `influx replication list`. (If this queue exceeds ?,

content/influxdb/v2.2/reference/release-notes/influxdb.md
```diff
@@ -25,12 +25,19 @@ This release includes the following new [features](#features) and several [bug f
...
+
+Prior to the fix, you can remediate the issue by restarting your node. We recommend doing this earlier rather than later as the queue can build up to a max, resulting in dropped data.
```
kelseiv, line 34: earlier rather than later is a bit vague--any specifics we could add here would be helpful, for example, if you're writing xyz amount of data daily, we recommend checking hourly....or anything along those lines

content/influxdb/v2.2/reference/release-notes/influxdb.md
samhld: We could do that but I'm also a little worried about exposing that we don't have a good way to monitor this yet. If you think the benefit of being helpful outweighs the optics cost here, I'm good with it.

If we do it, the recommended cadence is going to be hard to pin down but we could say something like, "cadence is up to. Just keep in mind your ingest rate and how fast the queue bytes will build up given that."

content/influxdb/v2.2/reference/release-notes/influxdb.md
samhld: We could simply say "ASAP". In truth, we don't know what the risk level is so it's really hard to recommend a cadence because of that so I think all we can say is to keep in mind how long it would take their queue (size they configured) to fill up given their ingest volume to the buckets they've configured to replicate.

content/influxdb/v2.2/reference/release-notes/influxdb.md
kelseiv: I believe the more straightforward we are (to the extent possible), the better. 
Regarding optics costs, I don't think we're fooling anyone. Ambiguous text is apt to generate questions/confusion. Our best gauge --can support team address this issue if need be.

content/influxdb/v2.2/reference/release-notes/influxdb.md
```diff
@@ -25,12 +25,20 @@ This release includes the following new [features](#features) and several [bug f
...
 
-- Add the option to [Replicate data from InfluxDB OSS to InfluxDB Cloud](/influxdb/v2.2/write-data/replication).
+Add the option to [replicate data from InfluxDB OSS to InfluxDB Cloud](/influxdb/v2.2/write-data/replication).
 
-  This [**technical preview**](/influxdb/v2.2/reference/glossary/#technical-preview) feature lets you create a durable subscription on a per bucket basis for the purposes of replicating data on-write from an InfluxDB OSS instance to InfluxDB Cloud.
-  - Lets you store, analyze, and aggregate data locally while also forwarding newly arriving data to a centralized InfluxDB Cloud account.
-  - If InfluxDB OSS loses connectivity to InfluxDB Cloud, the data gets sent when network connectivity is reestablished.
-  - Configuration options include a maximum buffer size and data age restrictions to restrict the amount of data stored on disk.
+{{% warn %}}
+On rare occasions, remote write failures may cause data in the replication queue to get stuck. To ensure data is not dropped, restart your node as soon as possible.
```
kelseiv, line 31:
```suggestion
On rare occasions, remote write failures may cause data in the replication queue to get stuck. To ensure data is not dropped, restart the InfluxDB instance you're writing to as soon as possible.
```

content/influxdb/v2.2/reference/release-notes/influxdb.md
```diff
@@ -25,12 +25,20 @@ This release includes the following new [features](#features) and several [bug f
...
-  - Lets you store, analyze, and aggregate data locally while also forwarding newly arriving data to a centralized InfluxDB Cloud account.
-  - If InfluxDB OSS loses connectivity to InfluxDB Cloud, the data gets sent when network connectivity is reestablished.
-  - Configuration options include a maximum buffer size and data age restrictions to restrict the amount of data stored on disk.
+{{% warn %}}
+On rare occasions, remote write failures may cause data in the replication queue to get stuck. To ensure data is not dropped, restart the InfluxDB instance you're replicating data to as soon as possible.
+
+To assess whether this issue is occurring, we recommend periodically doing one of the following:
+- Verify your data is successfully replicated to your target bucket. 
+- View your queue size using [`influx replication list`](/influxdb/cloud/reference/cli/influx/replication/list/) for unexpected growth.
+{{% /warn %}}
```
kelseiv, line 36: @samhld , updated per your edits, please lmk if you have any more adjustments. Thank you!

PR #4130, opened 2022-06-16
Correct replication instructions to use bucket id

JamesGibo:
Correct mistake in docs that use a command line argument that does not exist.


JamesGibo:
!signed-cla

sanderson (approved):
Thanks @JamesGibo!

Issue #4127, opened 2022-06-15
formatting issue on replicate data page


lwandzura:
Under "Configure a replication stream", the steps jump from 2 to 7:
![image](https://user-images.githubusercontent.com/51929958/173900449-48a7b1a9-7729-4952-9bd1-77d271203b82.png)

https://docs.influxdata.com/influxdb/v2.2/write-data/replication/

PR #4131, opened 2022-06-16
added edge data replication to title, fixed formatting

lwandzura:
Closes https://github.com/influxdata/docs-v2/issues/4127

@kelseiv I tried to change the numbering, but it still renders as a 7 and not 3, as it should, under "configure a replication stream".  No matter what I tried, it keeps going from  2 to 7 - Can you take a look?  I am sure there's a few spaces that are causing the problem.

### DIFF
--- a/content/influxdb/v2.2/write-data/replication.md
+++ b/content/influxdb/v2.2/write-data/replication.md
@@ -1,5 +1,5 @@
 ---
-title: Replicate data to a remote InfluxDB instance
+title: Replicate data to a remote InfluxDB instance (InfluxDB Edge Data Replication )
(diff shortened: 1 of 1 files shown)


sanderson:
> I tried to change the numbering, but it still renders as a 7 and not 3, as it should, under "configure a replication stream". No matter what I tried, it keeps going from 2 to 7 - Can you take a look? I am sure there's a few spaces that are causing the problem.

@lwandzura This was a styling issue. I just pushed a fix to this branch.

kelseiv (approved):

kelseiv (approved):

PR #4135, opened 2022-06-17
adjust influxdb1.8 and telegraf to be consistent; add repo instructions for 2.2

jdstrand:
The install instructions have been inconsistent within this repo as well as the downloads portal. In addition to being inconsistent, the instructions lacked security features with apt setup. The private https://github.com/influxdata/influxdb-customers/pull/919 PR adjusts the downloads portal to make all apt downloads consistent and secure. This PR:
* adjusts telegraf install to be like the downloads portal
* adjusts influxdb 1.8 install to be like the downloads portal
* adjusts influxdb 2.2 install to include apt/yum repo configuration (the preferred distribution method since it allows for users to get security updates easily). These instructions are consistent with the downloads portal

Tested all updates against Ubuntu 20.04 and CentOS 8.

### DIFF
--- a/content/influxdb/v2.2/install.md
+++ b/content/influxdb/v2.2/install.md
@@ -212,20 +212,40 @@ For information about installing the `influx` CLI, see
 
 ### Install InfluxDB as a service with systemd
 
-1.  Download and install the appropriate `.deb` or `.rpm` file using a URL from the
-    [InfluxData downloads page](https://portal.influxdata.com/downloads/)
+1.  Download and install the appropriate `.deb` or `.rpm` file using either:
+    * `apt` or `yum` with the following commands:
+        ```sh
+        # influxdb.key GPG Fingerprint: 05CE15085FC09D18E99EFB22684A14CF2582E0C5
+
+        # Ubuntu/Debian
+        wget -q https://repos.influxdata.com/influxdb.key
+        echo '23a1c8836f0afc5ed24e0486339d7cc8f6790b83886c4c96995b88a061c5bb5d influxdb.key' | sha256sum -c && cat influxdb.key | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/influxdb.gpg > /dev/null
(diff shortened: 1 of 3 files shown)


content/influxdb/v2.2/install.md
```diff
@@ -212,20 +212,40 @@ For information about installing the `influx` CLI, see
...
+        ```sh
+        # Ubuntu/Debian
+        wget https://dl.influxdata.com/influxdb/releases/influxdb2-{{< latest-patch >}}-xxx.deb
+        sudo dpkg -i influxdb2-{{< latest-patch >}}-xxx.deb
+
+        # Red Hat/CentOS/Fedora
+        wget https://dl.influxdata.com/influxdb/releases/influxdb2-{{< latest-patch >}}-xxx.rpm
+        sudo yum localinstall influxdb2-{{< latest-patch >}}-xxx.rpm
+        ```
+        _Use the exact filename of the download of `.rpm` package (for example, `influxdb2-{{< latest-patch >}}-amd64.rpm`)._
```
powersj, line 248: for what it is worth, on the downloads page we used to list these same wget-based instructions and appear to have switched to only showing the repo-based instructions. As such it seems a little inconsistent to have wget-instructions here and only here. The change for [influxdb2](https://github.com/influxdata/influxdb-customers/pull/858/files) and for [telegraf](https://github.com/influxdata/influxdb-customers/commit/e6cf52688eda77144f3f640b116fb5ff4e24ded2)

I realize this isn't the purpose of this change and you are only keeping the existing content, but part of me says if we are trying to make things consistent, then this block should be removed as well.

Thoughts?

kelseiv: thanks for surfacing this issue, @powersj 
+1 on being consistent
@sjwang90 @samhld, how does product want to handle installs? And then, docs will follow/be consistent across all products

sjwang90: I'm all good with these changes. We're trying to get influxdb/telegraf and downloads/docs consistent. @powersj just removed the wget from the downloads pages so we should move them from docs too.

PR #4138, opened 2022-06-17
Flux function type signatures

sanderson:
- [x] Rebased/mergeable


content/flux/v0.x/function-type-signatures.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Basic transformation
+The following type signature describes a
+[transformation](/flux/v0.x/function-types/#transformations) that:
+
+- Takes a stream of tables as piped-forward input
+- Returns a new, modified stream of tables
+
+```js
+(<-tables: stream[A]) => stream[B]
```
nathanielc, line 172: This is a discouraged pattern in Flux. However only built-ins can do this. We call this _return type polymorphism_ and its generally bad. Notice how `B` only shows up on the return side of the type but not the input side? This means Flux cannot connect the output type to the input types in any way. Which means that Flux does not know anything about the return type, and so therefore can no longer provide the user with useful error messages. Most languages do not even allow for this kind of thing. Flux does in order to be backwards compatible with its pre-typed days. 

So this is example is accurate and it does occur in the wild, however the Flux team is actively trying to reduce/eliminate these cases. 


Perhaps a better although slightly more complex example would be map?

```
builtin map : (<-tables: stream[A], fn: (r: A) => B,) => stream[B]
```

Notice how its very similar in that you have a `stream[A]` as input and `stream[B]` as output. But this time we also have `fn` which tells us how `A` relates to `B`.  Since `B` occurs on the input side with this signature Flux can determine exactly what `B` is and therefore we can continue to provide useful error messages/hints to users.

Happy to chat about this more if you want more details.

content/flux/v0.x/function-type-signatures.md
```diff
@@ -0,0 +1,188 @@
...
+The following type signature describes a
+[transformation](/flux/v0.x/function-types/#transformations) that:
+
+- Takes a stream of tables as piped-forward input
+- Has a required **tag** parameter
+- Returns a new, modified stream of tables that includes a **tag** column with
+  string values
+
+```js
+(<-tables: stream[A], tag: B) => stream[{A with tag: string}]
```
nathanielc, line 187: This one confuses me a bit, what is the purpose of the type variable `B`? Should this be instead:


```
(<-tables: stream[A], tag: string) => stream[{A with tag: string}]
```

Meaning why would we allow the `tag` parameter to be any type if we intend to only place a string on the stream data?

nathanielc (changes requested):
Overall this is amazing,  Thanks for writing up a good explanation of type in Flux. I have some questions and suggestions below.

Also a link to the SPEC here https://github.com/influxdata/flux/blob/master/docs/SPEC.md#type-expressions would be good.

content/flux/v0.x/function-type-signatures.md
sanderson: Sorry, this was the signature from a function that applies `string()` to the `tag` column. It should have a constraint, but it wasn't output when I created it. I'll add the constraint.

PR #4159, opened 2022-06-23
Updates to Install page Windows download

### LINKED ISSUE #4056: Download/Install instructions no longer match download page
The referenced downloads page has changed and the docs no longer reflect the way a user would go about installing InfluxDB2
https://portal.influxdata.com/downloads/ now uses exclusively repositories to install on Linux

Additionally links like `https://dl.influxdata.com/influxdb/releases/influxdb2-2.2.0-amd64.rpm` look to no longer exist and thus cannot be used for downloads.

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2.2/install/

peterreg:
Closes #4056

This updates the Install InfluxDB Windows download button and the related links below it.

Should it be a concern that the Install page Windows download link does not exactly mirror the Windows link on the Download portal? 

Install page: https://dl.influxdata.com/influxdb/releases/influxdb2-2.3.0.x86_64.rpm
Downloads portal: https://dl.influxdata.com/influxdb/releases/influxdb2-2.3.0-windows-amd64.zip

### DIFF
--- a/content/influxdb/v2.3/install.md
+++ b/content/influxdb/v2.3/install.md
@@ -411,13 +411,13 @@ For information about installing the `influx` CLI, see
 [Install and use the influx CLI](/influxdb/v2.3/tools/influx-cli/).
 {{% /note %}}
 
-<a class="btn download" href="https://dl.influxdata.com/influxdb/releases/influxdb2-{{< latest-patch >}}.x86_64.rpm" download >InfluxDB v{{< current-version >}} (Windows)</a>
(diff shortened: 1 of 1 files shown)


jdstrand (changes requested):
`rpm` is not the correct extension for Windows installs. I believe the issue is that while `zip` (the currently documented extension` is the correct extension, the 2.3.0 produced a `tar.gz` for Windows by mistake. AIUI, the Windows portion does not need to change; the uploaded artifact needs to. @bnpfeife can confirm.

peterreg:
Thanks @jdstrand, I just tested the Windows download button again and it looks like the link has been fixed and the package is now a .zip file. Thanks @bnpfeife!

I'm closing this.

PR #4190, opened 2022-07-07
Chore/improve api docs generation

jstirnaman:
getswagger.sh improvements:

- Exit on invalid options.
- Show curl output and errors (e.g. when the filepath is incorrect).
- Replace $_ with $outFile for readability.
- Replace context with platform, the common jargon.

generate.sh improvements:

- Speed up dev and deployments by only regenerating HTML if the current file differs from master.
- Refactor to remove code repetition.

- [x] Rebased/mergeable


jstirnaman:
@sunbryely-influxdata Please try this for a while to shake out any potential problems.

sunbryely-influxdata (approved):

PR #4201, opened 2022-07-12
telegraf 1.23.2

kelseiv:


content/telegraf/v1.23/release-notes-changelog.md
```diff
@@ -11,6 +11,18 @@ menu:
     weight: 60
 ---
 
+## v1.23.2 [2022-7-11]
+
+## Bug fixes
+- Remove unexpected deprecation warnings for non-deprecated packages that occurred in 1.23.1.
+- `common.http`: Allow 201 responses for cookies. Also update the cookie header docs to show a TOML map rather than a string.
```
kelseiv, line 18: @sjwang90 is this the HTTP response plugin, or all HTTP input plugins?

kelseiv, line 18: shall we say allows 200 and 201 responses here?

content/telegraf/v1.23/release-notes-changelog.md
```diff
@@ -11,6 +11,18 @@ menu:
     weight: 60
 ---
 
+## v1.23.2 [2022-7-11]
+
+## Bug fixes
+- Remove unexpected deprecation warnings for non-deprecated packages that occurred in 1.23.1.
+- HTTP input plugin (`inputs.http`): Allow both 200 and 201 response codes when generating authentication using the cookie authentication plugin. Also update the cookie header docs to show a TOML map rather than a string.
```
kelseiv, line 18: @sjwang90, I couldn't find a cookie auth plugin in our list--is there another name?

sjwang90: This [cookie common plugin](https://github.com/influxdata/telegraf/tree/master/plugins/common/cookie). These aren't plugins that users interact with but that an actual plugin, in this case HTTP input plugin, would reference.

kelseiv: thank you, @sjwang90 --good conceptual info.
Which is preferable:

- Link to internal plugin
- Clarify this is an internal plugin that HTTP input references
- Omit explicit reference, and say something like: "_Allow both 200 and 201 response codes when generating cookie authentication_"

sjwang90: i think the last bullet point is good and covers everything someone would reading this would need to know. 
> Omit explicit reference, and say something like: "Allow both 200 and 201 response codes when generating cookie authentication"

PR #4241, opened 2022-07-21
hotfix:400-error

sunbryely-influxdata:
Closes #https://github.com/influxdata/DAR/issues/312, https://github.com/influxdata/openapi/pull/447

- Hotfix Kelly and I worked on to update 400 partial write error description in the API docs.
- Updates to Scripts API are also included since I pulled all changes instead of just cherry picking the single hotfix commit.

- [ ] Rebased/mergeable

### DIFF
--- a/api-docs/cloud/ref.yml
+++ b/api-docs/cloud/ref.yml
@@ -4144,13 +4144,12 @@ components:
     ScriptCreateRequest:
       properties:
         description:
+          description: Script description. A description of the script.
           type: string
         language:
           $ref: '#/components/schemas/ScriptLanguage'
         name:
-          description: >-
-            The name of the script. The name must be unique within the
-            organization.
+          description: Script name. The name must be unique within the organization.
           type: string
         script:
           description: The script to execute.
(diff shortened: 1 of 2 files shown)


api-docs/cloud/ref.yml
```diff
@@ -4144,13 +4144,12 @@ components:
     ScriptCreateRequest:
       properties:
         description:
+          description: Script description. A description of the script.
```
sanderson, line 4147: Seems redundant

```suggestion
          description: Script description.
```

api-docs/cloud/ref.yml
```diff
@@ -11039,35 +11038,135 @@ paths:
         - System information endpoints
   /api/v2/scripts:
     get:
+      description: >
+        Retrieves a list of
+        [scripts](/influxdb/cloud/api-guide/api-invokable-scripts/).
```
sanderson, line 11043:
```suggestion
        [invokable scripts](/influxdb/cloud/api-guide/api-invokable-scripts/).
```

api-docs/cloud/ref.yml
```diff
@@ -11039,35 +11038,135 @@ paths:
...
+
+
+        #### Limitations
+
+
+        - Paging with an `offset` greater than the number of records will result
+        in
+
+        an empty response--for example:
```
sanderson, line 11052: Is the extra whitespace here on purpose?

api-docs/cloud/ref.yml
```diff
@@ -14635,9 +14810,14 @@ tags:
...
-      missing or malformed or the API token does not have permission for the
-      operation. |
+      | `400`       | Bad request              | May indicate one of the
+      following: <li>Line protocol is malformed. The response body contains the
+      first malformed line in the data and indicates what was expected. For
+      partial writes, the number of points written and the number of points
+      rejected are also included. For more information, check the
+      `rejected_points` measurement in your `_monitoring`
+      bucket.</li><li>`Authorization` header is missing or malformed or the API
+      token does not have permission for the operation.</li> |
```
sanderson, line 14820: I know redoc will take care of it, but I think we should make the inline HTML valid.

```suggestion
      | `400`       | Bad request              | May indicate one of the
      following: <ul><li>Line protocol is malformed. The response body contains the
      first malformed line in the data and indicates what was expected. For
      partial writes, the number of points written and the number of points
      rejected are also included. For more information, check the
      `rejected_points` measurement in your `_monitoring`
      bucket.</li><li>`Authorization` header is missing or malformed or the API
      token does not have permission for the operation.</li></ul> |
```

api-docs/v2.3/ref.yml
```diff
@@ -16494,9 +16499,14 @@ tags:
...
-      missing or malformed or the API token does not have permission for the
-      operation. |
+      | `400`       | Bad request              | May indicate one of the
+      following: <li>Line protocol is malformed. The response body contains the
+      first malformed line in the data and indicates what was expected. For
+      partial writes, the number of points written and the number of points
+      rejected are also included. For more information, check the
+      `rejected_points` measurement in your `_monitoring`
+      bucket.</li><li>`Authorization` header is missing or malformed or the API
+      token does not have permission for the operation.</li> |
```
sanderson, line 16509: Same suggestion as above.

sanderson (approved):
Some minor suggestions. I know these were auto-gen'd, so take them with a grain of salt :)

api-docs/cloud/ref.yml
jstirnaman: Let's keep `scripts` as they're likely to decouple from the invokable-ness.

api-docs/cloud/ref.yml
jstirnaman: I think 

> Is the extra whitespace here on purpose?

I think it gets added by openapi-cli.

api-docs/cloud/ref.yml
jstirnaman: > I know redoc will take care of it, but I think we should make the inline HTML valid.

I'll fix this in my next openapi PR

jstirnaman (approved):

PR #4260, opened 2022-07-27
update sortBy to orderBy

kelseiv:
https://influxdata.slack.com/archives/C0VAV6N92/p1657764936862999 Andrew originated this request, and I see the parameter was updated in our Flux docs

@sanderson - updated geo files ~and sample data json files~

@jstirnaman, looks like we still have `sortBy` in the api docs -- I imagine we may need to update those to `orderBy` as well?

### DIFF
--- a/static/downloads/influxdb-enterprise-monitor-dashboard.json
+++ b/static/downloads/influxdb-enterprise-monitor-dashboard.json
@@ -102,7 +102,7 @@
 				"legend": {},
 				"tableOptions": {
 					"verticalTimeAxis": false,
-					"sortBy": {
+					"orderBy": {
 						"internalName": "",
 						"displayName": "",
 						"visible": false
@@ -213,7 +213,7 @@
 				"legend": {},
 				"tableOptions": {
 					"verticalTimeAxis": false,
-					"sortBy": {
+					"orderBy": {
 						"internalName": "",
 						"displayName": "",
 						"visible": false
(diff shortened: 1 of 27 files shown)


static/downloads/influxdb-enterprise-monitor-dashboard.json
```diff
@@ -102,7 +102,7 @@
 				"legend": {},
 				"tableOptions": {
 					"verticalTimeAxis": false,
-					"sortBy": {
+					"orderBy": {
```
sanderson, line 105: These shouldn’t have been replaced.

static/downloads/influxdb-oss-monitor-dashboard.json
```diff
@@ -102,7 +102,7 @@
 				"legend": {},
 				"tableOptions": {
 					"verticalTimeAxis": false,
-					"sortBy": {
```
sanderson, line 105: These shouldn’t have been replaced

static/downloads/kapacitor-monitor-dashboard.json
```diff
@@ -102,7 +102,7 @@
 				"legend": {},
 				"tableOptions": {
 					"verticalTimeAxis": false,
-					"sortBy": {
+					"orderBy": {
```
sanderson, line 105: These shouldn’t have been replaced

sanderson (changes requested):

static/downloads/influxdb-enterprise-monitor-dashboard.json
kelseiv: thanks, @sanderson - will revert!

sanderson (approved):

PR #4274, opened 2022-07-28
fix: json example, string example

helenosheaa:
- Removes the word cloud as we are now referring to it as native subscription
- Cleans up some strange "" in the JSON object
- Updates so the json path example points to the key
- Updates string example so it doesn't include info about json parsing

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/v2.3/write-data/no-code/load-data.md
+++ b/content/influxdb/v2.3/write-data/no-code/load-data.md
@@ -204,7 +204,7 @@ Associate **String** key/value pairs with **InfluxDB elements** (measurements, t
    {{% /warn %}}
 
   2. Under **Measurement**, if the string is `device_type=temperature_sensor` use regex to find the measurement name. For example:
-      - `device_type=([\s\S]*?)\n` captures the value after the `=` until the EOL (end of line) is reached) in this case the value would be temperature_sensor.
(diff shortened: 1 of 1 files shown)


content/influxdb/v2.3/write-data/no-code/load-data.md
```diff
@@ -203,12 +203,12 @@ Associate **String** key/value pairs with **InfluxDB elements** (measurements, t
    **Important**: Configure the timestamp format that matches the format in your messages.
    {{% /warn %}}
 
-  2. Under **Measurement**, enter the **JSON path** (start with `$.`) to assign the InfluxDB measurement key, for example, `$.device_id` or `$.device_information.device_id` for a nested measurement key.
+  2. Under **Measurement**, if the string is `device_type=temperature_sensor` use regex to find the measurement name. For example:
+      - `device_type=([\s\S]*?)\n` captures the value after the `=` until the EOL (end of line) is reached) in this case the value would be temperature_sensor.
```
kelseiv, line 207:
```suggestion
      - `device_type=([\s\S]*?)\n` captures the value after the `=` until the EOL (end of line) is reached), in this case the value would be `temperature_sensor`.
```

kelseiv (approved):

genehynson (approved):

garylfowler (approved):

PR #4315, opened 2022-08-08
Fixed issue in JSON sample

sjwang90:
missing comma causes errors when reading in with new native subscriptions feature


sanderson (approved):
Thanks @sjwang90

PR #4322, opened 2022-08-09
Update fieldkeys.md

ffisc:
Closes #

- [ ] Rebased/mergeable


sanderson:
Thanks @ffisc! The Flux standard library docs are generated from the Flux source code, so this change would be overwritten the next time a new Flux release rolls out. I've updated in the inline documentation as part of this PR: https://github.com/influxdata/flux/pull/5061. Once merged and the documentation is generated, this will be fixed.

Thanks for pointing this out.

PR #4359, opened 2022-08-18
update download links

kelseiv:
hotfix for https://github.com/influxdata/DAR/issues/319


sanderson (approved):
We need to be sure to port these changes to the `enterprise-1.10` branch.

aaronkoerner (changes requested):
It's just the amd64.deb link that seems to be wrong, at least looking at 1.9.6 and 1.9.8 versions. Doesn't seem to follow the same convention as the x86_64.rpm URLs

aaronkoerner:
For reference, I'm testing with 
works: `curl https://dl.influxdata.com/enterprise/releases/influxdb-data_1.9.6-c1.9.6_amd64.deb`
doesn't work ("The specified key does not exist"): `curl https://dl.influxdata.com/enterprise/releases/influxdb-data_1.9.8_c1.9.8_amd64.deb`

PR #4375, opened 2022-08-22
Native subscription updates

### LINKED ISSUE #4275: Expand Native Subscriptions Docs / Examples
We want more robust parsing examples in the docs that we can link to from the InfluxDB Cloud UI. 

Instead of adding more examples to the existing [Load data source in UI](https://docs.influxdata.com/influxdb/cloud/write-data/no-code/load-data/) page we want to add a separate "Native Subscriptions" section and parsing guide so we can easily add more examples that would be helpful for our end users. 

To do: 
1. Add "Native Subscriptions" bullet under No Code section of of Write Data
2. Add new "JSON Parsing Guide" page. Include existing [parsing example](https://docs.influxdata.com/influxdb/cloud/write-data/no-code/load-data/?t=JSON#define-parsing-rules). 

- No-code solutions
  - Native Subscriptions
    - JSON Parsing Guide
  - Telegraf (agent)
  - Third-party technologies

image

@garylfowler

noramullen1:
Closes #4275 . Expanding/bumping up these docs as part of this week's marketing launch. Note that right now the only subscription is MQTT, but more are coming eventually, so page and examples are set up for that.


noramullen1:
> 2\. It will also remain Native Subscriptions in the UI.

We discussed over slack but repeating here -- going to keep it matching the UI text per tech writing princples (document what the product actually says).

sjwang90 (approved):

kelseiv (approved):

PR #4388, opened 2022-08-24
Update 'exists' guide with correct static example, 

### LINKED ISSUE #4387: Example for `exists` produces an error
Running the first example from the page on [`Check if a value exists`](https://docs.influxdata.com/influxdb/cloud/query-data/flux/exists/)
```
$ flux
> p = {firstName: "John", lastName: "Doe", age: 42}
> exists p.height
error: record is missing label height
  ┌─ main:1:8
  │
1 │ exists p.height
  │        ^
```
The docs indicate that this would produce `false`, but instead it produces this error.

This error is expected because the type of `p` is known statically. If instead you issue a query against an external data source like
```js
from |> range |> filter
|> map(fn: (r) => ({r with cnt: if exists r.cnt then r.cnt + 1 else 0}))
```
Then Flux doesn’t know statically (before querying InfluxDB) if the `cnt` field exists, and so this query succeeds. I guess the docs need to call out this subtlety since it can be confusing.

##### Relevant URLs
- https://docs.influxdata.com/influxdb/cloud/query-data/flux/exists/

sanderson:
Closes #4387

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/v2.4/query-data/flux/exists.md
+++ b/content/influxdb/v2.4/query-data/flux/exists.md
@@ -3,8 +3,8 @@ title: Check if a value exists
 seotitle: Use Flux to check if a value exists
 list_title: Exists
 description: >
-  Use the Flux `exists` operator to check if a record contains a key or if that
-  key's value is `null`.
+  Use the Flux `exists` operator to check if a row record contains a column or if
+  that column's value is `null`.
 influxdb/v2.4/tags: [exists]
 menu:
   influxdb_2_4:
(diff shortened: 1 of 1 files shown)


content/influxdb/v2.4/query-data/flux/exists.md
```diff
@@ -3,8 +3,8 @@ title: Check if a value exists
 seotitle: Use Flux to check if a value exists
 list_title: Exists
 description: >
-  Use the Flux `exists` operator to check if a record contains a key or if that
-  key's value is `null`.
+  Use the Flux `exists` operator to check if a row record contains a column or if
```
wolffcm, line 6: Did you mean to remove "record" and just write "... if a row contains ..." here? Not sure if "row record" is a term we're using in the docs.

wolffcm (approved):
Looks good to me, just had a question about "row record."

content/influxdb/v2.4/query-data/flux/exists.md
sanderson: It's a term we've started to use sparingly anytime row data is represented as a record (`r`) like in `filter`, `map`, and `reduce`.

sanderson: But it's something we can back off on as well.

wolffcm: I don't have a strong opinion on it. I just wasn't sure if it was a typo. We've called the type of `r` a few different things in that situation, including "row", "record" and even "object" (at least in our code). I guess that readers should just understand that they're all the same thing.

PR #4458, opened 2022-09-15
fix(api): expand URL replacement to descriptions in schema nodes

jstirnaman:
Check every node in the spec for a description field:
- Replace named nodes in decorator with the `any` key.
- If the node description field is a string (and not an object named "description"), replace URLs in the description.

- [ ] Rebased/mergeable


sanderson (approved):
👌🏻 much simpler

jstirnaman:
> 👌🏻 much simpler

and not more dangerous, I hope :)

PR #4528, opened 2022-10-05
fix: add imports to query script

sunbryely-influxdata:
Closes #
New user feedback!
Page: https://docs.influxdata.com/influxdb/cloud/api-guide/client-libraries/python/
Feedback: The code example doesn't work. I get an Atribute error

I ran through all the example code on the page and was not able to recreate the error the user found.
One possibility i saw was that our `query` code example doesn't include import statements so I added that into the script, but that shouldn't be the cause for the error the user is seeing themselves.

Additional information from the user would be needed on this feedback.  

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/v2.4/api-guide/client-libraries/python.md
+++ b/content/influxdb/v2.4/api-guide/client-libraries/python.md
@@ -114,11 +114,11 @@ write_api.write(bucket=bucket, org=org, record=p)
 2. Create a Flux query, and then format it as a Python string.
 
    ```python
-   query = ' from(bucket:"my-bucket")\
+   query = 'from(bucket:"my-bucket")\
    |> range(start: -10m)\
    |> filter(fn:(r) => r._measurement == "my_measurement")\
-   |> filter(fn: (r) => r.location == "Prague")\
-   |> filter(fn:(r) => r._field == "temperature" ) '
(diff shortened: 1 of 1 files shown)


content/influxdb/v2.4/api-guide/client-libraries/python.md
```diff
@@ -157,12 +158,28 @@ write_api.write(bucket=bucket, org=org, record=p)
...
+    org=org
+)
+
+# Query script
 query_api = client.query_api()
-query = ‘ from(bucket:"my-bucket")\
+query = 'from(bucket:"my-bucket")\
 |> range(start: -10m)\
 |> filter(fn:(r) => r._measurement == "my_measurement")\
 |> filter(fn: (r) => r.location == "Prague")\
```
kelseiv, line 181: does `(fn:(r)` vs `(fn: (r)` make a difference?

kelseiv (approved):

content/influxdb/v2.4/api-guide/client-libraries/python.md
sunbryely-influxdata: Thanks for catching the formatting.  it does not make a difference or cause an error though

Issue #4563, opened 2022-10-14
The example CSV file for defining an explicit schema bucket is invalid


barbaranelson:
If you create a CSV file using the example on this page (usage-resources.csv) and try to apply it to a cloud 2 bucket, you will get an error - csv headers are not correct: they need to be: "name, type, dataType"

If you change data_type to dataType and re-upload the schema file, you will get an error - column file is not formatted correctly.

##### Relevant URLs
- _Provide relevant URLs (the doc page in question, project issues, community threads, etc.)_
https://docs.influxdata.com/influxdb/cloud/organizations/buckets/bucket-schema/

sanderson:
Thanks @barbaranelson. The CSV example did include an incorrect header. The error after that appears to have be causes by a newline at the end of the CSV. I removed it and validated that the updated example does work.

PR #4568, opened 2022-10-18
fix: the cli expects snake case for data_type in csv

jeffreyssmith2nd:
Closes https://github.com/influxdata/influxdb/issues/23807

- [x] Rebased/mergeable

### DIFF
--- a/shared/text/bucket-schema/usage-resources.csv
+++ b/shared/text/bucket-schema/usage-resources.csv
@@ -1,4 +1,4 @@
-name,type,dataType
+name,type,data_type
 time,timestamp,
 host,tag,
 service,tag,


sanderson:
@jeffreyssmith2nd While I'm ok with this change, it does contradict https://github.com/influxdata/docs-v2/issues/4563 (and https://github.com/influxdata/docs-v2/commit/b69ca52fda5b2078e9f57772c0701464d7b3d58f). It appears there is a product issue here. The UI and CLI seem to be expecting different headers in the CSV.

jeffreyssmith2nd:
> @jeffreyssmith2nd While I'm ok with this change, it does contradict #4563 (and [b69ca52](https://github.com/influxdata/docs-v2/commit/b69ca52fda5b2078e9f57772c0701464d7b3d58f)). It appears there is a product issue here. The UI and CLI seem to be expecting different headers in the CSV.

Ah that is interesting. That implies to me that the CLI just doesn't work as is then, which is a good argument for changing in there instead.

sanderson:
It may not be the CLI. I ran into an issue yesterday where none of the CSV files I was using to test in the UI were working with either header. I think there may be issues in how the UI is parsing bucket-schema CSV files that are uploaded.

jeffreyssmith2nd:
Closing in favor of https://github.com/influxdata/influx-cli/pull/449

Issue #4583, opened 2022-10-20
InfluxDB OSS 2.5 release notes


lwandzura:
**ETA:** October 21, 2022

InfluxDB OSS 2.5 changelog:

### Features
1. [485968c](https://github.com/influxdata/influxdb/commit/485968c): Unpin ui to point at latest
1. [b72848d](https://github.com/influxdata/influxdb/commit/b72848d): Optimize saving changes to fields.idx
1. [f36646d](https://github.com/influxdata/influxdb/commit/f36646d): Bump to latest UI

### Bug Fixes
1. [daaf866](https://github.com/influxdata/influxdb/commit/daaf866): Several minor quality issues
1. [ee8ca45](https://github.com/influxdata/influxdb/commit/ee8ca45): Use copy of loop variable in parallel test
1. [11019d2](https://github.com/influxdata/influxdb/commit/11019d2): Check that user IDs are not in use in user create
1. [b87deb4](https://github.com/influxdata/influxdb/commit/b87deb4): Don't allow creating an auth with instance resources
1. [b51fefd](https://github.com/influxdata/influxdb/commit/b51fefd): Set limited permissions on package installs
1. [663d43d](https://github.com/influxdata/influxdb/commit/663d43d): Allow backup of all buckets
1. [a0f1184](https://github.com/influxdata/influxdb/commit/a0f1184): Manually scheduled task runs now run when expected
1. [4ed184d](https://github.com/influxdata/influxdb/commit/4ed184d): Fixes an error querying virtual dbrps
1. [2ad8995](https://github.com/influxdata/influxdb/commit/2ad8995): Improve delete speed when a measurement is part of the predicate
1. [b0430cf](https://github.com/influxdata/influxdb/commit/b0430cf): Sql scan error on remote bucket id when replication to 1.x

### Other
1. [accce86](https://github.com/influxdata/influxdb/commit/accce86): Chore: update CHANGELOG_frozen.md for 2.4
1. [785a465](https://github.com/influxdata/influxdb/commit/785a465): Refactor: remove reference to flux.Spec in query tests
1. [728070e](https://github.com/influxdata/influxdb/commit/728070e): Chore: upgrade Rust to 1.63.0
1. [aa9c49e](https://github.com/influxdata/influxdb/commit/aa9c49e): Build(flux): update flux to v0.180.1
1. [8f15620](https://github.com/influxdata/influxdb/commit/8f15620): Build(flux): update flux to v0.181.0
[...]

Issue #4582, opened 2022-10-20
Influx cli 2.5 release notes


lwandzura:
**ETA:** October 21, 2022

See changelog for all PRs associated with this release:

### Bug Fixes
1. [3a593e7](https://github.com/influxdata/influx-cli/commit/3a593e7): Don't allow creating an auth with instance resources
1. [e2aa4d2](https://github.com/influxdata/influx-cli/commit/e2aa4d2): Fix stack error typo
1. [77cca94](https://github.com/influxdata/influx-cli/commit/77cca94): Fixes an error where stdin could not be used to create tasks
### Features
1. [6142b7a](https://github.com/influxdata/influx-cli/commit/6142b7a): Support username-password in config update
### Other
1. [e39e365](https://github.com/influxdata/influx-cli/commit/e39e365): Chore: go1.19
1. [188c393](https://github.com/influxdata/influx-cli/commit/188c393): Chore: fix go version in `go.mod`

Doc to update:
https://docs.influxdata.com/influxdb/v2.4/reference/release-notes/influx-cli/
https://docs.influxdata.com/influxdb/v2.4/reference/cli/influx/auth/create/
https://docs.influxdata.com/influxdb/v2.4/security/tokens/create-token/

kelseiv:
@jeffreyssmith2nd to verify--the new feature [6142b7a](https://github.com/influxdata/influx-cli/commit/6142b7a)
means we can remove the workaround in [this hotfix](https://github.com/influxdata/docs-v2/pull/4556/files)?

jeffreyssmith2nd:
That is correct, we should not need that workaround anymore.

jeffreyssmith2nd:
These are the latest notes, we're planning to get the process going today. Thanks!

kelseiv:
thanks, @jeffreyssmith2nd --updated

PR #4610, opened 2022-11-01
OSS 2.5 full docs set

lwandzura:
Closes the following:

https://github.com/influxdata/docs-v2/issues/4583

https://github.com/influxdata/docs-v2/issues/4582


kelseiv (approved):
other than the local rendering issue w release notes on my end--all looks good!

PR #4656, opened 2022-12-01
Flux dynamic type

### LINKED ISSUE #4510: Document Flux dynamic types
**ETA:** The `dynamic` type has been merged into the master branch of Flux, but functions for working with dynamic types are still a WIP.

**PRs:**
- https://github.com/influxdata/flux/pull/5212
- https://github.com/influxdata/flux/pull/5242

---

- [ ] Update the Flux SPEC with the dynamic type
- [x] Update the "Work with data types" section with the new dynamic type
    - This page should cover the purpose of the dynamic and it's primary use case—working with JSON and Flux. 

##### Relevant URLs
- https://docs.influxdata.com/flux/v0.x/spec/types/
- https://docs.influxdata.com/flux/v0.x/data-types/

sanderson:
Closes #4510 

- [x] Rebased/mergeable


content/flux/v0.x/data-types/dynamic.md
```diff
@@ -0,0 +1,622 @@
...
+    value to a regular expression type.
+
+```js
+import "experimental/dynamic"
+import "regexp"
+
+dynamicValue = dynamic.dynamic(v: "^[abc][123]{1,}")
+stringValue = string(v: dynamicValue)
+
+regexp.compile(v: stringValue)
```
onelson, line 291: This is one of the wrinkles of the status quo. I don't know if there's anything to add in the docs since it'll likely be distracting to talk about what _won't work_.

As things are, you can build a `dynamic.dynamic(v: /regex literal/)` but you can never just "unwrap" the already compiled regex from the dynamic.

We're in a similar situation for a few other types iirc. Function objects are one, but I think there are probably others I'm forgetting.

content/flux/v0.x/data-types/dynamic.md
```diff
@@ -0,0 +1,622 @@
...
+
+// Returns dynamic(Example string)
+```
+
+## Reference values in a dynamic type
+
+When working dynamic records and arrays, use bracket and dot notation to reference
+values in each.
+
+##### Reference values in a dynamic record
```
onelson, line 96: With the change made (right at the end) to make attempted member access on a `dynamic(null)` an error, I might mention using `exists` with dynamic objects.

`exists` will "see through" the wrapper to check if there's a null inside or not, allowing us to guard against invalid access.

content/flux/v0.x/data-types/dynamic.md
```diff
@@ -0,0 +1,622 @@
...
+- [Convert dynamic types to Flux types](#convert-dynamic-types-to-flux-types)
+  - [Basic types](#basic-types)
+  - [Composite types](#composite-types)
+- [Include dynamic types in a table](#include-dynamic-types-in-a-table)
+- [Convert a JSON array to a Flux table](#convert-a-json-array-to-a-flux-table)
+- [Encode a dynamic type as JSON](#encode-a-dynamic-type-as-json)
+  - [Encode a dynamic record as JSON](#encode-a-dynamic-record-as-json)
+  - [Encode a dynamic array of different basic types as JSON](#encode-a-dynamic-array-of-different-basic-types-as-json)
+
+### Convert dynamic types to Flux types
```
onelson, line 134: Something that _might be worth mentioning_ is the boundary at the "table" level (ie when trying to use dynamic values to write back to a bucket etc).

We have this requirement that dynamic values are cast to proper basics before they can be used "as column data."

I don't know the best approach, but I'd think saying something about this boundary (which is real and intentional). Using these cast functions is required to cross the threshold, but it may not be clear what the threshold is unless we state it.

If we introduce this concept, it might even be good to mention up top in the overview then elaborate/restate down here.

onelson: :rofl: I now see (moments later) you've got a separate section about building flux tables, which could also be a place to mention _the boundary_.

I still think there might be room for my initial feedback about discussing _the boundary_ explicitly, and in multiple locations in the docs. You never know where a user will enter the conversation.

content/flux/v0.x/data-types/dynamic.md
```diff
@@ -0,0 +1,622 @@
...
+
+### Include dynamic types in a table
+
+**Dynamic types are not supported as column values in Flux tables.**
+To include a dynamic value in a Flux table,
+[cast the dynamic type to a Flux basic type](#basic-types).
+
+The following example uses [`array.from()`](/flux/v0.x/stdlib/array/from/) to 
+build and ad hoc table using dynamic types.
+Each dynamic type is cast to a Flux basic type when defining the row record.
```
onelson, line 433: > Each dynamic type **must be** cast to a Flux basic type when defining the row record.

Weak rephrase suggestion here, just to reinforce the whole idea of _the boundary_ I mentioned previously.

onelson (commented):
This is looking really great :heart_eyes: 

I don't see anything about `dynamic.isType()` which may warrant a mention just to draw attention to the fact it "sees through" the dynamic wrapper, and also knows about a longer list of types than the regular version.

Some other notes attached to the source.

PR #4697, opened 2023-01-10
Fix api version

### LINKED ISSUE #4696: api-docs: Wrong version and missing `info` in API reference
1. The Info section at the top of https://docs.influxdata.com/influxdb/v2.5/api/ contains a link to influxdata/openapi v2.6 instead of v2.5.
2. [v2.6](https://docs.influxdata.com/influxdb/v2.6/api/) is missing the `info` content altogether.

##### Relevant URLs

- https://docs.influxdata.com/influxdb/v2.5/api/
- https://docs.influxdata.com/influxdb/v2.6/api/

jstirnaman:
Closes #4696 

I noticed some incorrect versioning and missing content that is handled by the `info.yml` customizations for OSS API docs.
This PR moves custom content override files from `api-docs/openapi/content/[version]` and into `api-docs/[version]/content` in an attempt to make them harder to miss. So, now the typical workflow would be to copy the old version directory (for example, `api-docs/v2.6`) into the newly released version (for example, `api-docs/v2.7`), and then
update version and repo link in the `api-docs/v2.7/content/info.yml` file. The same process would go for `tag-groups.yml`, although there's nothing in there that requires updating for new releases.

I've also tried to simplify the JavaScript code a little, remove some cruft, and prevent `getswagger.sh` from processing all platforms when it should only be processing a specific one (a side effect of the `redocly/cli lint` exit code).

- [ x] Signed the [InfluxData CLA](https://www.influxdata.com/legal/cla/)
- [x ] Rebased/mergeable


sanderson (approved):
Thank you @jstirnaman!!

PR #4707, opened 2023-01-23
added sql reference

lwandzura:
Please review and let me know:

what is missing
what needs correcting
what should be removed


alamb:
Hi @lwandzura  -- I would be happy to review but I don't seem to be able to find the document you would like me to review and this PR has no way to leave comments (it looks like it is moving a file to a different location on a branch)

lwandzura:
@alamb sorry about that, the file is now available

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,596 @@
...
+
+InfluxDB Cloud backed by InfluxDB IOx supports the following metedata schema queries:
+
+```sql
+SHOW tables
+
+SHOW columns FROM <measurement>
+```
+
+## Functions
```
sanderson, line 462: Do we not have another section specifically for functions? It's going to be a pain to maintain those and this section. I think we can just point users to the other function docs.

sanderson (changes requested):

alamb:
I am happy to review this PR;  it looks like there are outstanding suggestions, so I am not quite sure what to review yet. I'll let you respond to the feedback from @sanderson  and then I will give it a read!

content/influxdb/cloud-iox/sql-reference.md
lwandzura: we need a separate section, I started working on aggregates

I doubt this will be ready by Jan 31, but I can add after

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,596 @@
...
+Literals are the same as constants.  
+
+### String literals
+
+String literals must be surrounded by double quotes. 
+
+```sql
+"santa_monica"
+"pH"
+"average temperature"
```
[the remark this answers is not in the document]
lwandzura: the single vs double quote issue is going to confuse people

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,596 @@
...
+| UNION           | Used to combine the result set of at least two queries. Keeps only unique records.                                                                                                                                               |
+| UNION  ALL      | Like UNION, but keeps all records, including duplicates.                                                                                                                                                                         |
+| WHERE           | Used to filter results based on fields, tags, and/or timestamps.                                                                                                                                                                 |
+| WITH            | Provides the ability to write auxiliary statements for use in a larger query.                                                                                                                                                    |
```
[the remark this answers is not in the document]
lwandzura: I'll just make a list then

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,596 @@
...
+The following query will return an error if the measurement `h2o-pH` and the field `pH` are not double quoted:
+
+```sql
+SELECT "pH", location, time
+FROM "h2o_pH"
+```
+
+## Literals
+
+Literals are the same as constants.
```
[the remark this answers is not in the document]
lwandzura: I really don't think a list here is needed

content/influxdb/cloud-iox/sql-reference.md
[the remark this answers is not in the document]
lwandzura: done

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,614 @@
...
+description: >
+  InfluxDB SQL reference
+menu:
+  influxdb_cloud_iox:
+    name: InfluxDB SQL reference
+    parent: Query data with SQL
+weight: 190
+---
+
+InfluxDB Cloud backed by InfluxDB IOx uses the Apache Arrow DataFusion implementation of SQL.
```
alamb, line 12: I don't know if you have linked it elsewhere but it might help to link  https://arrow.apache.org/datafusion/ here for context

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,614 @@
...
+SELECT "pH" FROM "Water"
+```
+
+{{% note %}}
+**Note:** We recommend always double-quoting identifiers, regardless of case-sensitivity.
+{{% /note %}}
+
+
+Unquoted identifiers **are not** case-sensitive and match any measurement, tag key, or field key with the same characters, despite case.
+For example, if you have two fields in a measurement named `ph` and `pH`, the unquoted identifier, `pH` will match both.
```
alamb, line 60: I actually think the unquoted `pH` will only match `ph` 

I agree with the suggestion to always double quote identifiers

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,614 @@
...
+ALL  
+AS  
+ASC  
+BOTTOM 
+CASE
+DESC 
+DISTINCT                                       
+EXISTS                                                      
+EXPLAIN  
+EXPLAIN ANALYZE
```
alamb, line 169: I think technically double words aren't keywords. So maybe the list should be more like:

```
EXPLAIN
ANALYZE
```

Rather than `EXPLAIN ANALYZE`

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,614 @@
...
+| DATEBIN()             | Bins the input timestamp into a specified interval.                                             |
+| DATE_TRUNC()          | Truncates a timestamp expression based on the date part specified, such as hour, day, or month. |
+| DATE_PART()           | Returns the specified part of a date.                                                           |
+| NOW()                 | Returns the current time.                                                                       |
+|                       
+                 
+#### Examples
+
+```sql
+SELECT time_bucket_gapfill('1 day', time) as day,
```
alamb, line 554: This function is not yet completely implemented -- I am not sure what our policy on documenting "in the works" features is.  @wolffcm  may be able to give you more context

content/influxdb/cloud-iox/sql-reference.md
```diff
@@ -0,0 +1,614 @@
...
+| LOG2()   | Base 2 logarithm                                                                 |
+| POWER()  | Returns the value of a number raised to the power of the number                  |
+| ROUND()  | Round to the nearest integer                                                     |
+| SIGNUM() | Sign of the argument (-1, 0, +1)                                                 |
+| SINE()   | Sine                                                                             |
+| SQRT()   | Returns the square root of a number                                              |
+| TAN()    | Tangent                                                                          |
+| TRUNC()  | Truncates a number to the specified number of decimal places                     |
+
+### Conditional functions
```
alamb, line 600: I believe `COALESCE` and `NULLIF` were already covered above (with CASE)

alamb (approved):
This is looking quite nice @lwandzura  🏆

content/influxdb/cloud-iox/sql-reference.md
wolffcm: I have begun implementing a function called `DATE_BIN_GAPFILL` that takes the same arguments as `DATE_BIN` but fills gaps. At the moment it will return a `not implemented` error if you try to use it.

We will also be providing a `TIME_BUCKET_GAPFILL` function that is an alias for `DATE_BIN_GAPFILL`, for users who migrate from the other platforms. So, long story short, neither of those functions are ready for users yet.

As far as documenting in-the-works-features, we tended not to for Flux, to avoid people being confused/disappointed if they try something that was documented but did not work.

lwandzura: We generally don't add features or things "in the works" because it does cause confusion.  Do you think this will most likely be implemented after the Jan 31 release?  If so I will remove it for now.  I can alway add it back once it's ready, that's easy to do.

wolffcm: Yes, it will be implemented after Jan 31. Best to remove it for now.

content/influxdb/cloud-iox/sql-reference.md
lwandzura: added link

content/influxdb/cloud-iox/sql-reference.md
lwandzura: fixed

content/influxdb/cloud-iox/sql-reference.md
lwandzura: removed for now

PR #4709, opened 2023-01-24
feat: add new config value log-timedout-queries

jeffreyssmith2nd:
This a new config value that will be added to the enterprise 1.11 line.


sanderson (approved):
Thanks @jeffreyssmith2nd!

PR #4712, opened 2023-01-24
Fix Authorization header name in text

sulate:
Text has `Authentication` as the header name but the correct name is `Authorization` (all examples are correct).


sanderson (approved):
Thanks @sulate!

Issue #4732, opened 2023-01-31
Telegraf v1.20, v1.21, and v1.22 install instructions do not work


powersj:
As reported by @gm0neyl0ve in https://github.com/influxdata/telegraf/issues/12563, it appears that versions v1.20, v1.21, and v1.22 install instructions currently do not work with Telegraf.

* v1.19 and older use `apt-key add` to install the GPG and these instructions work 
* v1.23 and newer `dearmor` the GPG key and add the `signed-by` to the `/etc/apt/sources.list.d/influxdata.list` entry and these instructions work
 
##### Relevant URLs

- https://docs.influxdata.com/telegraf/v1.20/introduction/installation/
- https://docs.influxdata.com/telegraf/v1.21/introduction/installation/
- https://docs.influxdata.com/telegraf/v1.22/install/

PR #4733, opened 2023-01-31
fix(telegraf): update v1.20, v1.21, v1.22 install instructions

powersj:
This change takes the install instructions from v1.24 which correctly install the GPG key during the install for curl and wget examples.

fixes: #4732

- [x] Rebased/mergeable


sanderson (approved):
Thanks @powersj!

PR #4771, opened 2023-02-27
Use SQL and Flux together

sanderson:
- [x] Rebased/mergeable


content/influxdb/cloud-iox/query-data/flux-sql.md
```diff
@@ -0,0 +1,522 @@
...
+[`sort()`](/flux/v0.x/stdlib/universe/sort/) to re-sort your data time **after**
+performing other transformations._
+
+### Rename the `time` column to `_time`
+
+Many Flux functions expect or require a column named `_time` (with a leading underscore).
+The IOx storage engine stores each point's timestamp in the `time` column (no leading underscore).
+Depending on which Flux functions you use, you may need to rename the `time`
+column to `_time`.
+YOu can do this in your SQL query with an `AS` clause or in Flux with the
```
Jayclifford345, line 306: Y0u -> You

I am wondering if we reiterate here that for the best performance, they should use SQL `AS` rather than rename()

content/influxdb/cloud-iox/query-data/flux-sql.md
```diff
@@ -0,0 +1,522 @@
...
+  hum,
+  co
+FROM home
+WHERE
+  time >= '2022-01-01T08:00:00Z'
+  AND time <= '2022-01-01T20:00:00Z' 
+"
+
+iox.sql(bucket: "get-started", query: query)
+    |> group(columns: ["room"])
```
Jayclifford345, line 410: Is it worth explaining again why you group by room here when you unpivot?

content/influxdb/cloud-iox/query-data/flux-sql.md
```diff
@@ -0,0 +1,522 @@
...
+and the flexibility of Flux.
+
+## What to do in SQL versus Flux?
+
+We recommend doing as much of your query as possible in SQL for the most
+performant queries.
+Do any further processing in Flux.
+The following chain of Flux functions can be replicated in SQL:
+
+```js
```
Jayclifford345, line 63: I tried to turn the bones of a SQL query into this format so a user can see the bare bones comparison of each before the examples but can't quite get it to work:
```sql
SELECT
  DATE_BIN() AS _time,
  avg() AS ###,
FROM <>
WHERE
  time >= ###
  AND time < ###
GROUP BY _time
ORDER BY _time
```
Your examples build the aggregate to probs isn't needed

content/influxdb/cloud-iox/query-data/flux-sql.md
```diff
@@ -0,0 +1,522 @@
...
+build a **Flux query** that uses the [`iox.sql()` function](/flux/v0.x/stdlib/experimental/iox/sql/)
+to execute a SQL function.
+The SQL query should return the base data set for your query.
+If this data needs further processing that can't be done in SQL, those operations
+can be done with native Flux.
+
+{{% note %}}
+The process below uses the `/api/v2/query` endpoint and can be used to execute
+SQL queries against an InfluxDB IOx-powered bucket with all existing
+InfluxDB 2.x clients.
```
Jayclifford345, line 139: We could mention here infuxdb data source for grafana cloud so those users don't feel like they can't currently integrate (since the flight-sql plugin is install only)

Also worth mentioning users could just use a raw rest request here as you mentioned with `/api/v2/query`

Jayclifford345 (commented):
Such a great doc! I think we could get a blog out of this aswell to elevate some flux concerns.  I just added a few comments but more questions than suggestions

content/influxdb/cloud-iox/query-data/flux-sql.md
sanderson: I want to keep the example below just to show an actual query. The examples are "collapsed" by default, so users have to opt into viewing them. I did add something similar to your proposal along with the chain of generic Flux functions.

content/influxdb/cloud-iox/query-data/flux-sql.md
sanderson: Good call. I added the following list:

- InfluxDB 2.x client libraries
- Grafana and Grafana Cloud InfluxDB data source
- Flux VS code extensions
- InfluxDB OSS 2.x dashboards

content/influxdb/cloud-iox/query-data/flux-sql.md
sanderson: Added a recommended note after "an `AS` clause".

content/influxdb/cloud-iox/query-data/flux-sql.md
sanderson: Added this to the section on unpivoting.

PR #4773, opened 2023-03-02
Added SQL string function documentation and examples

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud-iox/reference/sql/functions/string.md
+++ b/content/influxdb/cloud-iox/reference/sql/functions/string.md
@@ -1,0 +1,1114 @@
+---
+title: SQL string functions
+list_title: String functions
+description: >
+  Use string functions to operate on string values in SQL queries.
+menu:
+  influxdb_cloud_iox:
+    name: String
+    parent: sql-functions    
+weight: 307
+---
+
+The InfluxDB SQL implementation supports the following string functions for
+operating on string values:
+
+
+- [ascii](#ascii)
+- [bit_length](#bit_length)
+- [btrim](#btrim)
+- [char_length](#char_length)
+- [character_length](#character_length)
+- [concat](#concat)
+- [concat_ws](#concat_ws)
+- [chr](#chr)
+- [initcap](#initcap)
+- [left](#left)
+- [length](#length)
+- [lower](#lower)
+- [lpad](#lpad)
+- [ltrim](#ltrim)
+- [md5](#md5)
+- [octet_length](#octet_length)
+- [repeat](#repeat)
+- [replace](#replace)
+- [reverse](#reverse)
+- [right](#right)
+- [rpad](#rpad)
+- [rtrim](#rtrim)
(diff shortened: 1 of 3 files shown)


sanderson:
@alamb I noticed some behavior with some of the string functions that may be a bug. If I need to modify a tag value, I have to explicitly cast the tag column to a string. I know that tags and fields use different types (tags seem to be a key-value pair), but in the context of InfuxDB, tags are always strings, so casting shouldn't be necessary. For example:

```sql
SELECT DISTINCT
  upper(room::STRING)
FROM home
```

I feel like I should just be able to do the following:

```sql
SELECT DISTINCT
  upper(room)
FROM home
```

Is this something worth filing an issue about? I think this is fairly unique to our implementation of SQL, so if I do create an issue, where should I create it?

alamb:
Hi @sanderson  -- I can reproduce the error you are describing here:

```
26f7e5a4b7be365b_917b97a92e883afc> select distinct host from cpu;
+---------------------+
| host                |
+---------------------+
| MacBook-Pro-8.local |
+---------------------+
```

But this fails:

```
26f7e5a4b7be365b_917b97a92e883afc> select distinct upper(host) from cpu;
Error running remote query: Tonic(Status { code: Internal, message: "Error while planning query: Internal error: The \"upper\" function can only accept strings.. This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker", metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Thu, 02 Mar 2023 21:27:29 GMT", "content-length": "0"} }, source: None })
26f7e5a4b7be365b_917b97a92e883afc>
```

I think this is a bug/limitation in DataFusion that should be fixed upstream. I will do so

content/influxdb/cloud-iox/reference/sql/functions/string.md
```diff
@@ -0,0 +1,1114 @@
+---
```
alamb, line 1: This looks like great content. Would you be amenable to figuring out how we could move the content into the DataFusion repo (and maybe have a script to sync it over here or something)? That way we could get more reviewers in the community to help review

sanderson: My plan is to port all of this content over to the DataFusion docs once it's complete. That will likely be sometime next week. Pulling content from upstream will be a bit of a challenge since not all functionality upstream is applicable downstream in InfluxDB, but we may be able to figure something out.

alamb:
> I think this is a bug/limitation in DataFusion that should be fixed upstream. I will file a bug.

https://github.com/apache/arrow-datafusion/issues/5471

content/influxdb/cloud-iox/reference/sql/functions/string.md
alamb: Filed https://github.com/influxdata/idpe/issues/17251 to track helping this effort

PR #4774, opened 2023-03-02
SQL regular expression functions

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud-iox/reference/sql/functions/regular-expression.md
+++ b/content/influxdb/cloud-iox/reference/sql/functions/regular-expression.md
@@ -1,0 +1,74 @@
+---
+title: SQL regular expression functions
+list_title: Regular expression functions
+description: >
+  Use regular expression functions to operate on data in SQL queries.
+menu:
+  influxdb_cloud_iox:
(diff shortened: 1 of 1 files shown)


sanderson:
@alamb An issue I came across when documenting the regular expression functions is that `regexp_match` returns a list/array of substring matches, but our SQL implementation doesn't support lists. I tried to index into the list and the query doesn't error, but the function returns a null value:

```sql
SELECT DISTINCT
  room,
  regexp_match(room::STRING, '\sRoom')[0] AS regexp_match
FROM home
```

Returns:

| room        | regexp_match |
| :---------- | :----------- |
| Living Room |              |
| Kitchen     |              |

I just left that function commented out for now.

alamb:
> An issue I came across when documenting the regular expression functions is that regexp_match returns a list/array of substring matches, but our SQL implementation doesn't support lists. I tried to index into the list and the query doesn't error, but the function returns a null value:

Thanks for the report @sanderson  -- I agree it is a bug and I filed https://github.com/apache/arrow-datafusion/issues/5479 upstream to track it

PR #4775, opened 2023-03-03
SQL conditional functions

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud-iox/reference/sql/functions/conditional.md
+++ b/content/influxdb/cloud-iox/reference/sql/functions/conditional.md
@@ -1,0 +1,97 @@
+---
+title: SQL conditional functions
+list_title: Conditional functions
+description: >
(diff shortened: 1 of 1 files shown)


Jayclifford345:
So this one looks great but I am struggling to imagine using it in practice. Is there a good scenario to bring this together? Like would it make sense to use this to handle missing tag values? 

For instance, is this a good way to assume missing tag values? Say location ?

sanderson:
@Jayclifford345 `coalesce` is essentially a fill functions. So you can use it to provide a default value for a field if a field is null.

For example, the following query selects the `field1` field and fills any null values with `0`:

```sql
SELECT coalesce(field1, 0) FROM measurement
```

`nullif`... I'm not really sure of a valid use case, but it's a common SQL function, so I know use cases exist 🤷🏻‍♂️.

PR #4777, opened 2023-03-06
docs: explain a few important fields on replication ls

jeffreyssmith2nd:
Hoping to clear up some confusion around the `replication ls` results by adding this new field and the documentation to go with it.

### DIFF
--- a/content/influxdb/v2.7/reference/cli/influx/replication/list.md
+++ b/content/influxdb/v2.7/reference/cli/influx/replication/list.md
@@ -13,15 +13,22 @@ influxdb/v2.7/tags: [write, replication]
 Replication remotes and replication streams can only be configured for InfluxDB OSS.
 {{% /cloud %}}
 
-
 The `influx replication list` command lists all InfluxDB replication streams and their corresponding metrics.
 
+When listing replications, there are several fields that provide information about how the replication queue is performing:
+
+- `Latest Status Code` indicates the status code of the last `write` request to the remote. This should be a 204 during healthy operation.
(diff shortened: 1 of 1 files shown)


content/influxdb/v2.7/reference/cli/influx/replication/list.md
```diff
@@ -13,15 +13,22 @@ influxdb/v2.7/tags: [write, replication]
...
 {{% /cloud %}}
 
-
 The `influx replication list` command lists all InfluxDB replication streams and their corresponding metrics.
 
+When listing replications, there are several fields that provide information about how the replication queue is performing:
+
+- `Latest Status Code` indicates the status code of the last `write` request to the remote. This should be a 204 during healthy operation.
+- `Remaining Bytes to be Synced` is the number of bytes that have not been synced to the remote. This number should stay relatively low/close to 0, and should not be constantly increasing.
+- `Current Queue Bytes on Disk` is the total size of the replication queue. The replication queue is cleaned up every `--max-age` seconds. If your queue is filling up your disk, lower this value at the cost of potentially lower reliability.
```
sanderson, line 22:
```suggestion
The following metrics are listed for each replication and provide information about how the replication queue is performing:

- **Latest Status Code**: Status code of the last `write` request to the remote.
  This should be a 204 during healthy operation.
- **Remaining Bytes to be Synced**: Number of bytes that have not been synced to the remote.
  This number should stay close to 0 and should not be constantly increasing.
- **Current Queue Bytes on Disk**: Total size of the replication queue in bytes.
  The replication queue is cleaned up every `--max-age` seconds.
  If your queue is filling up your disk, lower this value at the cost of potentially lower reliability.
```

sanderson (approved):
Thanks @jeffreyssmith2nd Just some suggestions to the content and it's gtg 👍🏻

PR #4780, opened 2023-03-07
Document the SQL table value constructor

sanderson:
This documents the SQL table value constructor (TVC) for building ad hoc tables at query time. Although TVC should work with both `FROM` and `JOIN` clauses, there is a bug that prevents it from working with `JOIN` (see https://github.com/influxdata/influxdb_iox/issues/7152). For now, the `JOIN`-related content is commented out.

- [x] Rebased/mergeable


content/influxdb/cloud-iox/reference/sql/table-value-constructor.md
```diff
@@ -0,0 +1,121 @@
...
+to build an ad hoc table at query time.
+
+```sql
+VALUES (row_value_list)[,...n]
+```
+
+##### Arguments
+
+- **row_value_list**:
+  Comma separate list of column values.
```
jstirnaman, line 24:
```suggestion
  Comma-delimited list of column values.
```

jstirnaman (changes requested):

jstirnaman (approved):
🎸

PR #4811, opened 2023-03-17
fix(telegraf): remove PSI input plugin

powersj:
This is an external plugin, not something that comes with telegraf.

- [x] Rebased/mergeable

### DIFF
--- a/data/telegraf_plugins.yml
+++ b/data/telegraf_plugins.yml
@@ -1647,14 +1647,6 @@ input:
     introduced: 1.16.0
     tags: [linux, macos, windows]
 
-  - name: PSI
-    id: psi
-    description: |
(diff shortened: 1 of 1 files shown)


sanderson:
@powersj We have been including external plugins in the list of available plugins with special metadata, `external: true`. Should we not include external plugins in the list of plugins?

sanderson:
Ok, it sounds like we need to change the way external plugins are listed in the plugin list AND improve the external plugin documentation.

I've never used an external plugin, so walk me through how to. Do use the `execd` plugin? How to you load and run a specific external plugin? What does the config look like?

powersj:
so I think I learned that all the external plugins are listed here, so removing just one is not the right thing. I am going to close this for now.

PR #4816, opened 2023-03-20
Feat: fixed sidebar to the page for easy navigation

### LINKED ISSUE #4815: Feat: fix sidebar to the page for easy navigation
Hello there! I was reading through the InfluxData Docs and something caught my eyes. I noticed that the current the sidebar isn't fixed to the page. From my experience and from most documentation sites I've found on the web, a best practice is to have the sidebar fixed to page i.e it shouldn't scroll along with the reader as they scroll across the page.

With the current sidebar, when a reader scrolls to the bottom of a very long page and wants to navigate to another page linked on the sidebar. It will have them scrolling all the way to the top, trying to find the page on the sidebar. That might not be ideal in every situation.

I propose that the sidebar gets fixed to the page to address the issues above.

##### Relevant URLs
- _Provide relevant URLs_

victor-cody:
Closes #4815

My changes will fix the sidebar to page and with this, the sidebar won't be moved when a reader scrolls through a page, the reader will be able to know where the are in the InfluxData Docs at all times and they will be able to easily navigate to another page from the sidebar without losing where they are on the current page.

- [x] Rebased/mergeable

### DIFF
--- a/assets/styles/layouts/_sidebar.scss
+++ b/assets/styles/layouts/_sidebar.scss
@@ -1,9 +1,16 @@
 .sidebar {
   display: block;
-  position: relative;
   flex-grow: 1;
-  padding: 0 1em;
+  padding: 0 1em 0.8rem;
   width: 25%;
+
+  height: 100%;
+  max-height: 100vh;
+  position: sticky;
+  top: 0;
+  overflow-y: scroll;
+  scrollbar-width: thin;
+  scroll-behavior: smooth;
 
   &--search {
     position: relative;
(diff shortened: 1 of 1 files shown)


sanderson:
Thanks for this @victor-cody. We do appreciate the work you did to make this change, but the choice to have a fixed left nav was a conscious one. In the early feedback we got as we redesigned the docs (a few years ago), users preferred a left nav that scrolled with the page rather than being fixed to it. It's certainly something we can revisit, but I'm going to close this PR for now.

PR #4835, opened 2023-03-28
Provide links to alternate pages in IOx and TSM docs

sanderson:
This PR extends the functionality of the product dropdown menu and the IOx wayfinding modal to link users to an appropriate alternative page in the other engine's documentation. For the product dropdown, this will also work for linking to alternative OSS docs when in the IOx documentation.

The `alt_engine` frontmatter is used to point to the correct alternative page in the other engine documentation. If the page exists at the same path, the `alt_engine` frontmatter isn't needed.

- [x] Rebased/mergeable


jstirnaman (approved):
Good idea! LGTM

PR #4935, opened 2023-05-11
adjust windows install guide

l3nnartt:
On windows it wont work with just telegraf.exe & it provides an error if you run the test command.
image
image

- [✅] Signed the [InfluxData CLA](https://www.influxdata.com/legal/cla/)
- [✅] Rebased/mergeable


Dustin284 (approved):
Looks good

content/telegraf/v1.26/install.md
```diff
@@ -318,11 +318,12 @@ In PowerShell _as an administrator_, do the following:
     ```powershell
     > C:\"Program Files"\InfluxData\telegraf\telegraf.exe --config C:\"Program Files"\InfluxData\telegraf\telegraf.conf --test
     ```
+    > On Windows you need to disable inputs.processes in the `telegraf.conf`.
```
jstirnaman, line 321: @powersj Is it necessary to disable `inputs.processes` for Windows? Or is a different configuration needed?

powersj: I would not say it is required. While the processes plugin does not support Windows the user will get a warning message which says "Current platform is not supported". No errors or issues would prevent the user from using the config otherwise.

jstirnaman: @powersj 
> I would not say it is required. While the processes plugin does not support Windows the user will get a warning message which says "Current platform is not supported". No errors or issues would prevent the user from using the config otherwise.

Finally testing this on Windows. I do get the error below when running with `--test`. Should this be a warning instead?
While `--service start` doesn't warn or throw anything, and I see the process running, `--debug` also doesn't output anything.

```powershell
PS C:\Program Files\InfluxData\telegraf> .\telegraf.exe `
>> --config "C:\Program Files\InfluxData\telegraf\telegraf.conf" --test
2023-08-14T20:53:06Z I! Loading config: C:\Program Files\InfluxData\telegraf\telegraf.conf
2023-08-14T20:53:06Z E! error loading config file C:\Program Files\InfluxData\telegraf\telegraf.conf: plugin inputs.processes: line 3828: configuration specified the fields ["use_sudo"], but they weren't used
```

powersj: That is a bug and we should comment out that option as it is the default anyway.

jstirnaman:
Resolved with the workaround in #5086 
@l3nnartt Thanks for catching this. I expect it to be fixed upstream in the Telegraf sample config.

l3nnartt:
thanks for testing & fixing!

PR #4957, opened 2023-05-25
InfluxQL ORDER BY, LIMIT, SLIMIT, OFFSET, and SOFFSET clauses

sanderson:
Adds documentation for the following InfluxQL clauses:

- ORDER BY
- LIMIT
- SLIMIT
- OFFSET
- SOFFSET

---

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud-dedicated/reference/influxql/offset-and-soffset.md
+++ b/content/influxdb/cloud-dedicated/reference/influxql/offset-and-soffset.md
@@ -50,8 +50,8 @@ SELECT_clause FROM_clause [WHERE_clause] [GROUP_BY_clause] [ORDER_BY_clause] [LI
 
 ### Notable OFFSET clause behaviors
 
-- If there is no `LIMIT` clause in a query with an `OFFSET` clause, InfluxQL
-  assumes the behavior, `LIMIT 1`.
+- If there is no `LIMIT` clause in a query with an `OFFSET` clause, the query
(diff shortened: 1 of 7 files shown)


content/influxdb/cloud-dedicated/reference/influxql/offset-and-soffset.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+#### Arguments 
+
+- **N**: Number of rows to skip in each InfluxQL group before returning results. 
+
+### Notable OFFSET clause behaviors
+
+- If there is no `LIMIT` clause in a query with an `OFFSET` clause, InfluxQL
+  assumes the behavior, `LIMIT 1`.
```
stuartcarnie, line 54: This is not accurate – it assumes no `LIMIT` clause, which I verified with OSS and also how it is implemented in IOx.

stuartcarnie (changes requested):
Great stuff, @sanderson! There is one amendment about `OFFSET` and `LIMIT` behaviour

content/influxdb/cloud-dedicated/reference/influxql/offset-and-soffset.md
sanderson: Interesting. In my testing, if I included an offset without a limit, it would return a single row per group at the correct offset. Is that the expected behavior?

PR #4966, opened 2023-06-01
Code placeholders

sanderson:
- Adds the ability to custom code placeholders inline.
- All placeholders with the same name/key are updated when one is updated
- Custom values do not persist through a refresh. This was done for security reasons as these values are potentially sensitive.

Kapture 2023-06-01 at 09 08 20

---

- [x] Rebased/mergeable


jstirnaman (approved):
LGTM! 🧨

PR #4990, opened 2023-06-22
Downsampling guides for Serverless and Dedicated

sanderson:
Adds downsampling guides for InfluxDB Cloud Serverless and Dedicated

_Only includes serverless for now, but once reviewed, will port the instructions to dedicated._

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud-serverless/process-data/downsample.md
+++ b/content/influxdb/cloud-serverless/process-data/downsample.md
@@ -15,8 +15,10 @@ Query and downsample time series data stored in InfluxDB and write the
 downsampled data back to InfluxDB.
 
 This guide uses [Python](https://www.python.org/) and the
-[InfluxDB v3 Python client library](https://github.com/InfluxCommunity/influxdb3-python).
-and assumes you have already
+[InfluxDB v3 Python client library](https://github.com/InfluxCommunity/influxdb3-python),
+but you can use your runtime of choice and any of the available
+[InfluxDB v3 client libraries](/influxdb/cloud-serverless/reference/client-libraries/v3/).
+This guide also assumes you have already
 [setup your Python project and virtual environment](/influxdb/cloud-serverless/query-data/sql/execute-queries/python/#create-a-python-virtual-environment).
 
 - [Install dependencies](#install-dependencies)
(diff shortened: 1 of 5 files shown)


content/influxdb/cloud-serverless/process-data/downsample.md
```diff
@@ -0,0 +1,354 @@
...
+- `influxdb_client_3`
+- `pandas`
+
+```sh
+pip install influxdb3-python pandas
+```
+
+## Prepare InfluxDB buckets
+
+The downsampling process involves two InfluxDB buckets:
```
Jayclifford345, line 44: I wonder if we explain here why we use two buckets rather than 2 measurements. So they understand the benefit of retention policies.

Jayclifford345 (commented):
Just the one commend the only other question I had was on this SQL query:
```sql
SELECT
  DATE_BIN(INTERVAL '1 hour', time) AS time,
  room,
  AVG(temp) AS temp,
  AVG(hum) AS hum,
  AVG(co) AS co
FROM home
WHERE time >= now() - INTERVAL '24 hours'
GROUP BY DATE_BIN(INTERVAL '1 hour', time), room
ORDER BY time
```
Is the `GROUP BY DATE_BIN(INTERVAL '1 hour', time)` now the best way to perform an aggregation? 
```
SELECT
  DATE_BIN(INTERVAL '1 hour', time) AS time,
  room,
  AVG(temp) AS temp,
  AVG(hum) AS hum,
  AVG(co) AS co
FROM home
WHERE time >= now() - INTERVAL '24 hours'
GROUP BY 1, room
ORDER BY time
```
is that considered bad practice now? Just for my own sanity :)

content/influxdb/cloud-serverless/process-data/downsample.md
sanderson: Good call. I'll add it.

sanderson:
> Is the GROUP BY DATE_BIN(INTERVAL '1 hour', time) now the best way to perform an aggregation?
>
> ```
> SELECT
>   DATE_BIN(INTERVAL '1 hour', time) AS time,
>   room,
>   AVG(temp) AS temp,
>   AVG(hum) AS hum,
>   AVG(co) AS co
> FROM home
> WHERE time >= now() - INTERVAL '24 hours'
> GROUP BY 1, room
> ORDER BY time
> ```
>
> is that considered bad practice now? Just for my own sanity :)

Nope, definitely not bad practice. Much more elegant actually. I just forgot that you can index-reference queried columns. I'm going to update to use your version.

PR #5038, opened 2023-07-19
Correct influx cli flag - it's --errors-file rather than --error-file.

btasker:
It also currently seems to need the file to already exist, but I'm going to raise a PR against the CLI shortly to fix that

- [x ] Signed the [InfluxData CLA](https://www.influxdata.com/legal/cla/)
- [x ] Rebased/mergeable


sanderson (approved):
Thanks @btasker!

PR #5046, opened 2023-07-25
chore(v3): Use python client library

### LINKED ISSUE #5025: Recommend Flight over FlightSQL
Recommend using Flight instead of FlightSQL.

jstirnaman:
- Replaces Flight SQL examples with Client Library examples.
- Closes #5025.


sanderson (approved):
Looks good 👍🏻

PR #5102, opened 2023-08-24
influxctl authentication note

reidkaufmann:
This just adds a helpful pointer to eliminate an annoying and unnecessary warning message that may trip up parsing JSON output from the utility.  The note is Linux specific.  It may be Ubuntu specific as well, but this is one of the most popular distros.

Reference: https://stackoverflow.com/questions/75406844/not-loading-module-atk-bridge-the-functionality-is-provided-by-gtk-natively

- [x] Rebased/mergeable


jdstrand (changes requested):
IMHO it would be better to adjust influxctl to not cause the output rather than telling the user to modify their system. The software involved is for assistive technologies and advising to remove it could break a system for someone who requires the assistive technologies.

My guess is this seems like a system that went through several upgrades has some old packages installed and is not influxctl specific. I'd also guess that this is happening due to how influxctl is launching the browser-- the thing that influxctl is calling is likely what is causing this output, not influxctl itself. It could also be the browser itself that is outputting this.

I suggest one of:
* figure out if there is a better way to launch the browser (which suppressing the output)
* have `influxctl` remove `gail:atk-bridge` from GTK_MODULES right before it invokes the browser. This would be a quick fix and likely easy to do, but in the event the user needs the module to use their browser, this could regress functionality but doing this would allow you to suppress the output and defer looking into it more until we receive a bug report
* redirect stderr from the browser launcher or browser and ignore it (or filter out this specific message; feel free to undo the redirect after launching the browser). This is probably the best solution

content/influxdb/cloud-dedicated/reference/cli/influxctl/_index.md
```diff
@@ -343,3 +343,11 @@ credentials you should have created when setting up your InfluxDB Cloud
...
 Auth0 issues a short-lived (1 hour) token that authenticates access to your
 InfluxDB Cloud Dedicated cluster.
+
+
+When authenticating on Linux, if you see a warning on `stderr` like
+`Not loading module "atk-bridge": The functionality is provided by GTK natively. Please try to not load it.`,
+this can be prevented by either:
+- Unsetting the `GTK_MODULES` environment variable
+  (or at least removing `gail:atk-bridge` from it's value)
+- Removing the `libatk-adaptor` package and rebooting
```
jstirnaman, line 353:
```suggestion
## Troubleshoot

- **Not loading module "atk-bridge"**: When authenticating, some Linux systems might report the following warning in the terminal (stderr):

  ```sh
  Not loading module "atk-bridge": The functionality is provided by GTK natively. Please try to not load it.
  ```

  To silence the warning, make one of the following changes to your system:
  
  - Unset the `GTK_MODULES` environment variable or remove `gail:atk-bridge` from it's value.
  - Remove the `libatk-adaptor` package and reboot.
```

jstirnaman (changes requested):

jdstrand:
> While I like a technical solution, I went this route because I doubted working around an OS issue likely be fixed externally is the best prioritization of our limited resources. I would not risk regressing functionality in older distros to fix a cosmetic annoyance. Finding a better way is fine with me but that implies some research/time. Any objection to documenting this and then removing this if and when we have a technical solution later?
> 
> My intent is just to help the customer and I don't think the approaches need to be exclusive.

I don't think that recommending turning off accessibility system-wide is the way to go as this could affect how they use their computer outside of `influxctl`. I don't think `influxctl` redirecting stderr would be a big lift, but if you are going to document a workaround, I might suggest:
"
... (your other text) 

To silence the warning, when invoking `influxctl` unset the `GTK_MODULES` environment variable (or remove `gail:atk-bridge` from it's value). Eg:
```
$ GTK_MODULES= influxctl ...
```

Users can then choose to create a shell alias, updated GTK_MODULES system-wide, etc if they want to make this permanent, but the info we provide won't break their system.

content/influxdb/cloud-dedicated/reference/cli/influxctl/_index.md
```diff
@@ -343,3 +343,17 @@ credentials you should have created when setting up your InfluxDB Cloud
...
+
+- **Not loading module "atk-bridge"**: When authenticating, some Linux systems might report the following warning in the terminal (stderr):
+
+  ```sh
+  Not loading module "atk-bridge": The functionality is provided by GTK natively. Please try to not load it.
+  ```
+
+  To silence the warning, make one of the following changes to your system:
+  
+  - Unset the `GTK_MODULES` environment variable or remove `gail:atk-bridge` from it's value.
```
jstirnaman, line 358:
```suggestion
  - Unset the `GTK_MODULES` environment variable or remove `gail:atk-bridge` from its value.
```

content/influxdb/cloud-dedicated/reference/cli/influxctl/_index.md
```diff
@@ -343,3 +343,18 @@ credentials you should have created when setting up your InfluxDB Cloud
 Dedicated cluster.
 Auth0 issues a short-lived (1 hour) token that authenticates access to your
 InfluxDB Cloud Dedicated cluster.
+
+
```
jstirnaman, line 347:
```suggestion
```

jstirnaman (approved):

PR #5103, opened 2023-08-24
chore: explain null tags in primary keys

jacobmarble:

Clarify that primary keys depend on tags, but tag values can be null.

- [x] Rebased/mergeable


sanderson (approved):
Thanks @jacobmarble

PR #5127, opened 2023-09-06
hotfix(v3): missing shared example files

jstirnaman:
- replace paths to missing shared-text files with examples.
- update timestamps for custom-timestamps in older examples.
- the shared-text solution isn't really working for us, as it can't render the host shortcode and the placeholder names differ between platforms. I'll probably start doing away with it and tinker on a better way to achieve what I was trying to do.

Closes #

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/guides/api-compatibility/v1/_index.md
+++ b/content/influxdb/clustered/guides/api-compatibility/v1/_index.md
@@ -171,7 +171,7 @@ Use `Bearer` to authenticate a write request:
 {{% influxdb/custom-timestamps %}}
 {{% code-placeholders "DATABASE_NAME|DATABASE_TOKEN" %}}
 ```sh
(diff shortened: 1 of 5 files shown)


sanderson (approved):
I see why this needed to change, but I also see the value in being able to reuse code blocks. Do you think it would be worth updating the `shared` shortcodes to be able to handle an embedded `influxdb/host` shortcode? It would involve a little regexp trickery, but I've done it before.

jstirnaman:
> I see why this needed to change, but I also see the value in being able to reuse code blocks. Do you think it would be worth updating the `shared` shortcodes to be able to handle an embedded `influxdb/host` shortcode? It would involve a little regexp trickery, but I've done it before.

I think I could go either way. If we can use some regexp that allows for common placeholders in the examples, and DRY things up, then ok. My main goal with those was to make the examples testable, but I obviously didn't finish that. Less duplicate code and margin for error is also a big win.

PR #5141, opened 2023-09-15
fix(v3): python client synchronous and batch writing, remove async

jstirnaman:
- Synchronous is the default. Async isn't supported for now (so that write matches the synchronous behavior of the query endpoint.
- Remove confusion around batch; batch is also synchronous and it's the only way to use callbacks.

- [ ] Rebased/mergeable


sanderson (approved):
👌🏻 🚀

PR #5184, opened 2023-10-18
Add Homebrew install instructions for influxctl

sanderson:
Adds `influxctl` install instructions using Homebrew to Dedicated and Clustered.

- [x] Rebased/mergeable


powersj (approved):
Thank you! And a big than you for watching the influxctl repo :)

PR #5193, opened 2023-10-25
Update database naming restrictions for Dedicated and Clustered

### LINKED ISSUE #5167: Correct Database Naming Guidelines
The documentation on valid DB names https://docs.influxdata.com/influxdb/cloud-dedicated/admin/databases/create/#database-naming-restrictions is inaccurate on all counts.  It says names:

- Must contain two or more characters
- Cannot start with an underscore (_)
- Cannot contain whitespace characters, double quotes ("), or percent signs (%)

Because

- I can create a single character name
- I can create a name with a leading underscore 
- Most special characters are disallowed.  The real criteria is that we allow anything Rust considers an alphanumeric plus underscore, dash, and forward slash.  These characters are allowed: `abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789ª²³µ¹º¼½¾ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõöøùúûüýþÿ_-/`

This is true for `dedicated` and `clustered`.

###
Suggested fixes (take or leave)

- Document maximum length (64 characters) and strike the minimum or couch it as a "should be longer than 2".
- Do not address/change the leading underscore restriction for now -- it is good advice since it may become restricted.  Maybe say "should not".
- Rewrite the restricted/allowed characters bullet... something like "Cannot contain whitespace, punctuation, or special characters.  Only alphanumeric characters, underscore (`_`), dash, (`-`), and forward-slash (`/`) are permitted: `abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789ª²³µ¹º¼½¾ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõöøùúûüýþÿ_-/`"

sanderson:
Closes #5167

Updates the database naming restrictions for InfluxDB Cloud Dedicated and InfluxDB Clustered to match the findings outline in #5167.

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/admin/databases/create.md
+++ b/content/influxdb/clustered/admin/databases/create.md
@@ -89,9 +89,11 @@ The retention period value cannot be negative or contain whitespace.
 
 Database names must adhere to the following naming restrictions:
 
-- Must contain two or more characters
-- Cannot start with an underscore (`_`)
-- Cannot contain whitespace characters, double quotes (`"`), or percent signs (`%`)
+- Cannot contain whitespace, punctuation, or special characters.
+  Only alphanumeric, underscore (`_`), dash (`-`), and forward-slash
+  (`/`) characters are permitted.
+- Should not start with an underscore (`_`)
+- Maximum length of 64 characters.
 
 ## InfluxQL DBRP naming convention
 
--- a/content/influxdb/clustered/admin/databases/create.md
+++ b/content/influxdb/clustered/admin/databases/create.md
@@ -92,7 +92,7 @@ Database names must adhere to the following naming restrictions:
 - Cannot contain whitespace, punctuation, or special characters.
   Only alphanumeric, underscore (`_`), dash (`-`), and forward-slash
   (`/`) characters are permitted.
-- Should not start with an underscore (`_`)
+- Should not start with an underscore (`_`).
 - Maximum length of 64 characters.
 
 ## InfluxQL DBRP naming convention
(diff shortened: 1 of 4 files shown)


content/influxdb/clustered/admin/databases/update.md
```diff
@@ -93,9 +93,11 @@ The retention period value cannot be negative or contain whitespace.
 
 Database names must adhere to the following naming restrictions:
 
-- Must contain two or more characters
-- Cannot start with an underscore (`_`)
-- Cannot contain whitespace characters, double quotes (`"`), or percent signs (`%`)
+- Cannot contain whitespace, punctuation, or special characters.
+  Only alphanumeric, underscore (`_`), dash (`-`), and forward-slash
+  (`/`) characters are permitted.
+- Should not start with an underscore (`_`)
```
jstirnaman, line 99:
```suggestion
- Should not start with an underscore (`_`).
```

content/influxdb/clustered/admin/databases/create.md
```diff
@@ -89,9 +89,11 @@ The retention period value cannot be negative or contain whitespace.
 
 Database names must adhere to the following naming restrictions:
 
-- Must contain two or more characters
-- Cannot start with an underscore (`_`)
-- Cannot contain whitespace characters, double quotes (`"`), or percent signs (`%`)
+- Cannot contain whitespace, punctuation, or special characters.
+  Only alphanumeric, underscore (`_`), dash (`-`), and forward-slash
+  (`/`) characters are permitted.
+- Should not start with an underscore (`_`)
```
jstirnaman, line 95:
```suggestion
- Should not start with an underscore (`_`).
```

jstirnaman (approved):
Looks good. Missing period.

PR #5195, opened 2023-10-25
fix(v3): remove v2-prefer note, replace with specific recommendations

jstirnaman:
- removes the shared v2-prefer shortcode note.
- adds "Flight client" tag to all Flight RPC and Flight SQL client pages.
- replaces the note with specific write/query notes in each platform.
- misc fixes

Closes #

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud-dedicated/reference/client-libraries/v2/go.md
+++ b/content/influxdb/cloud-dedicated/reference/client-libraries/v2/go.md
@@ -16,8 +16,7 @@ Use the [InfluxDB Go client library](https://github.com/influxdata/influxdb-clie
 {{% note %}}
 ### Use the InfluxDB v3 client library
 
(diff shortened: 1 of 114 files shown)


content/influxdb/cloud-dedicated/reference/client-libraries/v2/go.md
```diff
@@ -16,8 +16,7 @@ Use the [InfluxDB Go client library](https://github.com/influxdata/influxdb-clie
 {{% note %}}
```
sanderson, line 16: Wondering if this should be a warning instead of a note. Not super opinionated about it though 😄

sanderson (approved):
Looks good! Thanks for this.

sanderson (approved):

PR #5196, opened 2023-10-26
Telegraf: update and fix install instructions

### LINKED ISSUE #5087: Telegraf: After next release, remove Win workaround
We can remove this with the next Telegraf release.

_Originally posted by @jstirnaman in https://github.com/influxdata/docs-v2/pull/5086#discussion_r1294733660_

jstirnaman:
-  remove win workaround for previous release (Closes #5087)
- adds instructions for choosing an RPM repo package and constructing the command.
- adds new brew prefix and config file location
- adds Windows-specific syntax
- add headings
- cleans up duplicate content
- grammar

### DIFF
--- a/content/telegraf/v1/install.md
+++ b/content/telegraf/v1/install.md
@@ -320,20 +320,7 @@ The remaining instructions assume that `telegraf.exe` and `telegraf.conf` files 
 `C:\Program Files\InfluxData\telegraf` or that you created a Symlink to point to this directory.
     {{% /note %}}
 
-3.  Optional: Disable the [`inputs.processes` plugin](/telegraf/v1/plugins/#input-processes).
-    This plugin doesn't support Windows and returns an error when run with the `--test` flag.
-    Open `telegraf.conf` in your editor and comment the `inputs.processes` configuration lines.
-
-    ```toml
-    ...
-    # This plugin ONLY supports non-Windows
-    # [[inputs.processes]]
-    ...
-    #  # use_sudo = false
-    ...
-    ```
-
-4.  Optional: Enable a plugin to collect Windows-specific metrics--for example, uncomment the [`inputs.win_services`  plugin](/telegraf/v1/plugins/#input-win_services) configuration line:
+3.  Optional: Enable a plugin to collect Windows-specific metrics--for example, uncomment the [`inputs.win_services`  plugin](/telegraf/v1/plugins/#input-win_services) configuration line:
 
     ```toml
     ...
@@ -343,7 +330,7 @@ The remaining instructions assume that `telegraf.exe` and `telegraf.conf` files 
     ...
     ```
 
-5.  Run the following command to install Telegraf and the configuration as a Windows service.
+4.  Run the following command to install Telegraf and the configuration as a Windows service.
     For the `--config` option, pass the absolute path of the `telegraf.conf` configuration file.
 
     ```powershell
@@ -351,7 +338,7 @@ The remaining instructions assume that `telegraf.exe` and `telegraf.conf` files 
     --config "C:\Program Files\InfluxData\telegraf\telegraf.conf"
     ```
 
-6.  To test that the installation works, enter the following command:
+5.  To test that the installation works, enter the following command:
 
     ```powershell
     .\telegraf.exe `
--- a/content/telegraf/v1/install.md
+++ b/content/telegraf/v1/install.md
@@ -102,7 +102,7 @@ To manually install the Debian package from a `.deb` file:
 {{% /tab-content %}}
 <!---------- BEGIN RedHat & CentOS ---------->
 {{% tab-content %}}
-To learn how to manually install the RPM package from a file, see the [downloads page](https://influxdata.com/downloads/).
+To learn how to manually install the RPM package from a file, see the [downloads page](https://portal.influxdata.com/downloads/).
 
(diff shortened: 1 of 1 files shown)


content/telegraf/v1/install.md
```diff
@@ -100,68 +99,44 @@ To manually install the Debian package from a `.deb` file:
...
-Install telegraf once the repository is added to the `yum` configuration:
-
-```bash
-sudo yum install telegraf
-```
+To learn how to manually install the RPM package from a file, see the [downloads page](https://influxdata.com/downloads/).
+
+Use the `yum` package manager to install the latest stable version of Telegraf:
+
+1. In the [InfluxData package repository](https://repos.influxdata.com/stable/), find your system architecture name in the file list (for example, `x86_64`), and then copy the repository URL for your architecture (for example, https://repos.influxdata.com/stable/x86_64).
```
powersj, line 109: Why are you using this approach or the existing?

If the user uses the `\$basearch` these steps are handled for them by yum and there is no need for them to navigate the repo website or export the baseurl variable.

content/telegraf/v1/install.md
```diff
@@ -396,47 +348,125 @@ The following commands are available:
...
+1. [Prerequisites](#prerequisites)
+2. [Build the custom builder tool](#build-the-custom-builder-tool)
+3. [Run the custom builder to create a `telegraf` binary](#run-the-custom-builder-to-create-a-telegraf-binary)
 
-- Ensure you've installed [Go](https://go.dev/) version is 1.18.0 or later.
-- Create your Telegraf configuration file with the plugins you want to use. For details, see [Configuration options](/telegraf/v1/configuration/).
+### Prerequisites
 
-### Build and run the custom builder
+-  Install [Go](https://go.dev/) version 1.18.0 or later.
```
powersj, line 413: We usually are on the latest Go version, so I would not say 18 or later. If they are behind, they won't be able to build as it will say something to the effect of "requested go version not found"

content/telegraf/v1/install.md
```diff
@@ -203,79 +159,88 @@ It can be installed by running:
...
+<!---------- BEGIN MACOS ---------->
 {{% tab-content %}}
 Users of macOS 10.8 and higher can install Telegraf using the [Homebrew](http://brew.sh/) package manager.
-Once `brew` is installed, you can install Telegraf by running:
 
-```bash
-brew update
-brew install telegraf
-```
+1.  If you haven't already, follow the instructions to install the [Homebrew](http://brew.sh/) package manager.
```
powersj, line 170: I will point out that the homebrew package is not a static binary and not the same binary that we provide on the downloads page. This is fine, but it does mean that users find out when something goes wrong.

In practice other than the homebrew binary being static, it does mean that the CPU plugin for example works since it needs CGO support, while the binary we build does not support the CPU plugin.

powersj (commented):
Thank you for cleaning these up! I primarily looked through in terms of accuracy of the content. A few comments/questions below.

sanderson (approved):
Pending changes request by @powersj, looks good. Just one minor suggestion.

content/telegraf/v1/install.md
jstirnaman: > Why are you using this approach or the existing?
> 
> If the user uses the `\$basearch`

My lack of understanding. I didn't realize `$basearch` was a yum variable. Thanks!

content/telegraf/v1/install.md
jstirnaman: > We usually are on the latest Go version, so I would not say 18 or later.

👍 Questioned that, but forgot to call it out.

content/telegraf/v1/install.md
```diff
@@ -203,79 +159,88 @@ It can be installed by running:
...
+2.  Enter the following commands to update brew and install Telegraf:
 
-Run one of the following commands to start Telegraf and begin collecting and processing metrics:
+    ```zsh
+    brew update && brew install telegraf
+    ```
 
-To have launchd start Telegraf as a service after the next login:
+    The path to the `telegraf.conf` configuration file depends on the `brew` _prefix_ for your system:
```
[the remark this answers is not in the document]
jstirnaman: > Should this be depends on the architecture of your system? Or is prefix a homebrew-ism for arch?

[Homebrew-ism](https://docs.brew.sh/Shell-Completion). Yes, "architecture" is probably better.

jstirnaman:
@powersj @sanderson Please review fixes in https://github.com/influxdata/docs-v2/pull/5196/commits/b5c99f91084fe41e34e66e9da856ef707144d0e0

content/telegraf/v1/install.md
```diff
@@ -100,182 +99,167 @@ To manually install the Debian package from a `.deb` file:
...
-### Create a configuration file with default input and output plugins.
+### Install using Homebrew
 
-Every plugin will be in the file, but most will be commented out.
+Users of macOS 10.8 and higher can install Telegraf using the [Homebrew](http://brew.sh/) package manager.
 
-```
-telegraf config > telegraf.conf
-```
+{{% note %}}
```
jstirnaman, line 177: @sanderson Suggestions to say this better?

sanderson: I think this is fine as-is. Looks good.

sanderson (approved):

powersj (approved):
Thank you!

PR #5199, opened 2023-10-26
Alternate Clustered deployment with kubit

sanderson:
Closes https://github.com/influxdata/project-clustered/issues/157

Adds information about deploying a cluster using the kubit operator. I made many assumptions in this first draft and expect changes upon review.

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/install/deploy.md
+++ b/content/influxdb/clustered/install/deploy.md
@@ -41,7 +41,7 @@ your cluster, but from your terminal.**
 - You do not want to run the operator in your Kubernetes cluster.
 {{% /note %}}
 
-<!-- Hidden anchor for links to the kubctl/kubit tabs -->
+<!-- Hidden anchor for links to the kubectl/kubit tabs -->
 <span id="kubctl-or-kubit"></span>
 
 {{< tabs-wrapper >}}
--- a/content/influxdb/clustered/install/deploy.md
+++ b/content/influxdb/clustered/install/deploy.md
@@ -20,7 +20,7 @@ following tools:
 
 InfluxDB Clustered uses an `AppInstance` Kubernetes custom resource (CR) to
 configure and deploy your InfluxDB Cluster.
-Installing a CR requires cluster-wide permissions and may cause `kubectl` to
+Installing a `CustomResourceDefinition` (CRD) requires cluster-wide permissions and may cause `kubectl` to
 fail if you do not have those permissions in your cluster.
 
 `kubit` uses your local `kubectl` credentials to install CRs.
@@ -30,15 +30,14 @@ in the cluster.
 
 {{% note %}}
 **If you meet any of the following criteria, [install and use the `kubit` CLI](https://github.com/kubecfg/kubit#cli-tool)
-on your local machine to use it as an operator from your local terminal and
-deploy your cluster:**
+on your local machine. This allows you to act as the operator would and deploy your cluster, but from your terminal**
 
-- You do not have permissions to install a CR.
+- You do not have permissions to install a CRD.
 - You do not have permissions to install the operator in the `kubit` namespace.
 - You do not have permissions to create cluster-wide role-based access
   control (RBAC).
 - You want to preview the generated YAML.
-- You do not want to run the operator in your InfluxDB cluster.
+- You do not want to run the operator in your Kubernetes cluster.
 {{% /note %}}
 
 {{< tabs-wrapper >}}
--- a/content/influxdb/clustered/install/deploy.md
+++ b/content/influxdb/clustered/install/deploy.md
@@ -23,7 +23,7 @@ configure and deploy your InfluxDB Cluster.
 Installing a CRD requires cluster-wide permissions and may cause `kubectl` to
 fail if you do not have those permissions in your cluster.
 
-`kubit` uses the your local `kubectl` credentials to install CRDs.
+`kubit` uses your local `kubectl` credentials to install CRDs.
 This still requires you to have permissions to install cluster-wide resources
 (CRDs), but doesn’t require `kubit` to be installed with those permissions
 in the cluster. 
--- a/content/influxdb/clustered/install/deploy.md
+++ b/content/influxdb/clustered/install/deploy.md
@@ -14,24 +14,26 @@ This guide provides instructions for deploying your InfluxDB cluster using the
 following tools:
 
 - **kubectl**: CLI for controlling the Kubernetes cluster manager
-- **kubit operator**: A Kubernetes controller that can render and apply jsonnet
+- **kubit**: A Kubernetes controller that can render and apply jsonnet
   templates based on the [kubecfg](https://github.com/kubecfg/kubecfg) jsonnet
   tooling and framework
 
-InfluxDB Clustered uses an `AppInstance` Kubernetes custom resource (CRD) to
+InfluxDB Clustered uses an `AppInstance` Kubernetes custom resource (CR) to
 configure and deploy your InfluxDB Cluster.
-Installing a CRD requires cluster-wide permissions and may cause `kubectl` to
+Installing a CR requires cluster-wide permissions and may cause `kubectl` to
 fail if you do not have those permissions in your cluster.
 
-`kubit` uses your local `kubectl` credentials to install CRDs.
+`kubit` uses your local `kubectl` credentials to install CRs.
 This still requires you to have permissions to install cluster-wide resources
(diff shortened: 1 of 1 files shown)


content/influxdb/clustered/install/deploy.md
```diff
@@ -9,6 +9,44 @@ menu:
...
+- **kubit operator**: A Kubernetes controller that can render and apply jsonnet
+  templates based on the [kubecfg](https://github.com/kubecfg/kubecfg) jsonnet
+  tooling and framework
+
+InfluxDB Clustered uses an `AppInstance` Kubernetes custom resource (CRD) to
+configure and deploy your InfluxDB Cluster.
+Installing a CRD requires cluster-wide permissions and may cause `kubectl` to
+fail if you do not have those permissions in your cluster.
+
+`kubit` uses the your local `kubectl` credentials to install CRDs.
```
jstirnaman, line 26:
```suggestion
`kubit` uses your local `kubectl` credentials to install CRDs.
```

jstirnaman (approved):
Typo, otherwise LGTM 😅

content/influxdb/clustered/install/deploy.md
```diff
@@ -18,6 +56,25 @@ kubectl apply \
...
+
+1.  [Install the `kubit` operator](https://github.com/kubecfg/kubit/#installation)
+  and related tools.
+
+2.  Use the `kubit local apply` command to apply your custom-configured
+    `myinfluxdb.yml` and deploy your InfluxDB Cluster.
+
+    ```sh
+    kubit local apply myinfuxdb.yml
+    ```
```
jdockerty, line 72: Step 1 here requires the elevated permissions to install the `AppInstance` CRD. Once you run the 
`kubectl apply -k ...` part, that means the operator is running in the target cluster and there is no need to run `kubit local apply` (because it is already running).

I believe step 2 has been confused with the "I have minimal permissions/I'd like to act as the operator and try things out" possibility in `kubit`. If step 1 has been done, then a customer should run `kubectl apply -f myinfluxdb.yml`, i.e. "apply my `AppInstance` custom resource that I have" [1] and let the `kubit` operator handle that resource in the way it has been programmed to do.

By running the `kubit local apply` step, it would circumvent the previously installed operator entirely, as under the hood the `local apply` means "untar the given OCI bundle/package (Clustered), populate the required values, render the manifests as the operator would and then apply them". If a customer wants `kubit` to handle all of that then they should pass the `AppInstance` for Clustered to the Kubernetes API server (`kubectl apply`) and allow `kubit` to do the rest of it once it sees that the object has been created.

I understand that this can be quite confusing though, so I'm happy to go over it in more detail/step through it if you like too :blush: 

---
1. _Technically_ this does work, but only if you are wanting to act as the operator to test things out and step 1 isn't required in that case. I think it is potentially confusing to customers that they would _require_ `kubit` in order to apply their `AppInstance` (`myinfluxdb.yml` file), when in fact it is simply the regular `kubectl apply` operation that they would be used to.

jdockerty (commented):
Great write up! :rocket: 

I've added some clarifying comments/additional points to where I think there may be some confusion. Happy to go into more detail or clarify further if needed, just let me know :blush:

content/influxdb/clustered/install/deploy.md
```diff
@@ -9,6 +9,46 @@ menu:
...
+following tools:
+
+- **kubectl**: CLI for controlling the Kubernetes cluster manager
+- **kubit**: A Kubernetes controller that can render and apply jsonnet
+  templates based on the [kubecfg](https://github.com/kubecfg/kubecfg) jsonnet
+  tooling and framework
+
+InfluxDB Clustered uses an `AppInstance` Kubernetes custom resource (CR) to
+configure and deploy your InfluxDB Cluster.
+Installing a CR requires cluster-wide permissions and may cause `kubectl` to
```
jdockerty, line 23:
```suggestion
Installing a `CustomResourceDefinition` (CRD) requires cluster-wide permissions and may cause `kubectl` to
```

My apologies here, my explanation was a little poor on the CR vs CRD:

- `CustomResourceDefinition` is the API object in Kubernetes which defines the type.
- Custom Resource is an actual object which is not native to Kubernetes, watched by a custom controller. For example, an `AppInstance` is a custom resource watched by `kubit`.

It is the CRD itself which requires elevated permissions to install, because it installs a new object onto the entire cluster, i.e. that object becomes available to use in every namespace.

content/influxdb/clustered/install/deploy.md
```diff
@@ -9,6 +9,46 @@ menu:
...
+fail if you do not have those permissions in your cluster.
+
+`kubit` uses your local `kubectl` credentials to install CRs.
+This still requires you to have permissions to install cluster-wide resources
+(CRs), but doesn’t require `kubit` to be installed with those permissions
+in the cluster.
```
jdockerty, line 29:
```suggestion
`kubit` uses your local `kubectl` credentials to install CRs.
This still requires you to have permissions to install cluster-wide resources
(CRs), but doesn’t require `kubit` to be installed with those permissions
in the cluster. 
```

I _sort of_ understand what this is trying to say, but I think it's quite confusing (partially because my explanation was poor beforehand).

In my opinion, this can likely be removed or rejigged to something to the effect of:

"Your local `kubectl` credentials will be used to install the `AppInstance` CRD. If you do not have those permissions available to you, an alternative is to install the package yourself. To do so, please install the [`kubit`](https://github.com/kubecfg/kubit#cli-tool) tool."

---

Below this I see there is mention of the CLI tool and `kubit local apply`, so I'm not sure exactly how this should be structured, but my thinking is either a quick mention to prepare an install OR simply linking down to the alternative if CRD installation is not possible.

content/influxdb/clustered/install/deploy.md
```diff
@@ -9,6 +9,46 @@ menu:
...
+
+`kubit` uses your local `kubectl` credentials to install CRs.
+This still requires you to have permissions to install cluster-wide resources
+(CRs), but doesn’t require `kubit` to be installed with those permissions
+in the cluster. 
+
+{{% note %}}
+**If you meet any of the following criteria, [install and use the `kubit` CLI](https://github.com/kubecfg/kubit#cli-tool)
+on your local machine to use it as an operator from your local terminal and
+deploy your cluster:**
```
jdockerty, line 34:
```suggestion
**If you meet any of the following criteria, [install and use the `kubit` CLI](https://github.com/kubecfg/kubit#cli-tool)
on your local machine. This allows you to act as the operator would and deploy your cluster, but from your terminal**
```

My suggested wording might not be the best, but I think this makes more sense as the `kubit local apply` steps are there to allow a user to act as though _they_ were the operator instead from their terminal.

content/influxdb/clustered/install/deploy.md
```diff
@@ -9,6 +9,46 @@ menu:
...
+
+- You do not have permissions to install a CR.
```
jdockerty, line 36:
```suggestion
- You do not have permissions to install a CRD.
```

Similar to the explanation above, my bad on the poor explanation beforehand.

content/influxdb/clustered/install/deploy.md
```diff
@@ -9,6 +9,46 @@ menu:
...
+- You do not have permissions to install the operator in the `kubit` namespace.
+- You do not have permissions to create cluster-wide role-based access
+  control (RBAC).
+- You want to preview the generated YAML.
+- You do not want to run the operator in your InfluxDB cluster.
```
jdockerty, line 41:
```suggestion
- You do not want to run the operator in your Kubernetes cluster.
```

I believe this could be a typo, since it is a K8s cluster where the operator is installed rather than an InfluxDB cluster.

I might be missing some specific context though too :smile: Do we refer to a customer's K8s cluster as their "InfluxDB cluster"? Ignore me if so.

content/influxdb/clustered/install/deploy.md
```diff
@@ -40,6 +41,9 @@ on your local machine. This allows you to act as the operator would and deploy y
 - You do not want to run the operator in your Kubernetes cluster.
 {{% /note %}}
 
+<!-- Hidden anchor for links to the kubctl/kubit tabs -->
```
jdockerty, line 44:
```suggestion
<!-- Hidden anchor for links to the kubectl/kubit tabs -->
```

Minor typo here/where it's used, but otherwise this LGTM :rocket: Great stuff!

jdockerty (approved):

PR #5202, opened 2023-10-27
chore(v3): There is no index of Tag Values, Tag Values are "just data…

### LINKED ISSUE #5097: There is no index of Tag Values, Tag Values are "just data" in IOx.
This is a change from TSM, there is no index of Tag Values, Tag Values are "just data" in IOx. Tag Keys and other meta data are stored in the Catalog, so those are fast queries. I would still suggest using InfluxQL for these queries though because it is easier for humans to read and write it.

<@U9Q8DT96C> <@U01TCPBF3U4> have we documented this nuance anywhere?

[Slack Message](https://influxdata.slack.com/archives/C04768675QD/p1692798522947029?thread_ts=1692732617.941009&cid=C04768675QD)

jstirnaman:
…" in IOx. #5097

- Fix example

Closes #5097


sanderson (approved):
Looks good!

PR #5207, opened 2023-10-30
5025 fix and test python examples

jstirnaman:
Add replacement patterns for tests.
Fix wrong variable name.
Setup as shell and python tests, testing fixes.


content/influxdb/cloud-dedicated/get-started/query.md
```diff
@@ -364,24 +380,21 @@ _If your project's virtual environment is already running, skip to step 3._
...
       fh = open(certifi.where(), "r")
       cert = fh.read()
       fh.close()
 
       client = InfluxDBClient3(
-          host="{{< influxdb/host >}}",
-          token=TOKEN,
-          database="get-started",
+          host=f"{{< influxdb/host >}}",
+          token=f"DATABASE_TOKEN",
```
sanderson, line 394: Where these are being replace with "static" values, do these need to be f-strings? They shouldn't need any interpolation unless they're assigning these to a local or environment variable.

sanderson (approved):
Looks good. Just one question.

PR #5210, opened 2023-10-31
Update file names for Enterprise 1.11.3 packages

sanderson:
Closes influxdata/DAR#399

- [x] Rebased/mergeable


aaronkoerner (approved):
Looks good, though the way the formatting is done I'm guessing it may need to be updated again when the next version is released? Assuming the "-1" in the name was just because of a rebuild or something like that... can't say I know how that gets established

sanderson:
Yeah, I've made a mental note to check on the next enterprise release. It will be easy enough to just find and replace (`{{< latest-patch >}}-1` -> `{{< latest-patch >}}`).

PR #5214, opened 2023-11-03
5204 v2 update downloads and install instructions

### LINKED ISSUE #5204: Document Tree is displayed when clicked on the Download button
_Dbutton1

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2/install/
- _Provide other relevant URLs (project issues, community threads, etc.)_
Error

jstirnaman:
Closes #5204

- Update download URLs and paths for all distributions.
- Update and fix install instructions, make consistent between browser downloading and shell
- Make testable.

- test.sh diffs against origin/master
- test.sh builds the image for each run and then removes the container after the run.
- tesh.sh adds examples for running and debugging to remind myself.
- compose.yaml adds a bind mount to copy docs-v2/data into docs-v2/test.
- Dockerfile: simplify path use
- run-tests.sh: make default run less verbose, add support for a verbose option (and future options).
- run-tests.sh: generates a .env.products file from data/products.yml and sources the file variables as environment variables to use in test runs.
- run-tests.sh: adds some misc. test setup.
- requirements.txt: return to using the influxdb3-python release

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,8 @@ FROM python:${PYTHON_VERSION}-slim as base
 RUN apt-get update && apt-get install -y \
   curl \
   git \
+  gpg \
+  wget \
    && rm -rf /var/lib/apt/lists/*
 
 ARG SOURCE_DIR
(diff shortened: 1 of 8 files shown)


sanderson (approved):
This looks great! Thanks for doing this.

PR #5224, opened 2023-11-08
fix: remove static token permissions verbiage

eatondustin1:
This removes a section from the Dedicated and Clustered docs that says that token permissions cannot be updated since influxctl now supports token updates.

Closes #

- [x] Rebased/mergeable


sanderson (approved):
Thanks @eatondustin1!

PR #5241, opened 2023-11-17
Upgrade hugo to latest version 0.120.4

deining:
This PR bumps hugo site generator to latest released version v0.120.4.

- [x] Rebased/mergeable


.circleci/config.yml
```diff
@@ -2,12 +2,12 @@ version: 2
 jobs:
   build:
     docker:
-      - image: cimg/node:14.16
+      - image: cimg/node:20.9.0
         environment:
-          HUGO_VERSION: "0.81.0"
-          S3DEPLOY_VERSION: "2.3.5"
+          HUGO_VERSION: "0.120.4"
```
sanderson, line 7: The `HUGO_VERSION` variable is not longer used in the deploy process and can probably be used. We moved from downloading static `hugo` binaries to loading Hugo as a node dependency and executing it with `npx`. Looks like this didn't get removed when that change was made.

```suggestion
```

sanderson (changes requested):
Thanks for this @deining! One minor suggestion.

.circleci/config.yml
deining: Line was removed, as requested.

sanderson (approved):
Looks great! Thank you @deining!

PR #5264, opened 2023-12-06
Fix-5226 v2 and v3 series inaccuracy

### LINKED ISSUE #5226: Fix series definitions and reasons for rejected points
In OSS, field keys are not part of the series key.

The following is false for OSS v2, but true for Cloud v2:

 > A series key is a collection of points that share a measurement, tag set, and field key
https://docs.influxdata.com/influxdb/v2/reference/key-concepts/data-elements/#series

https://docs.influxdata.com/influxdb/v2/reference/glossary/#series
The glossary is right. We probably need to add some oss-/cloud-only shortcode there.

[Slack Message](https://influxdata.slack.com/archives/D01V6BP761M/p1699640867661179)

jstirnaman:
- Fixes v2 series definitions
- Remove questionable example for schema design
- Fixes v3 reasons for rejected points.

 closes #5226

### DIFF
--- a/content/influxdb/cloud-dedicated/write-data/troubleshoot.md
+++ b/content/influxdb/cloud-dedicated/write-data/troubleshoot.md
@@ -19,16 +19,12 @@ related:
 
 Learn how to avoid unexpected results and recover from errors when writing to {{% product-name %}}.
 
-<!-- TOC -->
-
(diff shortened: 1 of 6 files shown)


sanderson (approved):
👌🏻 Looks great! Thanks!

PR #5275, opened 2023-12-14
chore: add note about partial write failures

jacobmarble:
Helps https://github.com/influxdata/influxdb_iox/issues/5226

Small clarification to 204 responses to write requests.

- [ ] Rebased/mergeable


api-docs/cloud-serverless/ref.yml
```diff
@@ -14680,8 +14680,8 @@ tags:
       | &nbsp;Code&nbsp; | Status                   | Description      |
       |:-----------:|:------------------------ |:--------------------- |
       | `200`       | Success                  |                       |
-      | `201`       | Created                           | Successfully created a resource. The response body may contain details. |
-      | `204`       | No content      | The request succeeded, but InfluxDB doesn't typically return a response body for the operation. |
+      | `201`       | Created                  | Successfully created a resource. The response body may contain details. |
```
wiedld, line 14683: 201 no longer exists in serverless?

jstirnaman: This table isn't specific to writes--we still have v2 endpoints in Serverless that respond with `201`, correct?

api-docs/cloud-serverless/ref.yml
```diff
@@ -14680,8 +14680,8 @@ tags:
...
+      | `204`       | No content               | The request succeeded. For _writes_, a response body is returned if part of the request failed. See the write API spec for details. |
```
jstirnaman, line 14684:
```suggestion
      | `204`       | No content               | The request succeeded. For _writes_ that have [rejected points](/influxdb/cloud-serverless/write-data/troubleshoot/#troubleshoot-rejected-points), the [`/write` and `/api/v2/write` endpoints](#tag/Write) return a response body with error details. |
```

Thanks for this. I chose to remove the blurb about writes, and just leave it more generic (maybe a bit of self-serving DRYness), but we can add it back.

jstirnaman (changes requested):
Thanks for checking that--just a suggested edit.

api-docs/cloud-serverless/ref.yml
jacobmarble: In the context of this change, I decided to leave 201. This section of the doc seems to be about the API generally, not the write API specifically.

jacobmarble: TBH, I thought about whether we do or don't return 201 anywhere, and I chose not to go searching through the IOX and IDPE repos.

PR #5278, opened 2023-12-15
Add influxctl instructions to query and write task-based docs

sanderson:
- Updates the getting started docs in Dedicated and Clustered with instructions for using `influxctl write` and `influxctl query`
- Adds instructions for using `influxctl query` to the "Execute queries" section in Dedicated and Clustered.

**Note**: I plan on adding `influxctl write` instructions to the "Write data" section, but I think it's going to require a minor restructure. I'll do that in another PR.

- [x] Rebased/mergeable


jstirnaman (approved):
🥇

PR #5169, opened 2023-10-10
feature(v3): Write: partial writes write good lines and reject bad li…

jstirnaman:
…nes #5160

- Documents the proposed partial writes feature for Serverless.
- Updates Troubleshoot writes and API docs.

Closes #5160

Issue #5160, opened 2023-10-02
Write: partial writes write good lines and reject bad lines


jstirnaman:
See detail in https://github.com/influxdata/influxdb_iox/issues/5226#issuecomment-1716158763

Response:
`400`

```
{
  "code": "invalid",
  "line": 123,
  "message": "human-readable error message"
}
```
[Slack Message](https://influxdata.slack.com/archives/C04768675QD/p1696034786933909?thread_ts=1695822639.441929&cid=C04768675QD)

wiedld:
Note that partial writes is not the default. It has to be [specifically enabled](https://github.com/influxdata/influxdb_iox/issues/5226#issuecomment-1741239380) per cluster. Currently it's only on in stag101  stag102. Not sure who/how we decide where else to turn on.

wiedld:
For anything with an iox backend:
```
{
  "code": "invalid",
  "line": 123, // line number of first error encountered
  "message": "<human-readable error message for 1st line error>\n<human-readable error message for 1st line error>\n<human-readable error message for Nth line error up to N=100 lines>"
}
```

This^^ error is returned regardless of whether or not partial writes is enabled.

PR #5280, opened 2023-12-15
Revert "feature(v3): Write: partial writes write good lines and reject bad li…"

jstirnaman:
Reverts influxdata/docs-v2#5169
 due to the feature being reverted in https://github.com/influxdata/k8s-idpe/pull/35667

The commit was part of https://github.com/influxdata/docs-v2/issues/5160


savage-engineer:
Thank you for noticing the revert, much appreciated!  🙏

I did not think to check if there were any docs to revert at the time

PR #5294, opened 2024-01-08
fix(ent): incorrect influxd-ctl ldap links

jstirnaman:
Closes  https://github.com/influxdata/DAR/issues/409

- Fixes copy-pasta error, replaces /enterprise_influxdb/v1/reference/cli/ with /enterprise_influxdb/v1/tools/


sanderson (approved):
👍🏻

PR #5309, opened 2024-01-25
Document v3 custom partitions

### LINKED ISSUE #5289: Document custom partitioning
**ETA:** Next `influxctl` release
**PR:** https://github.com/influxdata/influxctl/pull/245#discussion_r1434487078

Custom partitioning allows users to custom define how their data is partitioned in InfluxDB v3 (Cloud Dedicated and Clustered). Partitioning is managed when creating a database using the `influxctl` CLI. Technically, it can also be done per measurement, but I'm still unclear how an end user would manage per-measurement (per-table) partitioning schemes. It can only be done at creation time. Once a database or table is created, the partitioning strategy cannot be updated.

There are performance implications when using custom partitioning:

**Advantages**:

- Improved query performance when the the partitioning strategy matches the query pattern of the workload. For example, if you query by a specific value of a tag, partitioning by that tag will improve query performance.

**Disadvantages**:

- Increased load on the Catalog as more references to Parquet files are stored
- Increased load on the Compactor as it has to compact more Parquet files

---

Items to add and update in both the Dedicated and Clustered docs:

- [x] Add guide for configuring custom partitioning to "Admin" section.
    - [x] Explain concepts of custom partitioning
    - [x] Explain the pros and cons of custom partitioning
    - [x] Explain how to define partitioning templates
- [ ] Update `influxctl database create` documentation
- [ ] Add information to query optimization guides

---

##### Relevant URLs
- https://github.com/influxdata/influxctl/pull/245#discussion_r1434487078
- [Internal custom partitioning docs](https://docs.influxdata.io/operations/specifications/iox_runbooks/local-test-cli/#how-to-define-customized-partition-template)

sanderson:
Closes #5289 

Provides information about the InfluxDB v3 storage engine architecture and custom partitioning.

DO NOT MERGE UNTIL THE NEXT MINOR VERSION OF `influxctl` IS RELEASED.

@powersj Some of the `influxctl` examples in here are speculative as I know other related functionality is still being built. Feel free to correct anything that is innacurate.

- [x] Rebased/mergeable


content/influxdb/cloud-dedicated/reference/internals/storage-engine.md
```diff
@@ -0,0 +1,180 @@
...
+    - ensure the schema assumed by the query plan matches the schema of written data.
+    - include recently written, yet-to-be-persisted data in query results.
+
+3.  Queries the [Catalog](#catalog) to identify the physical location of partitions
+    that contain the queried data.
+4.  Reads partition Parquet files from the [Object store](#object-store)
+    containing the queried data and scans each row to find those that match the
+    logic defined in the query plan.
+5.  Returns the data specified in the query and performs any additional
+    operations specified in the query plan.
```
jstirnaman, line 96:
```suggestion
    operations (for example: deduplicating, merging, and sorting) specified in the query plan.
```

content/influxdb/cloud-dedicated/reference/internals/storage-engine.md
```diff
@@ -0,0 +1,180 @@
...
+6.  Returns the query result to the client.
+
+##### Querier scaling strategies
+
+The Querier can be scaled both [vertically](#vertical-scaling) and
+[horizontally](#horizontal-scaling).
+Horizontal scaling increases query throughput to handle more concurrent queries.
+Vertical scaling improves the queriers ability to process computation-heavy queries.
```
jstirnaman, line 104:
```suggestion
Vertical scaling improves the Querier's ability to process computationally intensive queries.
```

content/influxdb/cloud-dedicated/reference/internals/storage-engine.md
```diff
@@ -0,0 +1,180 @@
...
+##### Catalog scaling strategies
+
+Scaling strategies available for the Catalog depend on the PostgreSQL-compatible
+database used to run the catalog. All support [vertical scaling](#vertical-scaling).
+Most support [horizontal scaling](#horizontal-scaling) for redundancy and failover. 
+
+### Object store
+
+The Object store contains time series data in [Apache Parquet](https://parquet.apache.org/) format.
+Each parquet file represents a partition.
```
jstirnaman, line 126:
```suggestion
Each Parquet file represents a partition.
```

content/influxdb/cloud-dedicated/admin/custom-partitions/partition-templates.md
```diff
@@ -0,0 +1,236 @@
+---
+title: Partition templates
+list_title: Use partition templates
+description: >
+  Learn how to define custom partitioning strategies using partition templates.
+  Data can be partitioned by tag and time.
+menu:
+  influxdb_cloud_dedicated:
+    parent: Manage data partitioning
```
jstirnaman, line 9:
```suggestion
    name: Partition templates
    parent: Manage data partitioning
```

content/influxdb/cloud-dedicated/admin/custom-partitions/partition-templates.md
```diff
@@ -0,0 +1,236 @@
...
+weight: 202
+---
+
+Use partition templates to define the patterned used to generate partition keys.
```
jstirnaman, line 13:
```suggestion
Use partition templates to define the patterns used to generate partition keys.
```

content/influxdb/cloud-dedicated/admin/custom-partitions/partition-templates.md
```diff
@@ -0,0 +1,236 @@
...
+A partition key uniquely identifies a partition and is used to name the partition
+Parquet file in [Object storage](/influxdb/cloud-dedicated/reference/internals/storage-engine/#object-storage).
```
jstirnaman, line 15:
```suggestion
Parquet file in the [Object store](/influxdb/cloud-dedicated/reference/internals/storage-engine/#object-store).
```

content/influxdb/cloud-dedicated/admin/custom-partitions/partition-templates.md
```diff
@@ -0,0 +1,236 @@
...
+
+A partition template consists of 1-8 _template parts_---dimensions to partition data by.
+There are two types of parts:
+
+- **tag**: [InfluxDB tag](/influxdb/cloud-dedicated/reference/glossary/#tag) to
+  partition by.
+  _A partition template can include up to 7 tag parts._
```
jstirnaman, line 22:
```suggestion
  _A partition template can include up to seven tag parts._
```

content/influxdb/cloud-dedicated/admin/custom-partitions/partition-templates.md
```diff
@@ -0,0 +1,236 @@
...
+### Reserved keywords
+
+The following reserved keywords cannot be used in partition templates:
+
+- `time`
+
+### Reserved Characters
+
+If used in template parts, non-ASCII characters and the following reserved
+characters must be percent encoded:
```
jstirnaman, line 58:
```suggestion
characters must be [percent encoded](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding):
```

content/influxdb/cloud-dedicated/admin/custom-partitions/partition-templates.md
```diff
@@ -0,0 +1,236 @@
...
+### Special specifiers
+
+| Variable | Example | Description             |
+| :------: | :------ | :---------------------- |
+|   `%t`   |         | Literal tab (`\t`).     |
+|   `%n`   |         | Literal newline (`\n`). |
+|   `%%`   |         | Literal percent sign.   |
+
+It is possible to override the default padding behavior of numeric specifiers `%?`.
+This is not allowed for other specifiers and will result in the `BAD_FORMAT` error.
```
jstirnaman, line 170:
```suggestion
This is not allowed for other specifiers and results in the `BAD_FORMAT` error.
```

jstirnaman (changes requested):
Nice work! Mostly nitpicky suggestions.

content/influxdb/cloud-dedicated/admin/custom-partitions/define-custom-partitions.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+## Create a measurement with a custom partition template
+
+The following example creates a new `example-meas` measurement and applies a
+partition template that partitions by two tags (`room` and `sensor-type`) and by
+week using the time format `%Y-%m-%d %H:00`:
+
+```sh
+influxctl measurement create \
```
powersj, line 69: Hmm I had assumed that I would call this `tables` instead of `measurement`. When creating, listing, and updating databases today with `influxctl`, we have `max tables` and `max columns per table` as options and in the output. I am inclined to stick with table to be consistent throughout.

I even see `Use SHOW TABLES to list measurements in your InfluxDB database.` in the docs, which feels like we are combining two terms.

content/influxdb/cloud-dedicated/admin/custom-partitions/define-custom-partitions.md
```diff
@@ -0,0 +1,97 @@
...
+  --template-tag room \
+  --template-tag sensor-type \
+  --template-time '%Y-%m-%d %H:00' \
+  example-meas
```
powersj, line 73: We probably will need to specify which database as well in this command.

powersj (commented):
Thanks for these! I skipped the scss/html pages and focused on the influxctl content and best practices. I have a couple comments inline, but would defer to @NGA-TRAN for a more technical review of the best practices.

content/influxdb/cloud-dedicated/admin/custom-partitions/define-custom-partitions.md
sanderson: Yeah, I had a feeling you were leaning this way. This brings up an interesting discussion that is honestly, probably bigger than this PR.

When describing the InfluxDB data structure, we still use "databases," "measurements," "fields," and "tags" which are all inherited from previous versions of InfluxDB and existing users have an established mental model based on these. With IOx and the first class citizen status of SQL, under the hood, we now use the traditional columnar data model (tables, columns, etc.) where TSM... wasn't that. Yes, under the hood, a "measurement" is represented by table and fields and tags are just columns in a table, but to users, we still push the measurement/tag/field model. So I guess the bigger question is--should we? 🤷🏻‍♂️

> I even see `Use SHOW TABLES to list measurements in your InfluxDB database.` in the docs, which feels like we are combining two terms.

This is specific to SQL and uses the metaqueries shipped with DataFusion. With InfluxQL, you'd use [`SHOW MEASUREMENTS`](https://docs.influxdata.com/influxdb/cloud-serverless/reference/influxql/show/#show-measurements).

I don't have the answer for this and I don't think you do either.

Maybe in the mean time, and as a middle-ground, `influxctl measurement` could be an alias of `influxctl table`?

content/influxdb/cloud-dedicated/admin/custom-partitions/define-custom-partitions.md
sanderson: Sounds good. I've just left this as a placeholder until the command is in place.

content/influxdb/cloud-dedicated/admin/custom-partitions/define-custom-partitions.md
powersj: > I don't have the answer for this and I don't think you do either.

Who is the decision maker?

> Maybe in the mean time, and as a middle-ground, influxctl measurement could be an alias of influxctl table?

We should have a preferred way of doing things and drive our users towards that. Adding two ways to do the same thing is something I would strongly like to avoid as the feedback from previous CLIs was the complexity and verbosity made them difficult to use.

sanderson: > Who is the decision maker?

I don't know who exactly would be, but I think @garylfowler, @barbaranelson, and @pauldix may want to weigh in (going right to the top 😄).

> We should have a preferred way of doing things and drive our users towards that. Adding two ways to do the same thing is something I would strongly like to avoid as the feedback from previous CLIs was the complexity and verbosity made them difficult to use.

I agree. My preference would be `measurement` since that's what we use _almost_ everywhere else (line protocol, data model descriptions, schemas, etc.).

jstirnaman (approved):

content/influxdb/cloud-dedicated/reference/internals/storage-engine.md
```diff
@@ -0,0 +1,178 @@
...
+The Querier handles query requests and returns query results for requests.
+It supports both SQL and InfluxQL through
+[Apache Arrow DataFusion](https://arrow.apache.org/datafusion/user-guide/introduction.html).
+
+#### Query life cycle
+
+At query time, the querier:
+
+1.  Receives the query request and builds a query plan.
+2.  Queries the [Ingester](#ingester) to:
```
NGA-TRAN, line 83: I wonder if you want to make `Ingester` her plural because the Querier (the one that handle the query) talks to all ingesters in the cluster for not-yet-persisted data

NGA-TRAN (approved):
I love it ❤️  and just have a couple of comments.
I cannot see the pictures and tables though

content/influxdb/cloud-dedicated/admin/custom-partitions/best-practices.md
```diff
@@ -0,0 +1,56 @@
...
+
+As you plan your partitioning strategy, keep in mind that data can be
+"over-partitioned"--meaning partitions are so granular that queries end up
+having to retrieve and read many partitions from the object store, which
+hurts query performance.
+
+- Avoid partition time intervals that are too small. A good rule of thumb is to
+  partition data using time intervals similar to your most commonly queried
+  time range. For example, if you often query data from the last hour, partitioning by
+  hour may improve query performance.
```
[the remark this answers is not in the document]
sanderson: Thanks @NGA-TRAN. I was curious about this an was hoping you mention something in review 😄. So is the recommend floor for time intervals 1 day?

content/influxdb/cloud-dedicated/admin/custom-partitions/best-practices.md
```diff
@@ -0,0 +1,56 @@
...
+
+Custom partitioning primarily benefits queries that look for a specific tag
+value in the `WHERE` clause. For example, if you often query data related to a
+specific ID, partitioning by the tag that stores the ID helps the InfluxDB
+query engine to more quickly identify what partitions contain the relevant data.
+
+{{% note %}}
+#### Be careful partitioning on high-cardinality tags
+
+Partitioning using tags with many (500K+) unique values can actually hurt
```
sanderson, line 30: @NGA-TRAN I just used the `500K+` number as a placeholder and don't have any evidence to suggest this is an accurate range for "high-cardinality" tags. Do you have any guidance here?

NGA-TRAN: If you really want a number, use `10K+`.

content/influxdb/cloud-dedicated/admin/custom-partitions/best-practices.md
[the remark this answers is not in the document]
NGA-TRAN: > recommend floor for time intervals 1 day

This sounds reasonable to me and I also think you want to avoid mention `A good rule of thumb is to partition data using time intervals similar to your most commonly queried   time range`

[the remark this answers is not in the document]
sanderson: Will do!

content/influxdb/cloud-dedicated/admin/custom-partitions/best-practices.md
sanderson: Thank you! It's not so much that _I_ want a number, but I know our customers will 😄 and will ask for it.

content/influxdb/cloud-dedicated/admin/custom-partitions/define-custom-partitions.md
garylfowler: My second preference would be table, since that is going to be a known term for people using us for the first time but familiar with other database.  But my first preference would be to allow for either (table or measurement) since we do talk about measurement still in a lot of places in our Docs, and certainly when we talk about line protocol and we also have customers migrating from previous versions.

pauldix: With 3.0 the preferred naming convention is database, tables, columns, rows. This will matter even more when we get to the 3.0 native data model and new write API.

I'd generally prefer not to use measurement as an alias unless we're talking specifically about line protocol or InfluxQL.

It's better to make it clear to the user that 3.0 has the model of organizing data into database > table > partition > parquet files and that we have a way to map 1.x and 2.x schema onto that model. But the sooner they start thinking about the model in the 3.0 way, the closer they'll be to creating a schema that will yield good results.

PR #5312, opened 2024-01-29
docs(clustered): fix incorrect padding and typos

jdockerty:
The padding for the list was off which caused the `span` tags were showing in the documentation.

A good eye by @mkmik spotted this :rocket: 

Beforehand it looked like this:

image

Now it looks like:

image

I've also added a few extras to help with the `config.toml` file and corrected a typo in the identity (typo as indentity) field which I spotted along the way.

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/get-started/setup.md
+++ b/content/influxdb/clustered/get-started/setup.md
@@ -80,12 +80,11 @@ If stored at a non-default location, include the `--config` flag with each
 {{% /code-placeholders %}}
 
 Replace the following with your {{< product-name >}} credentials:
-
-    - {{% code-placeholder-key %}}`PORT`{{% /code-placeholder-key %}}: the port to use to access your InfluxDB cluster
-    - {{% code-placeholder-key %}}`OAUTH_CLIENT_ID`{{% /code-placeholder-key %}}: the client URL of your OAuth2 provider
-      (for example: `https://indentityprovider/oauth2/v2/token`)
-    - {{% code-placeholder-key %}}`OAUTH_DEVICE_ID`{{% /code-placeholder-key %}}: the device URL of your OAuth2 provider
-      (for example: `https://indentityprovider/oauth2/v2/auth/device`)
+  - {{% code-placeholder-key %}}`PORT`{{% /code-placeholder-key %}}: the port to use to access your InfluxDB cluster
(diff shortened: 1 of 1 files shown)


mkmik (approved):

content/influxdb/clustered/get-started/setup.md
```diff
@@ -63,12 +63,12 @@ If stored at a non-default location, include the `--config` flag with each
...
     **Copy and paste the sample configuration profile code** into your `config.toml`:
 
-{{% code-placeholders "PORT|OAUTH_TOKEN_URL|OAUTH_DEVICE_URL" %}}
+{{% code-placeholders "PORT|HOST|OAUTH_TOKEN_URL|OAUTH_DEVICE_URL|OAUTH_CLIENT_ID" %}}
 ```toml
 [[profile]]
   name = "default"
   product = "clustered"
-  host = "{{< influxdb/host >}}"
+  host = "HOST"
```
sanderson, line 71: Host values are controlled by a global setting specified by the user. Code placeholders are scoped to a page and do not persist through a refresh. This should use the `influxdb/host` shortcode, not a code placeholder.

```suggestion
  host = {{< influxdb/host >}}
```

sanderson (changes requested):

content/influxdb/clustered/get-started/setup.md
mkmik: ah that's interesting. Should we use global placeholders for other stuff too? For example some of the oauth stuff is referenced by multiple pages.

sanderson: No, anything that could contain a sensitive secret should not be stored. We don't want to be responsible in any way for securing people's secrets. The only global, user-defined value we currently store is the host/url.

jdockerty: That makes sense! I'll revert this portion back

mkmik: The oauth config is not really a secret though. Is it still considered sensitive?

sanderson (approved):

content/influxdb/clustered/get-started/setup.md
mkmik: ok; but just for the record; these are not credentials; they are just hostnames and IDs that used to configure the mechanism that will securely provide credentials.

PR #5323, opened 2024-02-14
Add cookie migration

sanderson:
Adds a cookie migration process that checks for all old cookies, ports them to the new structure, and removes the old cookies. This introduces a new cookie to indicate whether or not cookies have been ported (`influxdata_docs_ported`) with a 30 day expiration. On March 15, I will remove the migration.

- [x] Rebased/mergeable


jstirnaman (approved):
Good idea. Looks good.

PR #5358, opened 2024-03-12
Correct yum meta node package name

sanderson:
Closes influxdata/DAR#414

Corrects the package name for the `yum` meta node package.

- [x] Rebased/mergeable


jstirnaman (approved):
👍

PR #5363, opened 2024-03-14
Add JSON page lists for use in the v3 UI

sanderson:
This change generates `pages.json` files for Dedicated, Serverless, and Clustered docs that provide the product docs navigation structure with doc names, URLs, and children. This will be used to link to documentation from the InfluxDB v3 management UI.

- [x] Rebased/mergeable


jstirnaman (approved):
LGTM

PR #5365, opened 2024-03-15
Replace query plan ASCII diagrams with HTML

sanderson:
This replaces all of the ASCII query plan diagrams with an HTML diagram that scales to the viewing size and shouldn't require horizontal scrolling on smaller screens.

image

- [x] Rebased/mergeable


jstirnaman (approved):
Thanks! Much better!

PR #5369, opened 2024-03-18
docs(clustered): data ingest lifecycle

jdockerty:
Closes #

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
+++ b/content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
@@ -1,0 +1,60 @@
+---
+title: Data ingest lifecycle best practices
+seotitle: InfluxDB best practices for data ingest lifecycle management
+description: >
+  Best practices for managing the lifecycle of data ingested into InfluxDB.
+menu:
+  influxdb_clustered:
+    name: Data ingest lifecycle
+    weight: 201
+    parent: write-best-practices
+---
+
+Data ingested into InfluxDB must conform to the retention period set on the database.
+Any data which lies outside of this retention period is no longer considered for retrieval by a query, but can have references maintained as in object storage or the Postgres catalog, causing an increase in operational overhead and drive higher costs.
+
+In order to reduce these factors, it is important to manage the lifecycle of data ingested into InfluxDB. You can achieve this by following the best practices below:
+
+- Use a retention policy which is appropriate for your requirements.
+- Tune the garbage collector cutoff period to ensure that data is removed in a timely manner.
+- Enable versioning in your object store and use lifecycle rules to expire old data.
+
+### Garbage collector tuning
+
+Once data falls outside of the retention period it now has the potential to be removed by the garbage collector.
+
+There are two variables which tune this process:
+
+- `INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF`: defines when objects are deleted from object storage. The default is `30d`.
+- `INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF`: defines when a row can be deleted from the catalog. The default is `30d`.
+
+You may set them to a value that is appropriate for your use case, for example a value of `6h` (6 hours) would be appropriate for maintaining a lean catalog that only maintains references to recent data.
+
+Our recommendation is to keep these values aligned. 
+
+### Lifecycle rules
+
+**We highly recommend that you use object versioning if you are expiring data using lifecycle rules.**
+
+You can use lifecycle rules to ensure that old versions of data are not kept around any longer than necessary - this could cause a significant increase in storage costs over a long period.
+
+We recommend that you use lifecycle rules to expire data after a set period of time, which is appropriate to your organisation. 
+
+For example, an [AWS S3 lifecycle rule](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html) to expire non-current objects after 90 days will look like the following:
+
+```json
+{
+    "Rules": [
+        {
+            "ID": "my-lifecycle-rule",
+            "Filter": {
+                "Prefix": ""
+            },
+            "Status": "Enabled",
+            "NoncurrentVersionExpiration": {
+                "NoncurrentDays": 90
+            }
+        }
+    ]
+}
+```


content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,82 @@
...
+---
+
+Data ingested into InfluxDB must conform to the retention period set on the database.
+Any data which lies outside of this retention period is no longer considered for retrieval by a query, but can have references maintained in object storage or the Postgres catalog, causing an increase in operational overhead and drive higher costs.
+
+In order to reduce these factors, it is important to manage the lifecycle of ingested data. You can achieve this by following the best practices below:
+
+- Use a retention policy which is appropriate for your requirements.
+- Tune the garbage collector cutoff period to ensure that data is removed in a timely manner.
+- Enable versioning in your object store and use lifecycle rules to expire old data.
```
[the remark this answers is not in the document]
jdockerty: > no need to enable object versioning to manage the file lifecycle

I was wondering whether this was being conflated as I was writing it.

My understanding was that the object versioning _enables_ our lower value of `6h` in Serverless/Dedicated, but perhaps that actually isn't quite right? Versioning is merely a mechanism for GC to delete things and creates a new "version" on the deletion - the catalog can be smaller because of this though.

The deleted version being around is a helper for DR, although from speaking with Raphael a bit, and as you say, the tooling is a little lacking for restores.

Is that a better/more correct understanding?

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Garbage collector tuning
+
+Once data falls outside of the retention period it can now be removed by the garbage collector service.
+
+There are two variables which tune this process:
+
+- `INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF`: defines when objects are deleted from object storage. The default is `30d`.
+- `INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF`: defines when a row can be deleted from the catalog. The default is `30d`.
```
savage-engineer, line 29: I reckon these could benefit from being a bit more explicit here, so the interaction is clearer. Perhaps something like this:

```suggestion
- `INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF`: the age at which a parquet file not referenced in the catalog becomes eligible for deletion from object storage. The default is `30d`.
- `INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF`: defines how long to retain the row referencing a parquet file in the catalog once marked for deletion. The default is `30d`.
```

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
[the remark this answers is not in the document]
domodwyer: > My understanding was that the object versioning enables our lower value of 6h in Serverless/Dedicated, but perhaps that actually isn't quite right?

Note quite - lets assume that you, as a customer, do not care about backups - there's no point keeping any file longer than you need it. As soon as a file is no longer queryable (expired due to being outside the retention period) then there's no point in keeping that file. It can be deleted immediately![^kinda] Set your GC retention to 6h and don't bother with bucket versioning - you're never gonna read the thing.

You only need to keep old files around, if you need to access them again in the future - you only need to do that if you roll back the catalog to an older version ("restoring a backup"). Keeping old files is **only for backup purposes**.

If you do want backups, then you need those old files in case you roll back! It becomes a question of how and when those files get deleted - either delete after 30days, or delete them now and let S3 delete them in 30 days. In either case, you only need them for backup purposes!

[^kinda]: Ignoring outstanding queries and caches, so not quite immediately, but close enough.

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,142 @@
...
+    parent: write-best-practices
+---
+
+Data ingested into InfluxDB must conform to the retention period set on the database.
+Any data which lies outside of this retention period is no longer considered for retrieval by a query, but can have references maintained in object storage or the Postgres catalog, causing an increase in operational overhead and drive higher costs.
+
+In order to reduce these factors, it is important to manage the lifecycle of ingested data. You can achieve this by following the best practices below:
+
+- Use a retention policy which is appropriate for your requirements.
+- Tune the garbage collector cutoff period to ensure that data is removed in a timely manner.
```
jdockerty, line 19: I've removed the specific recommendation of object store versioning here because it is organisation-dependent for their DR/backups strategy.

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,142 @@
...
+
+- `INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF`: the age at which a parquet file not referenced in the catalog becomes eligible for deletion from object storage. The default is `30d`.
+- `INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF`: defines how long to retain the row referencing a parquet file in the catalog once marked for deletion. The default is `30d`.
+
+Our recommendation is to keep these values aligned and you should set these to a value which matches your organisation's backup and recovery strategy.
+
+For example a value of `6h` (6 hours) would be appropriate for running a lean catalog that only maintains references to recent data.
+The assumption here is no backup requirement.
+
+Use the following scenarios as a guide for different use cases:
```
jdockerty, line 35: I've added some example scenarios and what they might mean for the variable tuning.

I'm not quite sure on the wording here for things like "leading edge data", it _sounds_ right to me but there's maybe a more common phrase that is used for things like this.

jdockerty (commented):
@domodwyer 

I've rejigged this a bit based on your feedback, do you mind taking another quick skim to ensure I'm saying that correct things here re: the write/ingest path.

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,142 @@
...
+                 INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF: '6h',
+                 INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF: '6h',
+```
+
+Using a low value here means that the garbage collector service will promptly delete files from the configured object store and remove rows from the catalog.
+Resulting in a lean catalog with minimal operational overhead.
+
+{{% /expand %}}
+
+{{% expand "Leading edge data with longer retention required" %}}
```
domodwyer, line 66: Maybe this?

```suggestion
{{% expand "Custom Backup Window - Using S3 Versioning" %}}
```

The only difference between this, and the other option below is that we utilise S3 to make sure there's files for 90 days, instead of the garbage collector.

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,142 @@
...
+
+In this scenario, we maintain a low cutoff point for the garbage collector service.
```
domodwyer, line 68: What do you think to:

```suggestion
In this scenario, we maintain a low cutoff point for the garbage collector service, and make use of an object versioning policy to ensure expired files are kept for the specified backup window time.
```

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,142 @@
...
+             iox:
+               env:
+                 INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF: '6h',
+                 INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF: '6h',
+```
+{{% /expand %}}
+
+{{% expand "Long retention" %}}
+
+If your organisation requires a longer retention period for files, you may wish to increase these values past their defaults.
```
domodwyer, line 118:
```suggestion
{{% expand "Custom Backup Window - No S3 Versioning" %}}

If your organisation cannot make use of S3 versioning policies but still requires a backup window, you must configure the garbage collector to retain the files for as long as your backup period requires.

If possible, using S3 versioning is preferred.
```

This makes it clear why you would use this over the one above I think? And a "don't do this unless you have to" pointer to get them on the right track if they can use S3 versioning.

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,142 @@
...
+
+{{% /expand %}}
+
+{{% expand "Leading edge data with longer retention required" %}}
+
+In this scenario, we maintain a low cutoff point for the garbage collector service.
+
+Object versioning is used to maintain the files for a longer duration, whilst allowing the catalog to no longer maintain their references.
+
+Noncurrent objects should be expired based on the requirements of your organisation.
```
domodwyer, line 72: We can hint to them what requirements should be considered when evaluating this:

```suggestion
Noncurrent objects should be configured to be expried as soon as possible, but retained long enough to satisfy your organisation's backup policy.
```

It'll help people who read this and go "which requirements is that?!" :D

domodwyer (approved):
Minor suggestions but nothing important - take it or leave it!

What is important and currently missing is a big, red, you-can't-miss-it banner that says:

> [!WARNING]
> Never configure the cutoff below 3h - IOx requires a grace period before files can be removed.

Maybe it's worth opening a ticket for the GC so that this can't actually be set to less than 3h to remove this footgun?

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -0,0 +1,187 @@
...
+  referenced in the Catalog become eligible for deletion from Object storage.
+  The default is `30d`.
+- `INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF`: how long to retain rows in the Catalog
+  that reference Parquet files marked for deletion. The default is `30d`.
+
+{{% warn %}}
+**These must NEVER be set to a value less than `3h` (3 hours).**
+
+IOx requires a grace period before files can be removed.
+{{% /warn %}}
```
sanderson, line 52: Where this will be/is enforced server-side, I think we can soften the language a bit.

```suggestion
{{% warn %}}
To ensure there is a grace period before files and references are removed, the
minimum garbage collector (GC) object store and Parquet file cutoff time is
three hours (`3h`).
{{% /warn %}}
```

sanderson (approved):
Thanks for this @jdockerty! I think it's good to go. Feel free to merge whenever you're ready for it to go live.

PR #5396, opened 2024-03-28
Update schema-design.md

JackieTien97:


sanderson (approved):
Thanks @JackieTien97! The column limit was raised, but this detail was missed.

PR #5403, opened 2024-03-29
fix: correct definition of `commit-timeout`

gwossum:
Correct definition of `commit-timeout` to match its use in Raft. See https://github.com/hashicorp/raft/issues/28 for more information.

### DIFF
--- a/content/enterprise_influxdb/v1/administration/configure/config-meta-nodes.md
+++ b/content/enterprise_influxdb/v1/administration/configure/config-meta-nodes.md
@@ -238,8 +238,8 @@ Environment variable: `INFLUXDB_META_LEADER_LEASE_TIMEOUT`
 
 Default is `"50ms"`.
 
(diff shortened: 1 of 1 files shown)


content/enterprise_influxdb/v1/administration/configure/config-meta-nodes.md
```diff
@@ -238,8 +238,8 @@ Environment variable: `INFLUXDB_META_LEADER_LEASE_TIMEOUT`
 
 Default is `"50ms"`.
 
-The commit timeout is the amount of time a Raft node will tolerate between
-commands before issuing a heartbeat to tell the leader it is alive.
+The commit timeout is the interval the leader waits between sending messages with
```
jstirnaman, line 241:
```suggestion
The commit timeout is the interval that the leader waits between sending messages with
```

jstirnaman (approved):
Thanks!

PR #5423, opened 2024-04-15
fix broken curl post commands

Abdullahsab3:
Closes #

The documentations uses the following command for POST requests using curl:
```
curl --post 'https://cluster-id.influxdb.io/api/v2/write?bucket=DATABASE_NAME&precision=s' \
  --header 'Authorization: Bearer DATABASE_TOKEN' \
  --data-binary 'home,room=kitchen temp=72 1463683075'
```
However this command returns an error:
```
curl: option --post: is ambiguous
curl: try 'curl --help' or 'curl --manual' for more information
```

My changes propose using the correct interface for curl, which is providing the request type using the `--request` flag

- [x] Rebased/mergeable


sanderson (approved):
Thanks @Abdullahsab3

PR #5427, opened 2024-04-15
Suggested edits to the Clustered ingest lifecycle guide

sanderson:
Rather than adding suggestions to the existing PR, it was easier to do the the proposed edits on a new branch.

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
+++ b/content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
@@ -1,31 +1,49 @@
 ---
 title: Data ingest lifecycle best practices
-seotitle: InfluxDB best practices for data ingest lifecycle management
 description: >
   Best practices for managing the lifecycle of data ingested into InfluxDB.
 menu:
   influxdb_clustered:
     name: Data ingest lifecycle
-    weight: 201
     parent: write-best-practices
+weight: 204
 ---
 
-Data ingested into InfluxDB must conform to the retention period set on the database.
-Any data which lies outside of this retention period is no longer considered for retrieval by a query, but can have references maintained in object storage or the Postgres catalog, causing an increase in operational overhead and drive higher costs.
+Data ingested into InfluxDB must conform to the retention period of the database
+in which it is stored.
+Points with timestamps outside of the retention period are no longer queryable,
+but may still have references maintained in
+[Object storage](/influxdb/clustered/reference/internals/storage-engine/#object-store)
+or the [Catalog](/influxdb/clustered/reference/internals/storage-engine/#catalog),
+resulting in an increase in operational overhead and cost.
+To reduce these factors, it is important to manage the lifecycle of ingested data.
 
-In order to reduce these factors, it is important to manage the lifecycle of ingested data. You can achieve this by following the best practices below:
+Use the following best practices to manage the lifecycle of data in your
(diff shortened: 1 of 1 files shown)


content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -1,48 +1,74 @@
...
+  referenced in the Catalog become eligible for deletion from Object storage.
+  The default is `30d`.
+- `INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF`: how long to retain rows in the Catalog
+  that reference Parquet files marked for deletion. The default is `30d`.
 
 {{% warn %}}
 **These must NEVER be set to a value less than `3h` (3 hours).**
 
 IOx requires a grace period before files can be removed.
 {{% /warn %}}
```
sanderson, line 52: Is this enforced server-side? If not, shouldn't it be?

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -52,32 +78,34 @@ metadata:
...
       components:
        garbage-collector:
          template:
            containers:
              iox:
                env:
-                 INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF: '6h',
-                 INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF: '6h',
+                 INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF: '6h'
+                 INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF: '6h'
```
sanderson, line 90: I wasn't sure about this environment variable format here. I'm fairly certain the commas are invalid yaml, but also, all the k8s documentation I can find uses the following format for environment variables:

```yml
env:
  - name: INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF
    value: '6h'
  - name: INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF
    value: '6h'
```

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
jdockerty: Yeah, this comes from Dom's suggestion [here](https://github.com/influxdata/docs-v2/pull/5369#pullrequestreview-1996750657). After that I also raised an IOx ticket for the garbage-collector and added the guard in too :+1:

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
jdockerty: Ah yeah! That's my bad :facepalm: 

So the commas are definitely incorrect, that comes from me copying these variables for a test we have internally, _but_ that is in Jsonnet so it includes commas which I forgot to strip out.

The mapping as-is here for `env` is correct though. It's part of how the Clustered operator interprets the given manifest - it follows the same pattern as our upcoming helm chart too so that they're kept aligned. We do this in a similar way to `atlantis` ([like this](https://github.com/runatlantis/helm-charts/blob/3116f468962206d14c57c84bc9a37e51d1bb3b0f/charts/atlantis/values.yaml#L475-L477)) and other products where you can provide the key-value mapping without the extra boilerplate.

PR #5435, opened 2024-04-22
docs(clustered): direct catalog modification warning

jdockerty:
I've recently spoken with the Write/Ingest team around catalog modifications and we should recommend that this **never** be done by a customer.

It can cause varying states which are unknown, most likely causing data loss and panics to occur in the various components of Clustered. It also makes it incredibly difficult for the team to provide help in diagnosing other issues if a customer reaches out for help.

- [ ] Rebased/mergeable


content/influxdb/clustered/admin/databases/delete.md
jdockerty, line 1: This is under the `influxctl database delete` command, but perhaps it should be elsewhere too?

Nowhere else came to mind though. I put it here as it came up with a customer around getting to a clean state, branching off of this command.

Happy to take suggestions.

content/influxdb/clustered/admin/databases/delete.md
```diff
@@ -40,4 +40,9 @@ Once a database is deleted, data stored in that database cannot be recovered.
 #### Cannot reuse database names
 
 After a database is deleted, you cannot reuse the same name for a new database.
-{{% /warn %}}
\ No newline at end of file
+
+#### Never directly modify the Postgres catalog
+
+Doing so will result in an undefined state for the various components.
+This may lead to data loss and crashes.
```
sanderson, line 47:
```suggestion
#### Never directly modify the Catalog

Never directly modify the [PostgreSQL-compatible Catalog](/influxdb/clustered/reference/internals/storage-engine/#catalog).
 Doing so will result in an undefined state for various components and may lead to data loss and crashes.
```

sanderson (approved):
Just one minor suggestion and it's good to go.

PR #5443, opened 2024-04-25
Add information about database token wildcard permissions

sanderson:
Users can use wildcards (`*`) to grant permissions to database tokens in InfluxDB Cloud Dedicated and InfluxDB Clustered. This PR updates token create and update documentation with information about using wildcards.

- [x] Rebased/mergeable


bcorbett6 (approved):
Just tested create and update myself :) docs are clear

jstirnaman (approved):
Looks good! I guess we need to add these for the API.

PR #5461, opened 2024-05-09
fix(flux): page not found (#5453)

### LINKED ISSUE #5453: Page not found
_Describe the issue here._

##### Relevant URLs
- https://docs.influxdata.com/flux/v0/stdlib/built-in/transformations/cumulativesum/

jstirnaman:
- Add alias for URL.
- Issue doesn't indicate where the old/incorrect URL was found.

Closes #5453

I'm not sure:
- where the reported URL was found
- whether we want this alias and if this is the place to put it
- if the problem is more widespread

### DIFF
--- a/data/flux_stdlib_frontmatter.yml
+++ b/data/flux_stdlib_frontmatter.yml
@@ -2667,6 +2667,8 @@
     - /influxdb/v2/reference/flux/functions/built-in/transformations/cumulativesum/
     - /influxdb/v2/reference/flux/stdlib/built-in/transformations/cumulativesum/
     - /influxdb/cloud/reference/flux/stdlib/built-in/transformations/cumulativesum/
+    - /flux/v0/stdlib/built-in/transformations/cumulativesum/
+
   related:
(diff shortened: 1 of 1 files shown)


content/flux/v0/stdlib/universe/cumulativesum.md
```diff
@@ -10,6 +10,8 @@ menu:
 weight: 101
 flux/v0/tags: [transformations]
 introduced: 0.7.0
+aliases:
+  - /flux/v0/stdlib/built-in/transformations/cumulativesum/
```
sanderson, line 14: Since the Flux stdlib docs are generated, we have a deploy script that injects additional "non-standard" frontmatter  into each stdlib docs. Rather than update this here, updated it `data/flux_stdlib_frontmatter.yml`. The additional frontmatter for this page starts on [line 2664](https://github.com/influxdata/docs-v2/blob/master/data/flux_stdlib_frontmatter.yml#L2664-L2672).

sanderson (approved):

PR #5479, opened 2024-05-23
Add SSO documentation to Cloud Dedicated

sanderson:
This PR adds information about setting up and using single sign-on (SSO) with InfluxDB Cloud Dedicated. It includes the following authorization flow diagram:

image

- [x] Rebased/mergeable


content/influxdb/cloud-dedicated/admin/sso.md
```diff
@@ -0,0 +1,102 @@
...
+    Auth0 and your {{< product-name >}} cluster.
+
+2.  **Provide the necessary connection credentials to InfluxData support**.
+    What credentials are needed depends on your identity provider and the
+    protocol you're using. For example:
+
+    | Protocol | Required credentials          |
+    | :------- | :---------------------------- |
+    | **OIDC** | Client secret                 |
+    | **SAML** | Identity provider certificate |
```
sanderson, line 64: @jlytle-influxdata You had mentioned that the support team has boilerplates for each of the providers/protocols. Would it be valuable to list the known details required for each protocol here? Or is it too context-specific?

content/influxdb/cloud-dedicated/admin/sso.md
```diff
@@ -0,0 +1,102 @@
...
+
+    InfluxData support will provide you with more information about what specific
+    credentials are required.
+
+3. **Add the InfluxData Auth0 connection URL as a valid callback URL** to your
+    identity provider application.
+    InfluxData support will provide you with the connection URL.
```
sanderson, line 71: @jlytle-influxdata Is this a standard callback URL or is it specific to each client? If it is standard, should we make it public in the documentation or should we hold it back for support to provide it to customers?

jstirnaman (changes requested):
Suggestions and a small typo.

jlytle-influxdata (commented):
I like this esp the idea that this is a change at the customer site that requires customer action. IDK how technical you want to get here (probably not at all) BUT when these changes happen, the customer sees a 500 error in the browser so it is not always obvious what the problem is. When certs get rotated, for example, the actual error is a mismatched hash-id of the 2 certs being compared - the error appears in the browser's address bar so the customer has the ability to identify but it is easy to miss/not know about.

content/influxdb/cloud-dedicated/admin/sso.md
jlytle-influxdata: > @jlytle-influxdata Is this a standard callback URL or is it specific to each client? If it is standard, should we make it public in the documentation or should we hold it back for support to provide it to customers?

another name for `callback URL` you will see out in the wild is `_post-back URL` - they are the same thing.

for C2, it is standard: https://auth.influxdata.com/login/callback
There is a receiving app in Auth0 called `cloud 2 - prod`, if this changes then the callback URL also could change.

content/influxdb/cloud-dedicated/admin/sso.md
jlytle-influxdata: > @jlytle-influxdata You had mentioned that the support team has boilerplates for each of the providers/protocols. Would it be valuable to list the known details required for each protocol here? Or is it too context-specific?

OIDC and SAML are far and away the most commonly used and what you have above will account for the differences. There may be other differences between IdPs that we have not encountered yet.

I may have mis-assumed that our team built the boilerplates, they may just be entrypoints that Auth0 provides - a screenshot so you know what it looks like - some of these do not have any configured connections because we haven't had a customer using them yet:
image

jlytle-influxdata: Another note on the difference between OIDC and SAML:
The certificate used for SAML is public and therefore does not require secure transmission to us.
The Client Secret used for OIDC needs to be transmitted securely, they cannot provide it in the form I supply to them to fill out. We have been using 1password shared vaults and inviting customers as guests so they can submit the client secret securely.

jlytle-influxdata (approved):

content/influxdb/cloud-dedicated/admin/sso.md
```diff
@@ -0,0 +1,160 @@
...
+
+  {{% note %}}
+  #### Keep client secrets secure
+
+  InfluxData provides a secure method for transmitting sensitive secrets such as
+  an OIDC client secret. Never send your client secret to InfluxData using an
+  insecure method.
+  {{% /note %}}
+
+- **Your using SAML and your identity provider certificate is rotated**:
```
jstirnaman, line 112:
```suggestion
- **You're using SAML and your identity provider certificate is rotated**:
```

jstirnaman (approved):
Minor fix. Thanks for doing this!

PR #5485, opened 2024-05-29
Ci lint and test fixes

jstirnaman:
- Sets the version of the Python client library to test
- Adds public/downloads as test data
- Dockerfile cleanup
- Adds Husky and lint-staged pre-commit hooks that run linting and tests (still local env only for now)
- Fixes directory-specific Vale configs and custom vocabs.


jstirnaman:
If this looks OK then I'll add documentation to CONTRIBUTING.md, or I can do that before you review it if it helps.

sanderson (approved):
👌🏻

Issue #5475, opened 2024-05-17
python v3 client library examples should use `write_options`, not `WriteOptions`


jstirnaman:
From https://github.com/influxdata/DAR/issues/427:

The documentation shows WriteOptions incorrectly
The parameter is write_options not WriteOptions, which is the object name

```
wco = write_client_options(success_callback=success,
                              error_callback=error,
                              retry_callback=retry,
                              WriteOptions=write_options)

```

## Additional links

- See fixes in https://github.com/InfluxCommunity/influxdb3-python/pull/87/

PR #5486, opened 2024-05-29
fix: reference: python v3 client library examples should use write_op…

jstirnaman:
…tions, not WriteOptions

Part of #5475


sanderson (approved):
Looks good.

PR #5502, opened 2024-06-25
docs(clustered): update ingester scaling strategy recommendation

jdockerty:
Helps with our continued [Clustered docs](https://github.com/influxdata/project-clustered/issues/443) work.

Whilst the ingesters _can_ be horizontally scaled, the usual recommendation is to actually vertically scale them first.

If we are recommending horizontal scaling by default in Clustered, this is going to cause more catalog load and extra work for the queriers. Whereas having beefier ingesters should be sought after first, as there are far more trade-offs which need to be considered when adding more ingesters.

- [ ] Rebased/mergeable


domodwyer (approved):
reverse reverse 🤦

PR #5514, opened 2024-07-09
docs(clustered): update ingress to mention grpc

jdockerty:
Clustered customers which are running on-premise need to know that components of the product utilise gRPC/HTTP2 as this may have to be configured
on other load balancers outside of their Kubernetes cluster.

We have ran into this with a customer recently, so no doubt it can happen with others.

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/install/configure-cluster.md
+++ b/content/influxdb/clustered/install/configure-cluster.md
@@ -315,10 +315,10 @@ spec:
 
 ### Set up cluster ingress
 
-{{% warn %}}
-Components of Clustered utilise gRPC/HTTP2. You may need to explicitly allow this
-on any external load balancers you use.
-{{% /warn %}}
+{{% note %}}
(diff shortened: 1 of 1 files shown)


jdockerty:
cc: @bcorbett6 as this came from a customer call

content/influxdb/clustered/install/configure-cluster.md
```diff
@@ -315,6 +315,11 @@ spec:
 
 ### Set up cluster ingress
 
+{{% warn %}}
+Components of Clustered utilise gRPC/HTTP2. You may need to explicitly allow this
+on any external load balancers you use.
+{{% /warn %}}
```
sanderson, line 321:
```suggestion
{{% note %}}
InfluxDB Clustered components use gRPC/HTTP2 protocols. If using an external load balancer,
you may need to explicitly enable these protocols on your load balancers.
{{% /note %}}
```

sanderson (approved):
Just a style suggestion and then its gtg. Thanks @jdockerty!

PR #5517, opened 2024-07-10
docs(clustered): clarify tuning garbage collector env vars

jdockerty:
A lower value here means that the garbage collector can remove files which are "newer", in the sense that a value of `6h` means the parquet files only need to be older than 6 hours and marked as `to_delete` for the GC to delete them. The opposite would be true in the case of a value of `30d`.

- [ ] Rebased/mergeable


jdockerty:
cc: @domodwyer If you don't mind, can you double check my understanding here as this is for public-facing docs :smile:

domodwyer (approved):
It's true!

content/influxdb/clustered/write-data/best-practices/data-lifecycle.md
```diff
@@ -45,6 +45,9 @@ Use the following environment variables to tune the garbage collector:
 - `INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF`: how long to retain rows in the Catalog
   that reference Parquet files marked for deletion. The default is `30d`.
 
+In short, these values tune how aggressive the garbage collector can be. A lower
+value means that files can be removed at a faster pace.
```
sanderson, line 49:
```suggestion
These values tune how aggressive the garbage collector can be. A shorter duration
value means that files can be removed at a faster pace.
```

sanderson (approved):
Just some minor style suggestions. Looks good. Thanks @jdockerty!

PR #5526, opened 2024-07-18
Scale InfluxDB Clustered

sanderson:
Adds information about scaling InfluxDB Clustered.

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/admin/scale-cluster.md
+++ b/content/influxdb/clustered/admin/scale-cluster.md
@@ -276,6 +276,44 @@ The Ingester can be scaled both [vertically](#vertical-scaling) and
 Vertical scaling increases write throughput and is typically the most effective
 scaling strategy for the Ingester.
 
+#### Ingester storage volume
+
+Ingesters use an attached storage volume to store the
+[Write-Ahead Log (WAL)](/influxdb/clustered/reference/glossary/#wal-write-ahead-log).
+With more storage available, Ingesters can keep bigger WAL buffers, which
+improves query performance and reduces pressure on the Compactor.
+Storage speed also helps with query performance.
+
+Configure the storage volume attached to Ingester pods in the
+`spec.package.spec.ingesterStorage` property of your `AppInstance` resource. 
+
+{{< expand-wrapper >}}
+{{% expand "View example Ingester storage configuration" %}}
+
+{{% code-placeholders "STORAGE_(CLASS|SIZE)" %}}
+
+```yml
+apiVersion: kubecfg.dev/v1alpha1
+kind: AppInstance
+# ...
+spec:
+  package:
+    spec:
+      # ...
+      ingesterStorage:
+        # (Optional) Set the storage class. This will differ based on the K8s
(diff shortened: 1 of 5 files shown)


jenniferplusplus (approved):
This all seems right, but I think it's also a good idea to discuss the ingester storage configuration. Unless that has its own page elsewhere?

If the ingesters have more storage available, then they can keep bigger WAL buffers, which helps with query performance and reduces pressure on the compactor. Also, obviously faster storage helps with query performance, in addition to just the quantity.

We expose storage configuration under `spec.package.spec.ingesterStorage`. That's another area where the options will depend on the underlying hardware in the cluster, and how it's configured.

content/influxdb/clustered/admin/scale-cluster.md
```diff
@@ -0,0 +1,314 @@
...
+components depend on the technology and underlying provider used for each.
+{{% /note %}}
+
+Use the `spec.package.spec.resources` property in your `AppInstance` resource
+defined in your `influxdb.yml` to define system resource minimums and limits
+for each pod and the number of replicas per component.
+`requests` are the minimum that the Kubernetes scheduler should reserve for a pod.
+`limits` are the maximum that a pod should be allowed to use.
+
+Your `AppInstance` resource can included the following properties to define
```
jstirnaman, line 86:
```suggestion
Your `AppInstance` resource can include the following properties to define
```

jstirnaman (approved):
A small fix, otherwise looks good pending other questions.

sanderson:
Thanks @jenniferplusplus. I added information about the Ingester storage. To you mind fact checking me before I push this out?

content/influxdb/clustered/reference/glossary.md
```diff
@@ -1131,15 +1131,17 @@ A statement that sets or updates the value stored in a variable.
...
+### WAL (Write-Ahead Log)
 
 The temporary cache for recently written points.
 To reduce the frequency that permanent storage files are accessed, InfluxDB
 caches new points in the WAL until their total size or age triggers a flush to
-more permanent storage. This allows for efficient batching of the writes into the TSM.
+more permanent storage. This allows for efficient batching of the writes into
+the storage engine.
 
 Points in the WAL can be queried and persist through a system reboot.
```
jstirnaman, line 1142:
```suggestion
Points in the WAL are queryable and persist through a system reboot.
```

jstirnaman (approved):
🔥

dsprogis:
Hi Scott, great job!  Two recommendations:
1. There are two places where you describe the process of applying the changes, one is highlighted in green (which, to me signals action) and the other is not.  I like the green signal and suggest it should be applied to both.
2. I think the user would benefit from knowing what to expect after applying the change - something such as, "Once the file is saved, the changes will be applied immediately.  You will see <this> happen.  If scaling up, you will begin to see <this>.  If scaling down, you will begin to see <that>.  If you do not see the changes, consult <this> documentation for troubleshooting tips."

PR #5527, opened 2024-07-22
Update the placeholder URL for dedicated

sanderson:
When we first assigned the placeholder URL for InfluxDB Cloud Dedicated, the placeholder did not include the provider-specific piece of the URL (`cluster-id.influxdb.io` vs `cluster-id.a.influxdb.io`). The `a` identifies the underlying provider as AWS. This was left out of the original placeholder URL in case other providers were added. At this point, there isn't much demand for other providers and nothing in the works to make Dedicated available on other providers, so we have been asked to update the Dedicated placeholder URL to `cluster-id.a.influxdb.io`.

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/query-data/execute-queries/client-libraries/go.md
+++ b/content/influxdb/clustered/query-data/execute-queries/client-libraries/go.md
@@ -26,7 +26,7 @@ list_code_example: |
 
     func Query() error {
         client, err := influxdb3.New(influxdb3.ClientConfig{
-            Host:       "https://cluster-id.influxdb.io",
+            Host:       "https://cluster-host.com",
             Token:      "DATABASE_TOKEN",
             Database:   "DATABASE_NAME",
         })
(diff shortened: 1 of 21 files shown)


sanderson:
I also added a migration for cookies where `cluster-id.influxdb.io` is already stored as the default.

gunnaraasen (approved):
Looks right and appears to work but I'm not an expert on how this component works.

content/influxdb/clustered/query-data/execute-queries/client-libraries/go.md
```diff
@@ -26,7 +26,7 @@ list_code_example: |
 
     func Query() error {
         client, err := influxdb3.New(influxdb3.ClientConfig{
-            Host:       "https://cluster-id.influxdb.io",
+            Host:       "https://cluster-host.com",
```
jstirnaman, line 29: Why not `"{{% influxdb/host %}}"`? I'm probably forgetting why.

jstirnaman (approved):
Looks good 👍 . One question.

content/influxdb/clustered/query-data/execute-queries/client-libraries/go.md
sanderson: Because this is in the frontmatter. Shortcodes don't work inside of frontmatter.

PR #5530, opened 2024-07-23
chore(ci): Refactor tests to Docker compose services, update Vale style, and use Lefthook for managing git hooks

jstirnaman:
- Pre-commit hook:
       - Redo test harness so commands can run in parallel
       - Cleanup Vale command args
       - Replace lint-staged and husky with Lefthook.
 - Pre-push hook:
     - Scan for package vulnerabilities
 - Tests - Run tests on demand, in parallel, or debugged using docker compose up, run, exec, etc.
     - Each test container (service) runs the prepare-content script before running tests.
     - Return to using Python instead of Docker to provide environment variables so they don't get exposed in Docker inspect.
     - Bind mount influxctl config.toml and .env.test files
     - Bind mount scripts for faster builds while iterating on test scripts.
 - Vale: 
     - Override Acronyms style def--ignore acronym if it's linked.

- [ ] Rebased/mergeable


sanderson (approved):
LGTM! Nice work!

PR #5536, opened 2024-07-25
fix(clustered): adjust keycloak AppInstance config example

waynr:
This came to my attention during a customer call earlier today where we discovered that these fields are deserialized into a struct at runtime and can cause a critical service to crashloop if not present.

### DIFF
--- a/content/influxdb/clustered/install/configure-cluster.md
+++ b/content/influxdb/clustered/install/configure-cluster.md
@@ -634,9 +634,8 @@ spec:
         jwksEndpoint: |-
           https://KEYCLOAK_HOST/auth/realms/KEYCLOAK_REALM/protocol/openid-connect/certs
         users:
-          # All fields are required but the firstName, lastName, and email can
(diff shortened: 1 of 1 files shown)


content/influxdb/clustered/install/configure-cluster.md
```diff
@@ -634,7 +634,13 @@ spec:
         jwksEndpoint: |-
           https://KEYCLOAK_HOST/auth/realms/KEYCLOAK_REALM/protocol/openid-connect/certs
         users:
-          - KEYCLOAK_USER_ID
+          # All fields are required but the firstName, lastName, and email can
+          # be arbitrary values -- only the `id` needs to match up with the
+          # identity provider user management system.
```
sanderson, line 639:
```suggestion
          # All fields are required but `firstName`, `lastName`, and `email` can be
          # arbitrary values. However, `id` must match the user ID provided by Keycloak.
```

sanderson (approved):
One suggestion on the added comment. Other than that, it looks great! Thanks @waynr!

PR #5544, opened 2024-07-30
Replace startsWith with startswith method

abalone23:
The startswith method is not capitalized:
https://docs.python.org/3/library/stdtypes.html#str.startswith

Closes #

- [ ] Rebased/mergeable


jstirnaman:
Thanks for this! You're right, but it passes the test, so obviously the callback functions aren't being called--not sure how I broke that.

jstirnaman (approved):

PR #5545, opened 2024-07-30
fix(v2): http-write-timeout

### LINKED ISSUE #5513: http-write-timeout documentation is wrong
[http-write-timeout](https://docs.influxdata.com/influxdb/v2/reference/config-options/#http-write-timeout)

This parameter is actually not related to write requests, but to the time permitted to write a response to an HTTP request.

It is used to set [this timeout](https://pkg.go.dev/net/http#Server.WriteTimeout) in the Go net/http.Server, whose documentation is close to what ours should be.

See this [issue](https://github.com/influxdata/influxdb/issues/25029)

jstirnaman:
Closes #5513 

- The substantive change fixes the [http-write-timeout definition](https://github.com/influxdata/docs-v2/pull/5545/files#diff-a2139f0fdd416bf569f905563f3a6bc13d50fd602757d959de8c21cf62c32d10R800), which should be similar to https://pkg.go.dev/net/http#Server.WriteTimeout

Related testing and linting changes:
- Add a Vale spelling config for v2 server options
- Add influxdb to the test container packages (for testing service influxdb...)
- Add Dockerfile config and test setup for testing some InfluxDB startup config options (using influxd)


content/influxdb/v2/reference/config-options.md
```diff
@@ -202,10 +219,24 @@ _Typically, InfluxData internal use only._
...
 
 ###### influxd flag
+
+<!--pytest.mark.skip-->
+
 ```sh
 influxd --assets-path=/path/to/custom/assets-dir
 ```
 
+<!--test-actual
```
jstirnaman, line 229:
```suggestion
<!--test:actual
```

content/influxdb/v2/reference/config-options.md
```diff
@@ -34,6 +34,23 @@ Server configuration commands require an [Operator token](/influxdb/v2/admin/tok
 Use the [`influx server-config` command](/influxdb/v2/reference/cli/influx/server-config/)
 to retrieve your runtime server configuration.
 
+<!--test: setup
```
jstirnaman, line 37:
```suggestion
<!--test:setup
```

jstirnaman:
@davidby-influx The substantive change fixes the [http-write-timeout definition](https://github.com/influxdata/docs-v2/pull/5545/files#diff-a2139f0fdd416bf569f905563f3a6bc13d50fd602757d959de8c21cf62c32d10R800) to be similar to https://pkg.go.dev/net/http#Server.WriteTimeout

davidby-influx (approved):
LGTM

davidby-influx:
This change should be for 1.X, as well.

jstirnaman:
> This change should be for 1.X, as well.

In v1, is it [`enqueued-write-timeout`](https://docs.influxdata.com/influxdb/v1/administration/config/#enqueued-write-timeout--0) or [`write-timeout`](https://docs.influxdata.com/influxdb/v1/administration/config/#write-timeout--10s)?
The [migration table](https://docs.influxdata.com/influxdb/v2/install/upgrade/v1-to-v2/manual-upgrade/#view-configuration-option-parity) says `enqueued-write-timeout`. Is that correct?

davidby-influx:
On looking more closely, it is not present in 1.X, and the given equivalents do something different.  In 2.X it is passed to the Go net/http package as a server setting; it 1.X those other two options are handled by Influx code and do slightly different things.  So I was wrong it is in 1.x (we don't change the defaults for net/http Server write timeouts.  

Let's not put it in 1.X, and I'll look into the equivalency table in more depth later....

PR #5556, opened 2024-08-07
Fix test product environments

jstirnaman:
Resolves important test environment stuff I broke when migrating over to Compose:
- Loads data/product.yml as environment variables.
- Sets the correct CONTENT_PATH value for each product test container.

- [ ] Rebased/mergeable


sanderson (approved):
👍🏻

PR #5571, opened 2024-08-19
docs: LDAPS and LDAP Client Certificate Docs

gwossum:
Add documentation via LDAP sample configuration for LDAPS and LDAP client certificates. These are new features in the yet-to-be released Enterprise 1.11.7


sanderson (approved):
Thanks @gwossum!

PR #5573, opened 2024-08-20
5568 clustered docs need to be updated claim hard limit of 250 columns per database

### LINKED ISSUE #5568: Clustered: Docs need to be updated - claim hard limit of 250 columns per database.
_Describe the issue here._

Please clarify this portion of the docs. This seems to be in conflict with the ability to set maximum columns per database using the `influxctl`. 

https://docs.influxdata.com/influxdb/clustered/write-data/best-practices/schema-design/#tables-can-contain-up-to-250-columns 

##### Relevant URLs
- https://docs.influxdata.com/influxdb/clustered/write-data/best-practices/schema-design/
- _Provide other relevant URLs (project issues, community threads, etc.)_

jstirnaman:
Closes #5568

- Update column limit (and default) for Cloud Dedicated and Clustered
- Provide more detailed guidance for the impact of wide schemas and non-specific (broad) queries
- DRY up the column limit value in case it changes again.

### DIFF
--- a/content/influxdb/cloud-dedicated/admin/databases/_index.md
+++ b/content/influxdb/cloud-dedicated/admin/databases/_index.md
@@ -11,6 +11,14 @@ menu:
     parent: Administer InfluxDB Cloud
 weight: 101
 influxdb/cloud-dedicated/tags: [databases]
+related:
+  - /influxdb/cloud-dedicated/write-data/best-practices/schema-design/
+  - /influxdb/cloud-dedicated/reference/cli/influxctl/
+alt_links:
+  cloud: /influxdb/cloud/admin/buckets/
+  cloud_serverless: /influxdb/cloud-serverless/admin/buckets/
+  clustered: /influxdb/clustered/admin/databases/
+  oss: /influxdb/v2/admin/buckets/
 ---
 
 An InfluxDB database is a named location where time series data is stored.
@@ -19,11 +27,13 @@ Each InfluxDB database has a [retention period](#retention-periods).
 {{% note %}}
 **If coming from InfluxDB v1**, the concepts of databases and retention policies
 have been combined into a single concept--database. Retention policies are no
-longer part of the InfluxDB data model. However, InfluxDB Cloud Dedicated does
+longer part of the InfluxDB data model.
+However, {{% product-name %}} does
 support InfluxQL, which requires databases and retention policies.
 See [InfluxQL DBRP naming convention](/influxdb/cloud-dedicated/admin/databases/create/#influxql-dbrp-naming-convention).
 
-**If coming from InfluxDB v2 or InfluxDB Cloud**, _database_ and _bucket_ are synonymous.
+**If coming from InfluxDB v2, InfluxDB Cloud (TSM), or InfluxDB Cloud Serverless**,
+_database_ and _bucket_ are synonymous.
 {{% /note %}}
 
 ## Retention periods
@@ -40,9 +50,10 @@ never be removed by the retention enforcement service.
 
 ## Table and column limits
 
-In {{< product-name >}}, table (measurement) and column limits can be
-customized when [creating](#create-a-database) or
-[updating a database](#update-a-database).
+You can customize [table (measurement) limits](#table-limit) and
+[table column limits](#column-limit) when you
+[create](#create-a-database) or
+[update a database](#update-a-database) in {{< product-name >}}.
 
 ### Table limit
 
@@ -72,7 +83,7 @@ data by measurement and time range and stores each partition as a Parquet
 file in your cluster's object store. By increasing the number of measurements
 (tables) you can store in your database, you also increase the potential for
 more `PUT` requests into your object store as InfluxDB creates more partitions.
-Each `PUT` request incurs a monetary cost and will increase the operating cost of
+Each `PUT` request incurs a monetary cost and increases the operating cost of
 your cluster.
 
 {{% /expand %}}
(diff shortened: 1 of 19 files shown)


content/influxdb/clustered/query-data/troubleshoot-and-optimize/optimize-queries.md
```diff
@@ -53,9 +52,39 @@ Some bottlenecks may be out of your control and are the result of a suboptimal e
...
 
+### Query only the data you need
+
+#### Include a WHERE clause
+
+InfluxDB v3 stores data in a Parquet file for each measurement and day, and
+retrieves files from the Object store to answer a query.
+To reduce the number of files that a query needs to retrieve from the Object store,
+include a [`WHERE` clause](/influxdb/clustered/reference/sql/where/) that
+filters data by a time range.
```
sanderson, line 68: In Clustered, users can custom-partition the data.

```suggestion
InfluxDB v3 stores data in a Parquet file for each partition.
By default, {{< product-name >}} partitions tables by day, but you can also
[custom-partition your data](/influxdb/clustered/admin/custom-partitions/).
At query time, InfluxDB retrieves files from the Object store to answer a query.
To reduce the number of files that a query needs to retrieve from the Object store,
include a [`WHERE` clause](/influxdb/clustered/reference/sql/where/) that
filters data by a time range or by specific tag values.
```

content/influxdb/cloud-dedicated/query-data/troubleshoot-and-optimize/optimize-queries.md
```diff
@@ -52,9 +52,39 @@ Some bottlenecks may be out of your control and are the result of a suboptimal e
...
 
+### Query only the data you need
+
+#### Include a WHERE clause
+
+InfluxDB v3 stores data in a Parquet file for each measurement and day, and
+retrieves files from the Object store to answer a query.
+To reduce the number of files that a query needs to retrieve from the Object store,
+include a [`WHERE` clause](/influxdb/cloud-dedicated/reference/sql/where/) that
+filters data by a time range.
```
sanderson, line 68: In Dedicated, users can custom-partition data.

```suggestion
InfluxDB v3 stores data in a Parquet file for each partition.
By default, {{< product-name >}} partitions tables by day, but you can also
[custom-partition your data](/influxdb/cloud-dedicated/admin/custom-partitions/).
At query time, InfluxDB retrieves files from the Object store to answer a query.
To reduce the number of files that a query needs to retrieve from the Object store,
include a [`WHERE` clause](/influxdb/cloud-dedicated/reference/sql/where/) that
filters data by a time range or by specific tag values.
```

content/influxdb/clustered/write-data/best-practices/schema-design.md
```diff
@@ -106,26 +116,37 @@ cardinality doesn't affect the overall performance of your database.
...
 
-### Tables can contain up to 250 columns
+### Maximum number of columns per table
+
+A table has a [maximum number of columns](/influxdb/clustered/admin/databases/#column-limit).
+Each row must include a time column.
+As a result, a table can have the following:
+
+- a time column
+- field and tag columns up to the configured maximum.
```
sanderson, line 132:
```suggestion
- field and tag columns up to the configured maximum
```

sanderson (changes requested):
Just some minor suggestions related custom-partitioning in Dedicated and Clustered.

helenosheaa (approved):
lgtm, i'll also mention it to the team

content/influxdb/cloud-dedicated/admin/databases/_index.md
```diff
@@ -40,9 +50,10 @@ never be removed by the retention enforcement service.
...
 
-In {{< product-name >}}, table (measurement) and column limits can be
-customized when [creating](#create-a-database) or
-[updating a database](#update-a-database).
+You can customize [table (measurement) limits](#table-limit) and
+[table column limits](#column-limit) when you
+[create](#create-a-database) or
+[update a database](#update-a-database) in {{< product-name >}}.
 
 ### Table limit
```
jstirnaman, line 58: @savage-engineer Would you mind reviewing the "Limits" sections here?

content/influxdb/cloud-dedicated/admin/databases/_index.md
```diff
@@ -89,22 +100,33 @@ operating cost of your cluster.
 
 ### Column limit
 
-**Default maximum number of columns**: 250
+**Default maximum number of columns**: 1000
```
domodwyer, line 103: I believe the default remains at 250 - [this work](https://github.com/influxdata/influxdb_iox/issues/11156) was done to prevent users further raising their default column limit to the point it became problematic for the querier.

My understanding is:

* By default, a column limit of 250 is imposed per table (the "soft limit").
* This limit can be raised (via a gRPC call, I'm not sure by who - support?) to up to 1000.
* Existing tables that had >1000 columns at the time the hard limit was introduced continue to work, but their column limit cannot be raised further.

Really we do not want users hitting 1000 columns, but we needed to keep some flexibility - I believe performance starts to degrade above the default 250 soft limit (or at least that is why it was chosen at the time).

domodwyer (commented):
@jstirnaman Fraser is off for the week 🙏

content/influxdb/cloud-dedicated/admin/databases/_index.md
dsprogis: I believe some customers would trade performance for width and perhaps others would not.  Ideally, they would like to know more about the tradeoff ... ie how much performance they will trade as they push the limit and I doubt we can quantify it for them.  But, perhaps there is a middle ground in which we give the customer the ability to move the limit but warn them of a likely decay in performance and recommend that they watch for the down-stream impact.  Are there any metrics in the Observability Dashboard that we could point to as an indicator that they might be widening their table too much?

domodwyer: It's not a grandual-perf-trade-off thing above ~1,000 columns; the queriers and (especially) the catalog service begin to fail outright.

Engineering took the decision to cap the columns after a number of serious and time-consuming incidents in caused by multiple customers just yolo'ing the limit as high as they wanted :(

dsprogis: Got it, thanks.
So what is the verdict?  I was on a call yesterday with VAST and got their hopes up regarding 1000 columns.  Shall I return to them with the news that the limit remains 250?  Is there a longer view on this that I can share, possibly an effort to re-architecture or move a bottleneck somewhere in the data pipeline?

domodwyer: The default limit is 250, but it can be raised to 1000 so they should be 👌 

We prevent the limit from being raised above 1,000 though :+1:

I'm not sure if it is an area being actively worked on or planned for.

jstirnaman: > I believe the default remains at 250 - [this work](https://github.com/influxdata/influxdb_iox/issues/11156) was done to prevent users further raising their default column limit to the point it became problematic for the querier.

@domodwyer ❤️ Thanks for the clarification!

jstirnaman (commented):
@dsprogis @sanderson I added a fix for Dom's clarification.

content/influxdb/cloud-dedicated/admin/databases/_index.md
```diff
@@ -91,20 +103,30 @@ operating cost of your cluster.
...
-The more columns that are in each row, the longer it takes to evaluate each row.
-
-Through performance testing, InfluxData has identified 250 columns as the
-threshold where query performance may be affected
-(depending on the shape of and data types in your schema).
+{{% expand "May adversely affect system performance" %}}
+
+InfluxData identified 250 columns as the safe limit for maintaining system
+performance and stability.
+Exceeding this threshold can result in
```
sanderson, line 126: Do we want to mention that 250 is a soft limit and users can raise the column limit to 1000 (a hard limit), but we don't recommend it?

content/influxdb/cloud-dedicated/write-data/best-practices/schema-design.md
```diff
@@ -106,26 +116,38 @@ cardinality doesn't affect the overall performance of your database.
...
+If you attempt to write to a table and exceed the column limit, then the write
+request fails and InfluxDB returns an error.
 
-A table can contain **up to 250 columns**. Each row requires a time column,
-but the rest represent tags and fields stored in the table.
-Therefore, a table can contain one time column and 249 total field and tag columns.
-If you attempt to write to a table and exceed the 250 column limit, the
-write request fails and InfluxDB returns an error.
+InfluxData identified the default maximum as the safe limit for maintaining system
+performance and stability.
```
sanderson, line 138: I think we should state or link to what the limit is here.

PR #5579, opened 2024-08-28
docs(clustered): how to set environment variables

jdockerty:
This explains and provides examples on how environment variables can be set within the Clustered `AppInstance` configuration.

>[!IMPORTANT]
>It **does not** provide an exhaustive list of environment variables that can be tweaked, as this may change based upon many factors.
>
>The general expectation is that customers tweak or set certain variables when instructed via Support, rather than on their own as there are a lot of moving parts to the database and as such there may be tradeoffs to consider as well.

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/admin/env-vars.md
+++ b/content/influxdb/clustered/admin/env-vars.md
@@ -1,0 +1,98 @@
+---
+title: Manage environment variables for Clustered components
+description: >
+  Manage environment variables for Clustered components.
+menu:
+  influxdb_clustered:
+    parent: Administer InfluxDB Clustered
+    name: Manage environment variables
+weight: 101
+influxdb/clustered/tags: []
+---
+
+There are numerous environment variables available to the components of Clustered.
+This is handled entirely through your `AppInstance` configuration.
+
+Environment variables enable you to tweak the running configuration of Clustered.
+Many of them are already set for you, but you may wish to tune them specifically,
+either by our request or for your own means. 
+
+An example of tuning environment variables for the `garbage-collector` component
+is provided below:
+
+```yaml
+apiVersion: kubecfg.dev/v1alpha1
+kind: AppInstance
+metadata:
+  name: influxdb
+  namespace: influxdb
+spec:
+  package:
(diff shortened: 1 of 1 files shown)


content/influxdb/clustered/admin/env-vars.md
jdockerty, line 1: There may be a better place to put this and perhaps I can also include the instruction on not really tweaking this unless instructed to do so by Support.

I'm not quite sure we want to surface this, happy to take suggestions :+1:

sanderson:
Thanks for this @jdockerty. I'm going to pull this information into another PR for partial writes in InfluxDB Clustered.

> I can also include the instruction on not really tweaking this unless instructed to do so by Support.

How dependent are our Clustered users on our Support team? Is there a contractual relationship between a Clustered customer and our Support team? Is there ever a situation where a Clustered customer might be "silo'd" and not have access to our support team? I guess I'm wondering if we should provide the information in the documentation with clear warnings that they could mess things up if they start tweaking things on their own.

jdockerty:
@sanderson 

> How dependent are our Clustered users on our Support team?

At the moment I think there's more of a reliance when something out of the norm happens or scaling snags are hit. There's no reason why that couldn't be self-served though, although the number of tunables in IOx is rather overwhelming and variables may be taken out of action depending on various factors - I think tracking that in Clustered docs could become rather strenuous on either team that handles it.

> Clustered customer might be "silo'd" and not have access to our support team?

I don't believe so. By virtue of what Clustered is as a product means that it is for enterprise customers, so there's a level of support that is required on our side there (you might argue even expected).

I think the main complication comes from the fact that there are so many knobs to change, but documenting everything is also prone to bitrot as well. For reference, from a rudimentary search over the env vars I've counted 136 instances of unique `INFLUXDB_IOX_$SOMETHING` variables. Perhaps there _is_ a way to properly document this stuff is there is an agreement on how environment variables should be handled; otherwise a customer may have something pulled from underneath them from different releases and behaviour may change/performance could degrade.

PR #5586, opened 2024-09-03
Update OSS verification and install instructions

### LINKED ISSUE #5559: Review and update OSS install instructions
- Some examples using curl in gpg verification don't work as expected in Docker.
- Fix tests
- Make influxdata_archive.key be the default and describe _compat.key as a fallback for "very old distros"
- Include SHA checksum verification

jstirnaman:
Closes #5559
- Updates Linux install instructions to reflect updates in the Downloads page.
- Uses curl, instead of wget, for examples.
- Reorg of download verification: moves expandable details before instructions, to encourage using verification.
  Includes verification in instructions.
- All tests pass if not explicitly skipped (for Docker and different architectures).
- Run `docker compose build` to update the Go version and other dependencies used by the
  test image.


sanderson (approved):
🚀

PR #5588, opened 2024-09-04
chore(influxql): Test and improve InfluxQL functions with regex argum…

jstirnaman:
…ents

- Part of investigation into https://github.com/influxdata/DAR/issues/434
- Tested using regex function arguments in InfluxQL functions in InfluxDB v1 and v2.
- Add test setup and a test for v2
- Add examples for v2
- Update lint and test configs

- [x] Rebased/mergeable


static/downloads/home-sensor-data.csv
```diff
@@ -1,3 +1,4 @@
+#datatype measurement,tag,field,field,field,time
```
sanderson, line 1: It looks like this CSV file is used in a few Python client library examples. Just want to make sure the added annotation doesn't break anything.

sanderson (approved):

PR #5594, opened 2024-09-06
InfluxDB Clustered partial writes

sanderson:
This lays the groundwork for partial writes in InfluxDB Clustered. Before this PR goes live, we will need to add information about disabling the partial writes feature using the available environment variable. This will like go in the release notes of the next Clustered release and point to the environment variable configuration docs included in this PR.

> [!IMPORTANT]
> Do no merge until the next Clustered release (planned for Sept 16th)

- Updates Clustered API docs with new partial writes behavior
- Adds information about configuring components with environment variables
- Updates write task-based docs with information about new partial writes status codes

---

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/admin/env-vars.md
+++ b/content/influxdb/clustered/admin/env-vars.md
@@ -1,24 +1,44 @@
 ---
-title: Manage environment variables for Clustered components
+title: Manage environment variables in your InfluxDB Cluster
 description: >
-  Manage environment variables for Clustered components.
+  Use environment variables to define settings for individual components in your
+  InfluxDB cluster.
 menu:
   influxdb_clustered:
     parent: Administer InfluxDB Clustered
     name: Manage environment variables
-weight: 101
-influxdb/clustered/tags: []
+weight: 208
 ---
 
-There are numerous environment variables available to the components of Clustered.
-This is handled entirely through your `AppInstance` configuration.
+Use environment variables to define settings for individual components in your
+InfluxDB cluster and adjust your cluster's running configuration.
+Define environment variables for each component in your `AppInstance` resource.
 
-Environment variables enable you to tweak the running configuration of Clustered.
-Many of them are already set for you, but you may wish to tune them specifically,
-either by our request or for your own means. 
+There are a number of environment variables available to InfluxDB Clustered components.
+Many environment variables have default settings, but you can override these
+defaults by setting custom values for environment variables.
 
-An example of tuning environment variables for the `garbage-collector` component
-is provided below:
+{{% warn %}}
+#### Overriding default settings may affect overall cluster performance
+
+{{% product-name %}} components have complex interactions with each other that
+can be affected when overriding default configuration settings.
+Before using environment variables to make configuration changes, you should
+consider consulting [InfluxData Support](https://support.influxdata.com/) to
+identify any potential unintended consequences of configuration changes.
+{{% /warn %}}
+
+## AppInstance component schema
+
+In your `AppInstance` resource, configure individual component settings in the
+`spec.package.spec.components` property. This property supports the following
+InfluxDB Clustered component keys:
+
+- `ingester`
+- `querier`
+- `router`
+- `compactor`
+- `garbage-collector`
 
 ```yaml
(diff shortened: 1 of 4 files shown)


content/influxdb/clustered/admin/env-vars.md
```diff
@@ -0,0 +1,160 @@
...
+      --filename myinfluxdb.yml \
+      --namespace influxdb
+    ```
+{{% note %}}
+#### Update instead of removing environment variables
+
+Removing environment variables from your `AppInstance` resource configuration
+will not revert those environment variables to their default setting.
+To revert back to the default settings, update the value in your `AppInstance`
+resource to the default value rather than removing the environment variable definition.
```
jdockerty, line 110: This is not quite true.

Removing a variable won't "remove it" in the sense of not setting anything. If there is a default then it will go back to that.

E.g. the cutoff variables from above (and other variables that get changed) are really _overrides_ for most things. The default there being `30d`, but altered in the example to `6h`.

content/influxdb/clustered/admin/env-vars.md
jdockerty, line 1: This looks very nice now, thank you :100:

content/influxdb/clustered/admin/env-vars.md
jstirnaman, line 110:
```suggestion
Most configuration settings that can be overridden by environment variables have default values that are used if the environment variable is unset.
Removing environment variables from your `AppInstance` resource configuration will not remove those environment variables entirely; instead, they will revert to their default settings.
To revert to the default settings, simply unset the environment variable or update the value in your `AppInstance` resource to the default value.

In the preceding example, the `INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF` environment variable is set to `6h`.
If you remove `INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF` from the `env` property, the cutoff reverts to its default setting of `30d`.
```
@jdockerty Is this correct?

jdockerty: Exactly that :100:

content/influxdb/clustered/admin/env-vars.md
```diff
@@ -0,0 +1,158 @@
...
+3.  Use `kubectl apply` to apply the configuration changes to your cluster and
+    add or update environment variables in each component.
+
+    ```sh
+    kubectl apply \
+      --filename myinfluxdb.yml \
+      --namespace influxdb
+    ```
+{{% note %}}
+#### Update environment variables instead of removing them
```
sanderson, line 102:
```suggestion
#### Update environment variables instead of removing them

```

content/influxdb/clustered/admin/env-vars.md
```diff
@@ -0,0 +1,159 @@
...
+                iox:
+                  env:
+                    INFLUXDB_IOX_GC_OBJECTSTORE_CUTOFF: '6h'
+                    INFLUXDB_IOX_GC_PARQUETFILE_CUTOFF: '6h'
+    ```
+
+3.  Use `kubectl apply` to apply the configuration changes to your cluster and
+    add or update environment variables in each component.
+
+    ```sh
```
sanderson, line 96:
~~~suggestion
    <!-- pytest.mark.skip -->

    ```bash
~~~

waynr:
@sanderson we are getting ready to release a new version of Clustered with partial writes enabled. Is that all this PR was waiting on?

sanderson:
@waynr I think so, yes. That and the release notes, but those will come after the release. @dsprogis mentioned that requiring the license will also be the new default in this release. Is that the case?

waynr:
@sanderson just FYI, I have referenced these new docs in the release notes: https://github.com/influxdata/k8s-idpe/pull/48148/files#diff-de6ce98acc9bb2452a84f41b7d74980ad01f561d3fb6a18441dc4873bd558954R101

> the license will also be the new default in this release. Is that the case?

Yes, that's also mentioned in the release notes.

waynr:
@sanderson just FYI, the new Clustered release went out the other day so this should be good to go!

sanderson:
Perfect! Thank you @waynr.

PR #5596, opened 2024-09-06
chore: remove unnecessary ::timestamp casts

jeffreyssmith2nd:
@sanderson This removes anywhere that we had `::timestamp` where the value will already be coerced to a timestamp. This is important because the `::timestamp` cast currently (and may always) removes the timezone information from the timestamp. PR'd into the v3/timezone-support branch.

- [ ] Rebased/mergeable


sanderson (approved):
Thanks for this @jeffreyssmith2nd!

PR #5602, opened 2024-09-19
fix(v3): Add optimizations to system table queries #5576

jstirnaman:
- Closes #5576
- Integrates guidance from https://github.com/influxdata/influxdb_iox/issues/11974 for Cloud Dedicated and Clustered.
- Updates examples to include filter conditions.

PR #5611, opened 2024-09-24
fix(v3): add filters on more system table examples

appletreeisyellow:
Follow-up of https://github.com/influxdata/docs-v2/pull/5602

Relate to https://github.com/influxdata/influxdb_iox/issues/11974

I updated all the system table examples that's missing filters

### DIFF
--- a/content/influxdb/cloud-dedicated/admin/custom-partitions/view-partitions.md
+++ b/content/influxdb/cloud-dedicated/admin/custom-partitions/view-partitions.md
@@ -35,7 +35,7 @@ your {{< product-name omit=" Clustered" >}} cluster.
 #### System tables are subject to change
 
 System tables are not part of InfluxDB's stable API and may change with new releases.
-The provided schema information and query examples are valid as of **August 22, 2024**.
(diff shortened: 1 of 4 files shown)


content/influxdb/cloud-dedicated/admin/custom-partitions/view-partitions.md
```diff
@@ -108,40 +108,26 @@ with the name of the table you want to query information about.
...
 ---
 
-{{% code-placeholders "TABLE_NAME" %}}
+{{% code-placeholders "TABLE_NAME_(1|2|3)|TABLE_NAME" %}}
 
-### View partition templates of all tables
+### View the partition template of a specific table
 
 ```sql
-SELECT * FROM system.tables
```
appletreeisyellow, line 116: Removing the example for all the tables. Even though `SELECT * FROM system.tables` works fine, it can be expensive, so I don't want to encourage users to so it

jstirnaman (approved):
🚀 Looks good - thanks for doing this!

PR #5612, opened 2024-09-24
fix(clustered): update admin token revocation steps

waynr:
While validating this process for a customer support issue, I found that it wasn't enough merely to delete the `rsa-keys` secret and the `key-gen-NNNNNNNN` job.

In order for the admin token that I had initially retrieved from my cluster to stop working, I also had to delete the `admin-token` secret and the corresponding `create-admin-token-NNNNNNN` job to get a newly-generated admin token and cause the old admin token to be invalidated.


sanderson (approved):
Thanks @waynr!

PR #5616, opened 2024-09-24
fix(oss): Update TLS example using openssl and including Subject Alte…

### LINKED ISSUE #5608: Update OpenSSL command for generating self-signed certificate
Following the instructions in the "Enable TLS" page and creating a self-signed certificate, the certificate created using OpenSSL has only the Common Name (CN) but not the subjectAltName (SAN).

The currently suggested command is here:
https://github.com/influxdata/docs-v2/blob/a29cb2a2f35efe53525762d48c49022d3496f438/content/influxdb/v2/admin/security/enable-tls.md?plain=1#L82-L85

The lack of subjectAltName causes this warning following the Python client tutorial:
```
/usr/lib/python3/dist-packages/urllib3/connection.py:455: SubjectAltNameWarning: Certificate for vitsolc1 has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/urllib3/urllib3/issues/497 for details.)
  warnings.warn(
```

And this error when trying to configure a new source in Grafana:
`
Get "https://myhost:8086/query?db=test&epoch=ms&q=SHOW+measurements": tls: failed to verify certificate: x509: certificate relies on legacy Common Name field, use SANs instead error reading influxDB`

An easy solution is to add a `-addext` field to the OpenSSL command, like:
` -addext "subjectAltName = DNS:myhost.mydomain.com,DNS:10.29.0.15"`

I found the `-addext` solution here: https://security.stackexchange.com/a/183973/273481

##### Relevant URLs
- https://docs.influxdata.com/influxdb/v2/admin/security/enable-tls/#configure-influxdb-to-use-tls

jstirnaman:
…rnative Name (SAN)

Closes #5608 
- Example uses Subject Alternative Name extension required by modern clients.
- Updated example is more verbose, but should work cross-platform.
- Added troubleshooting steps.
- Passes tests.
- Reformatted to headings and remove list nesting.

### DIFF
--- a/content/influxdb/v2/admin/security/enable-tls.md
+++ b/content/influxdb/v2/admin/security/enable-tls.md
@@ -23,7 +23,8 @@ To set up TLS over HTTPS, do the following:
   - [Self-signed certificates](#self-signed-certificates)
 - [Configure InfluxDB to use TLS](#configure-influxdb-to-use-tls)
 - [Connect Telegraf to a secured InfluxDB instance](#connect-telegraf-to-a-secured-influxdb-instance)
-  - [Example configuration](#example-configuration)
+  - [Example Telegraf configuration](#example-telegraf-configuration)
+- [Troubleshoot TLS](#troubleshoot-tls)
 
 {{% warn %}}
 InfluxData **strongly recommends** enabling HTTPS, especially if you plan on sending requests to InfluxDB over a network.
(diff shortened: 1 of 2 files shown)


content/influxdb/v2/admin/security/enable-tls.md
```diff
@@ -176,3 +255,56 @@ update the following `influxdb_v2` output settings in your Telegraf configuratio
 ```
 
 Restart Telegraf using the updated configuration file.
+
+## Troubleshoot TLS
+
+### 1. Check InfluxDB logs
```
sanderson, line 261: Is there a specific order to these troubleshooting steps? I've generally moved away from including these types of ordered numbers in headings, but I won't block the PR for these.

sanderson (approved):
Looks good. Just one question about the ordered numbers in the troubleshooting headings.

content/influxdb/v2/admin/security/enable-tls.md
jstirnaman: > I've generally moved away from including these types of ordered numbers in headings,

I've noticed and so I try to avoid them as well. I don't think there's a necessary order in this case; I'll remove them.

PR #5620, opened 2024-09-26
fix(v1): update client library recommendations for v1, v3

jstirnaman:
- Updates recommended client libraries to use with v1
- Adds v3 compat information

### DIFF
--- a/layouts/partials/article/stable-version.html
+++ b/layouts/partials/article/stable-version.html
@@ -1,6 +1,6 @@
 {{ $productPathData := findRE "[^/]+.*?" .RelPermalink }}
 {{ $product := index $productPathData 0 }}
-{{ $productName := (index .Site.Data.products $product).name }}
+{{ $productName := cond (isset (index .Site.Data.products $product) "altname") (index .Site.Data.products $product).altname (index .Site.Data.products $product).name }}
 {{ $currentVersion := index $productPathData 1 }}
 {{ $latestV2 := index (.Site.Data.products.influxdb.versions) 0 }}
 
--- a/content/influxdb/cloud-dedicated/reference/client-libraries/v2/_index.md
+++ b/content/influxdb/cloud-dedicated/reference/client-libraries/v2/_index.md
@@ -1,7 +1,7 @@
 ---
 title: InfluxDB v2 API client libraries
 description: >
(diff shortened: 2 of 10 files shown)


jstirnaman:
@sanderson Should we update this warning to say v3 or make it specific to OSS?

image

sanderson:
I think we should probably make it OSS specific. I think if you update line 3 in `layouts/partials/article/stable-version.html` with the following, it should make it so it says "InfluxDB OSS v2".

```html
{{ $productName := cond (isset (index .Site.Data.products $product) "altname") (index .Site.Data.products $product).altname (index .Site.Data.products $product).name }}
```

jstirnaman:
> I think we should probably make it OSS specific. I think if you update line 3 in `layouts/partials/article/stable-version.html` with the following, it should make it so it says "InfluxDB OSS v2".
> 
> ```
> {{ $productName := cond (isset (index .Site.Data.products $product) "altname") (index .Site.Data.products $product).altname (index .Site.Data.products $product).name }}
> ```

Thanks! I looked and was afraid I'd insert it somewhere else without knowing it.

jstirnaman:
> > I think we should probably make it OSS specific. I think if you update line 3 in `layouts/partials/article/stable-version.html` with the following, it should make it so it says "InfluxDB OSS v2".
> > ```
> > {{ $productName := cond (isset (index .Site.Data.products $product) "altname") (index .Site.Data.products $product).altname (index .Site.Data.products $product).name }}
> > ```
> 
> Thanks! I looked and was afraid I'd insert it somewhere else without knowing it.

✅ Can you review this?

layouts/partials/article/stable-version.html
```diff
@@ -1,6 +1,6 @@
 {{ $productPathData := findRE "[^/]+.*?" .RelPermalink }}
 {{ $product := index $productPathData 0 }}
-{{ $productName := (index .Site.Data.products $product).name }}
+{{ $productName := cond (isset (index .Site.Data.products $product) "altname") (index .Site.Data.products $product).altname (index .Site.Data.products $product).name }}
```
sanderson, line 3: I didn't test this. Is it working?

sanderson (approved):
Looks good. I didn't test the template update. Just want to make sure it's working.

PR #5629, opened 2024-10-02
fix: clarify behaviour of `influxdb.cardinality()`

btasker:
The functionality of `influxdb.cardinality()` is not directly analogous to `SHOW SERIES CARDINALITY` because it's time bounded. 

Unlike the InfluxQL command, it reports the cardinality of a specific dataset, but is often the cause of confusion as a result of being interpreted as providing the cardinality of the bucket _as a whole_.

There are various caveats which need to be considered when using it to obtain the cardinality of the bucket as a whole

* `start` needs to be earlier than the earliest timestamp (some docs pages suggest `1` which is good)
* `stop` needs to be later than the latest timestamp (the default is `now()` which is problematic if users are writing into the future)
* There's a small nuance on replicated systems (such as Cloud2) - the reported value may depend on which replica has respond (results will be eventually consistent, but may not always directly match). Not sure we necessarily need to squeeze that in

Note: a corresponding change will need to be made in the Flux code comments, but I thought it best to work on wording here first.

- [x] Rebased/mergeable

### DIFF
--- a/content/flux/v0/stdlib/influxdata/influxdb/cardinality.md
+++ b/content/flux/v0/stdlib/influxdata/influxdb/cardinality.md
@@ -1,7 +1,7 @@
 ---
 title: influxdb.cardinality() function
 description: >
-  `influxdb.cardinality()` returns the series cardinality of data stored in InfluxDB.
+  `influxdb.cardinality()` returns the series cardinality of data retrieved from InfluxDB.
 menu:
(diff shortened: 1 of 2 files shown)


sanderson (approved):

btasker:
I've raised https://github.com/influxdata/flux/pull/5506 to mirror these changes in the Flux comments

PR #5634, opened 2024-10-03
Note removal of Flux MQTT package support OSS 2.7+

sanderson:
Closes influxdata/DAR#437

Adds a warning to the top of the `experimental/mqtt` package and function pages:

image

Removes support from InfluxDB OSS 2.7+ and nightly in the InfluxDB version support modal:

image

- [x] Rebased/mergeable

### DIFF
--- a/data/flux_stdlib_frontmatter.yml
+++ b/data/flux_stdlib_frontmatter.yml
@@ -923,14 +923,15 @@
   prepend:
     block: warn
     content: |
-      #### Not supported in InfluxDB OSS
+      #### Only supported in InfluxDB Cloud (TSM)
 
-      The `experimental/mqtt` package is not supported in InfluxDB, but is
-      supported in InfluxDB Cloud (TSM). The package is still available to
-      import in OSS, but functions will not successfully publish to an MQTT broker.
+      The `experimental/mqtt` package only supported in InfluxDB Cloud (TSM).
+      It is still available to import in InfluxDB OSS and Enterprise, but
+      functions will not successfully publish to an MQTT broker.
   exclude_from:
-    oss: *
     nightly: true
+    oss: ^*
+    enterprise: ^*
 
 /flux/v0/stdlib/experimental/mqtt/to.md: |
   aliases:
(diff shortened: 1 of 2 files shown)


davidby-influx:
I don't think `mqtt` is supported by any of the OSS builds, not merely 2.7????

data/flux_stdlib_frontmatter.yml
```diff
@@ -920,11 +920,46 @@
...
     - /influxdb/v2/reference/flux/stdlib/experimental/mqtt/
     - /influxdb/cloud/reference/flux/stdlib/experimental/mqtt/
+  prepend:
+    block: warn
+    content: |
+      #### Not supported in InfluxDB OSS 2.7+
+
+      The `experimental/mqtt` package is not supported in InfluxDB 2.7+, but
+      is supported in InfluxDB Cloud (TSM). The package is still available to
+      import, but functions will not successfully publish to an MQTT broker.
```
jstirnaman, line 930:
```suggestion
      import in OSS, but functions will not successfully publish to an MQTT broker.
```

data/flux_stdlib_frontmatter.yml
```diff
@@ -920,11 +920,46 @@
...
     - /influxdb/v2/reference/flux/stdlib/experimental/mqtt/to/
     - /influxdb/cloud/reference/flux/stdlib/experimental/mqtt/to/
+  prepend:
+    block: warn
+    content: |
+      #### Not supported in InfluxDB OSS 2.7+
+
+      The `experimental/mqtt` package is not supported in InfluxDB 2.7+, but
+      is supported in InfluxDB Cloud (TSM). The package is still available to
+      import, but functions will not successfully publish to an MQTT broker.
```
jstirnaman, line 946:
```suggestion
      import in OSS, but functions will not successfully publish to an MQTT broker.
```

data/flux_stdlib_frontmatter.yml
```diff
@@ -920,11 +920,46 @@
...
+
+/flux/v0/stdlib/experimental/mqtt/publish.md: |
+  prepend:
+    block: warn
+    content: |
+      #### Not supported in InfluxDB OSS 2.7+
+
+      The `experimental/mqtt` package is not supported in InfluxDB 2.7+, but
+      is supported in InfluxDB Cloud (TSM). The package is still available to
+      import, but functions will not successfully publish to an MQTT broker.
```
jstirnaman, line 959:
```suggestion
      import in OSS, but functions will not successfully publish to an MQTT broker.
```

jstirnaman (changes requested):

jstirnaman (approved):
LGTM, pending David's approval.

data/flux_stdlib_frontmatter.yml
```diff
@@ -920,11 +920,46 @@
...
 /flux/v0/stdlib/experimental/mqtt/to.md: |
   aliases:
     - /influxdb/v2/reference/flux/stdlib/experimental/mqtt/to/
     - /influxdb/cloud/reference/flux/stdlib/experimental/mqtt/to/
+  prepend:
+    block: warn
+    content: |
+      #### Not supported in InfluxDB OSS 2.7+
+
+      The `experimental/mqtt` package is not supported in InfluxDB 2.7+, but
```
davidby-influx, line 944: I think it would be best to say `InfluxDB OSS 2.x` or whatever we use to designate all the OSS 2 versions.  Say `2,7/=` implies it is supported in earlier 2,x versions.

davidby-influx (changes requested):
A suggestion.

data/flux_stdlib_frontmatter.yml
sanderson: Was support for it removed from all previous v2 versions?

davidby-influx: I don't think it was ever supported on 2.X OSS.

sanderson: Ok, then that likely means it was never supported in OSS v1 as well?

davidby-influx: I doubt it, but have never tested it.  I suspect that it was only usable in C2.

sanderson: @davidby-influx I just pushed a change that removes support from all InfluxDB versions except InfluxDB Cloud 2 (TSM). I also updated the text in the warning.

Side question--does this package rely on dependencies outside of the Flux runtime? What does this work in Cloud, but not anything else?

PR #5636, opened 2024-10-04
Reorganize InfluxDB Clustered installation documentation

sanderson:
Closes #5592 
Closes influxdata/DAR#444
Related to influxdata/EAR/#5515

This PR reorganizes the InfluxDB Clustered installation to following the current Phased approach used with PoCs. This updates other documentation related to the installation process as well.

A preview of this branch will be available at our docs staging site (I'll send it to #team-clustered in Slack).

- [x] Rebased/mergeable

PR #5639, opened 2024-10-09
docs: add other required fields to user setup for clustered

david-rusnak:
Hey team! A customer attempted to use azure docs during clustered set up and ran into an issue with the current code example for setting azure users in the `appInstance`. I've added the rest of the fields as required (similar to keycloak).

I see there is [this PR](https://github.com/influxdata/docs-v2/pull/5636) out for reorganizing the clustered docs, so maybe it can be done there to avoid conflicts? Let me know the best way forward, thanks!

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/admin/users/add.md
+++ b/content/influxdb/clustered/admin/users/add.md
@@ -119,7 +119,12 @@ spec:
         jwksEndpoint: |-
           https://AUTH0_HOST/.well-known/openid-configuration
         users:
-          - AUTH0_USER_ID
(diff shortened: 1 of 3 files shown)


sanderson (approved):
Thanks @david-rusnak! I'll take care of porting these over to the other PR.

david-rusnak:
@sanderson Hey Scott, just a note on this that @eatondustin1 brought up to me. While it's true that the email can be arbitrary for auth purposes, for any auditing the customer wants to do - they should probably use a valid email/identifier so any actions that track the user who initiated it can be audited later.

PR #5641, opened 2024-10-10
InfluxDB Enterprise 1.11.7

sanderson:
Release notes and updates for InfluxDB 1.11.7

> [!NOTE]
> Don't merge until we get the 👍🏻 from Support.

- [x] Rebased/mergeable


davidby-influx (commented):
Should this also include changes to the code shared with OSS from 1.11.6?  There are some changes there, as well:

- chore: upgrade to go1.22.7
- fix: [GROUP BY queries with offset that crosses a DST boundary fail.](https://github.com/influxdata/influxdb/issues/25078)
- fix:  [range() does not filter correctly depending on year](https://github.com/influxdata/influxdb/issues/24669)

jstirnaman (approved):

sanderson:
@davidby-influx How do I identify the changes that were shared with OSS? The only release notes for OSS 1.11 that I can find don't specify what specific version the commits are associated with. Just the unreleased version of OSS.

Are these all that should be included or are there more?

sanderson:
> The big release notes will be for the first catch-up release of 1.11 OSS, where we have to list everything that happened from 1.8 to 1.11... That's going to be a lot of work.

Yep, doing that now 😄 .

PR #5652, opened 2024-10-16
Clustered sizing recommendations per vendor

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb/clustered/install/set-up-cluster/prerequisites.md
+++ b/content/influxdb/clustered/install/set-up-cluster/prerequisites.md
@@ -87,8 +87,8 @@ InfluxDB Clustered Kubernetes deployments require `kubectl` 1.27 or higher.
 
 ### Cluster sizing recommendation
 
-For a [medium-sized workload](https://www.influxdata.com/resources/influxdb-3-0-vs-oss/),
-InfluxData recommends the following sizing for {{% product-name %}} components:
+As a starting point for a production workload, InfluxData recommends the
+following sizing for {{% product-name %}} components:
 
 {{< tabs-wrapper >}}
 {{% tabs %}}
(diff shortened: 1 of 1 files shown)


content/influxdb/clustered/install/set-up-cluster/prerequisites.md
```diff
@@ -299,7 +374,7 @@ PostgreSQL-compatible database.
 ### PostgreSQL-compatible database requirements
 
 - PostgreSQL versions **13.8–14.6**.
-- Minimum of 4 GB of memory or equivalent provider-specific units.
+- **Minimum of 4 GB of memory** or equivalent provider-specific units.
 - To avoid conflicts and prevents issues caused by shared usage with other
```
jstirnaman, line 378:
```suggestion
- To avoid conflicts and prevent issues caused by shared usage with other
```

jstirnaman (approved):
Looks good! 1 minor unrelated fix.

content/influxdb/clustered/install/set-up-cluster/prerequisites.md
```diff
@@ -88,14 +88,89 @@ InfluxDB Clustered Kubernetes deployments require `kubectl` 1.27 or higher.
 ### Cluster sizing recommendation
 
 For a [medium-sized workload](https://www.influxdata.com/resources/influxdb-3-0-vs-oss/),
```
barbaranelson, line 90: Is this the correct link? I went to the linked doc, but it didn't seem to say anything about a medium sized workload.

barbaranelson (approved):
Looks good to me. Thanks.

content/influxdb/clustered/install/set-up-cluster/prerequisites.md
sanderson: I think this was the link that was in the original installation google doc. I assume it's referring to the workload used in the benchmarking the white-paper covers.

Edit: I just read through the white-paper and it doesn't ever explicitly outline the workload. It describes how the workload was increased until degradation was observed in each product. I think I'll just take this link out.

sanderson: @barbaranelson I don't have any insight into what comprises a medium workload, but in the [1.x hardware sizing doc](https://docs.influxdata.com/influxdb/v1/guides/hardware_sizing/#single-node-or-cluster), we state this for those considering single node OSS vs Enterprise:

> If your InfluxDB performance requires any of the following, a single node (InfluxDB OSS) may not support your needs:
>
> - more than 750,000 field writes per second
> - more than 100 moderate queries per second

Is this in the ballpark of what we would consider a medium workload?

PR #5656, opened 2024-10-18
Feature: Autogenerate Telegraf docs from influxdata/telegraf repo.

jstirnaman:
Adds templates from the `telegraf-internal` repo to permanently reside in `/docs-v2/telegraf-build`.

Supports auto-generating Telegraf documentation from the Telegraf release process.
The Telegraf release workflow will trigger the build of a `docs` binary. When run, the script extracts README file content from the Telegraf release, applies the templates in this PR, and generates a pull request to the `docs-v2` repo.

Depends on PR https://github.com/influxdata/telegraf-internal/pull/488

image
image


build-telegraf/plugins.sh
```diff
@@ -0,0 +1,112 @@
...
+menu:
+  telegraf_v1_ref:
+    parent: $parent
+    name: $(basename "$plugin_dir")
+tags: [$(basename "$plugin_dir")]
+---
+
+EOF
+  else
+    echo "Plugin directory not found: $plugin_dir"
```
sanderson, line 36: We currently have 20 external plugins that are listed on the plugins page but are not part of the Telegraf repo. How do we want to handle these? @srebhan, do you have any thoughts?

layouts/shortcodes/telegraf/plugins.html
```diff
@@ -23,5 +23,6 @@ <h3 id="{{ $type }}-{{ .id }}">{{ .name }}</h3>
     <a class="btn github-link" href="{{ if .link }}{{ .link }}{{ else }}https://github.com/influxdata/telegraf/blob/release-{{ $latestTelegrafVersion }}/plugins/{{ $type }}s/{{ .id }}/README.md{{ end }}" target="_blank">
       <span class="icon-github"></span> <span class="hide">View</span>
     </a>
+    <div>{{ .Site.GetPage "/telegraf/v1/input-plugins/apache/_index.md" }}</div>
```
sanderson, line 26: I assume you'll iterate over the page data to do something here?

build-telegraf/plugins.sh
```diff
@@ -0,0 +1,112 @@
...
+    mkdir -p "$dest_dir"
+    # Use a heredoc to write the frontmatter to the destination file
+    cat <<EOF > "$dest_file"
+---
+description: "Telegraf plugin for collecting metrics from $plugin_dir"
+menu:
+  telegraf_v1_ref:
+    parent: $parent
+    name: $(basename "$plugin_dir")
+tags: [$(basename "$plugin_dir")]
```
sanderson, line 31: One thing that would be nice to account for here is `introduced` and `deprecated` versions. Were you thinking we'd do that by injecting frontmatter or adding that frontmatter to the readmes?

jstirnaman:
@srebhan Merging to support PR influxdata/telegraf-internal#488

PR #5662, opened 2024-10-29
feat: add `--pid-file` and `--overwrite-pid-file`

gwossum:
Add documentation related to PID file.

These options will be available starting in OSS 2.7.11.

### DIFF
--- a/content/influxdb/v2/reference/config-options.md
+++ b/content/influxdb/v2/reference/config-options.md
@@ -155,6 +155,8 @@ To configure InfluxDB, use the following configuration options when starting the
 - [nats-max-payload-bytes](#nats-max-payload-bytes) <em style="opacity:.65">- (deprecated)</em>
 - [nats-port](#nats-port) <em style="opacity:.65">- (deprecated)</em>
 - [no-tasks](#no-tasks)
+- [overwrite-pid-file](#overwrite-pid-file)
+- [pid-file](#pid-file)
 - [pprof-disabled](#pprof-disabled)
 - [query-concurrency](#query-concurrency)
 - [query-initial-memory-bytes](#query-initial-memory-bytes)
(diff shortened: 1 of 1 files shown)


gwossum:
PR implementing feature in OSS: https://github.com/influxdata/influxdb/pull/25474

sanderson (changes requested):
Thanks for this @gwossum! Could you add these new flags to the yaml file `data/influxd_flags.yml` as well? We use this file to generate a [list of flags on the `influxd` landing page](https://docs.influxdata.com/influxdb/v2/reference/cli/influxd/#flags).

Also, when are we looking to release 2.7.11?

gwossum:
@sanderson The plan was to release this in 2.7.11. I noticed that [influxd_flags.yml](https://github.com/influxdata/docs-v2/blob/master/data/influxd_flags.yml) only lists major and minor for when flags were added. Is it terrible to add this new flag on a patch release, even if it is primarily for a single use-case?

sanderson:
@gwossum 

> I noticed that [influxd_flags.yml](https://github.com/influxdata/docs-v2/blob/master/data/influxd_flags.yml) only lists major and minor for when flags were added.

This data file was originally added when we were documenting each minor version in the versions specified in the `influxd_flags.yml` were used to filter flags based on the current minor version. We no longer document each minor version so this property is essentially obsolete. You can just put 2.7. It won't make a difference.

> Is it terrible to add this new flag on a patch release, even if it is primarily for a single use-case?

Generally this would be considered a "new feature" and would be added on a minor release, but I think it's fine. I don't think this needs to wait until 2.8.

gwossum:
@sanderson `influxd-flags.yml` has been updated.

sanderson (approved):
Thanks @gwossum. I added a new influxdb-2.7.11 feature branch and changed the base on this PR from `master` to that branch. That way we can merge this PR, but release all the 2.7.11 features and release notes when it goes live.

PR #5664, opened 2024-11-01
feat: add --storage-wal-flush-on-shutdown

gwossum:
Add description of `--storage-wal-flush-on-shutdown` configuration parameter. This option will be available in the OSS 2.7.11 release.


jstirnaman (approved):
Looks good. Thanks!

PR #5666, opened 2024-11-05
fix(clustered): document known bug and workaround in 20241024-1354148 release notes

waynr:
This has come up with a couple Clustered customers and with our own internal testing in environments where a given PostgreSQL DSN contains connection options.

### DIFF
--- a/content/influxdb/clustered/reference/release-notes/clustered.md
+++ b/content/influxdb/clustered/reference/release-notes/clustered.md
@@ -76,11 +76,6 @@ spec:
 # ...[remaining configuration]
 ```
 
-Your AppInstance may have additional configuration, but the important part here
-is the `spec.package.spec.images.overrides` section where we override the
-built-in image used by the `core` pods with one that has a bugfix for the DSN
-parsing error.
-
 ### Highlights
 
 #### AppInstance image override bug fix
--- a/content/influxdb/clustered/reference/release-notes/clustered.md
+++ b/content/influxdb/clustered/reference/release-notes/clustered.md
@@ -73,7 +73,7 @@ spec:
         overrides:
           - name: 'influxdb2-artifacts/granite/granite'
             newFQIN: 'us-docker.pkg.dev/influxdb2-artifacts/granite/granite:7acf9ca6e1ad15db80b22cd0bc071acdb561eb51'
-# ...additional configuration left out for brevity
+# ...[remaining configuration]
 ```
 
 Your AppInstance may have additional configuration, but the important part here
--- a/content/influxdb/clustered/reference/release-notes/clustered.md
+++ b/content/influxdb/clustered/reference/release-notes/clustered.md
@@ -53,8 +53,10 @@ Due to incorrect parsing of the
 `POSTGRES_DSN` environment variable, the `influxdb&options=-c%20search_path=` string is
 interpreted as the database name.
 
-The workaround to deal with this before a fix makes its way into the next
-release is to use an image override in your AppInstance that looks like:
+To work around this bug, in your AppInstance, 
+include a `spec.package.spec.images.overrides` section to override the
+`core` pods built-in image with an image that has the bugfix for the DSN
+parsing error--for example:
 
 ```
 apiVersion: kubecfg.dev/v1alpha1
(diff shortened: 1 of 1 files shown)


content/influxdb/clustered/reference/release-notes/clustered.md
```diff
@@ -36,6 +36,49 @@ spec:
...
+  namespace: influxdb
+spec:
+  package:
+    image: us-docker.pkg.dev/influxdb2-artifacts/clustered/influxdb:20241024-1354148
+    apiVersion: influxdata.com/v1alpha1
+    spec:
+      images:
+        overrides:
+          - name: 'influxdb2-artifacts/granite/granite'
+            newFQIN: 'us-docker.pkg.dev/influxdb2-artifacts/granite/granite:7acf9ca6e1ad15db80b22cd0bc071acdb561eb51'
```
eatondustin1, line 73: Is this how we would run a local or GKE Clustered instance with a locally-build or not-yet-promoted Granite image as well? If so, this is definitely useful outside of the release notes as well 🙌

eatondustin1 (approved):
🚀

content/influxdb/clustered/reference/release-notes/clustered.md
waynr: Yeah! I've done it with IOx back when I was implementing license enforcement.

Although, it's probably not very useful for you to develop a granite image against a released version of Clustered given how fast things change -- for example, if you tried to do this with a Granite build from today's main branch the `accounts` and `core` pods will crash loop because of new required args/envvars (I know because I tried a more recent version of Granite than what's shown here before I went with is image).

It's probably more useful to render Clustered yaml using `kubecfg` (see https://github.com/influxdata/k8s-idpe/tree/master/env/clustered#rendering-yaml-without-the-operator) and use `kubectl` to deploy those to the cluster with whatever Granite changes you want made directly in that repo.

content/influxdb/clustered/reference/release-notes/clustered.md
```diff
@@ -36,6 +36,49 @@ spec:
...
+logs that look like the following:
+
+```
+2024-11-04T01:00:00.000Z | 3: error returned from database: database "influxdb&options=-c%20search_path=" does not exist
+2024-11-04T01:00:19.000Z | 4: database "influxdb&options=-c%20search_path=" does not exist
+```
+
+The issue here is that entire `influxdb&options=-c%20search_path=` string is
+being interpreted as the database name, owing to incorrect parsing of the
+`POSTGRES_DSN` environment variable.
```
jstirnaman, line 54:
```suggestion
Due to incorrect parsing of the
`POSTGRES_DSN` environment variable, the `influxdb&options=-c%20search_path=` string is
interpreted as the database name.
```

content/influxdb/clustered/reference/release-notes/clustered.md
```diff
@@ -36,6 +36,49 @@ spec:
...
+
+The workaround to deal with this before a fix makes its way into the next
+release is to use an image override in your AppInstance that looks like:
```
jstirnaman, line 57:
```suggestion
To work around this bug, in your AppInstance, 
include a `spec.package.spec.images.overrides` section to override the
`core` pods built-in image with an image that has the bugfix for the DSN
parsing error--for example:
```

content/influxdb/clustered/reference/release-notes/clustered.md
```diff
@@ -36,6 +36,49 @@ spec:
...
+spec:
+  package:
+    image: us-docker.pkg.dev/influxdb2-artifacts/clustered/influxdb:20241024-1354148
+    apiVersion: influxdata.com/v1alpha1
+    spec:
+      images:
+        overrides:
+          - name: 'influxdb2-artifacts/granite/granite'
+            newFQIN: 'us-docker.pkg.dev/influxdb2-artifacts/granite/granite:7acf9ca6e1ad15db80b22cd0bc071acdb561eb51'
+# ...additional configuration left out for brevity
```
jstirnaman, line 74:
```suggestion
# ...[remaining configuration]
```

content/influxdb/clustered/reference/release-notes/clustered.md
```diff
@@ -36,6 +36,49 @@ spec:
...
+```
+
+Your AppInstance may have additional configuration, but the important part here
+is the `spec.package.spec.images.overrides` section where we override the
+built-in image used by the `core` pods with one that has a bugfix for the DSN
+parsing error.
+
```
jstirnaman, line 81:
```suggestion
```

jstirnaman (changes requested):
Just some cleanup suggestions. Thanks!

jstirnaman (approved):

PR #5668, opened 2024-11-07
Update Flux/InfluxDB version modal

sanderson:
We kept the InfluxDB `minor_versions` data in the `products.yml` for the sole purpose of generating the Flux/InfluxDB function and package compatibility modal. I moved the minor versions that support Flux into a separate data file and updated the Flux version fetch script to pull from the new file.

It the process of this update, I also noticed that the modal didn't include OSS 1.11 and to make it sort properly, I had to apply the same hack I had done for the Enterprise versions. It's verbose, but it works. I also future-proofed it in case v2 ever goes to a 2-digit minor version.

- [x] Rebased/mergeable


data/flux_influxdb_minor_versions.yml
```diff
@@ -0,0 +1,24 @@
+# This is a list of all InfluxDB minor versions that support Flux.
+# It is used by flux-build-scripts/update-flux-versions.js to generate the
+# "Supported Versions" section of the Flux page.
+# When new versions of InfluxDB are released that support Flux, this list should
```
jstirnaman, line 4: Just an idea: it would be cool to have a CI checklist prompt (if we don't ignore it) whenever we add a version or product.

flux-build-scripts/update-flux-versions.js
```diff
@@ -85,12 +85,12 @@ function isSemanticNewer(semverX, semverY) {
 }
 
 // Load product data from the product data file
-const productData = yaml.load(fs.readFileSync(path.resolve(__dirname,'../data/products.yml'), 'utf8'))
+const influxdbMinorVersions = yaml.load(fs.readFileSync(path.resolve(__dirname,'../data/flux_influxdb_minor_versions.yml'), 'utf8'))
 
 // Update InfluxDB version arrays by removing 'v' from each version
```
jstirnaman, line 90:
```suggestion
```

layouts/partials/footer/modals/flux-influxdb-versions.html
```diff
@@ -41,21 +41,83 @@ <h3>InfluxDB support for {{ $itemRef | safeHTML }}</h3>
...
+        {{ end }}
+      {{ end }}
+
+      <!-- Loop over only v1 versions with 2+-digit minor version -->
+      {{ range $key, $value := $oss }}
+        {{ $versionSemVer := split $value "." }}
+        {{ $supported := cond (ge (index $versionSemVer 0) (index $introducedSemVer 0)) (cond (ge (index $versionSemVer 1) (index $introducedSemVer 1)) (cond (ge (index $versionSemVer 2) (index $introducedSemVer 2)) true false) false) false }}
+        {{ $deprecated := and (isset $.Page.Params "deprecated") (cond (ge (index $versionSemVer 0) (index $deprecatedSemVer 0)) (cond (ge (index $versionSemVer 1) (index $deprecatedSemVer 1)) (cond (ge (index $versionSemVer 2) (index $deprecatedSemVer 2)) true false) false) false) }}
+        {{ $excluded := gt (len (findRE (string $excludePatternOSS) $key)) 0 }}
+        {{ if and (in $key "1.") (gt (len $key) 3) }}
```
jstirnaman, line 70: I guess I should learn to read this.

jstirnaman (approved):
LGTM!

layouts/partials/footer/modals/flux-influxdb-versions.html
sanderson: Yeah, I don't love Go templating, but it gets the job done. One key to reading it is understanding that the syntax always has the operator/function first, then the operands (`operator left-operand right-operand`).

PR #5701, opened 2024-12-04
docs: update argument in python tutorial authorizations code.

karel-rehor:
This is in reaction to https://github.com/influxdata/influxdb-client-python/issues/680, improved with https://github.com/influxdata/influxdb-client-python/issues/682.  

It was discovered that the Authorizations API currently requires a keyword argument when creating a new Authorization using an already existing instance.  The example here has mislead some users.  

- [x] Rebased/mergeable


jstirnaman (approved):
Thanks for fixing this!

PR #5706, opened 2024-12-09
fix: Clustered may use any PostgreSQL 13 or 14

jacobmarble:
PostgreSQL minor version releases do not add features, only fix bugs, so there's no need to limit to versions up to 14.6.
- https://www.postgresql.org/docs/13/release.html
- https://www.postgresql.org/docs/14/release.html

- [x] Rebased/mergeable


sanderson (approved):
Thanks!

PR #5707, opened 2024-12-09
Enterprise: use `-single-server` flag for a single meta node

### LINKED ISSUE #5695: Single-node enterprise meta node flag missing in docs?
The existing 'single node enterprise setup docs' found [here](https://docs.influxdata.com/enterprise_influxdb/v1/introduction/installation/single-server) do not appear to have a mention of a flag that's recommended in the plutonium docs [here](https://github.com/influxdata/plutonium/blob/master/README.md#single-node-setup), specifically the `-single-server` flag that ensures that the single meta node is able to become leader. 

Should we add that to the setup docs, along with an example how to alter the service to add the flag?

jstirnaman:
Closes #5695

Miscellaneous fixes for Enterprise

### DIFF
--- a/content/enterprise_influxdb/v1/administration/manage/clusters/rebalance.md
+++ b/content/enterprise_influxdb/v1/administration/manage/clusters/rebalance.md
@@ -100,7 +100,7 @@ The following command lists every shard in our cluster:
 influxd-ctl show-shards
 ```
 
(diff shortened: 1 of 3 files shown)


danatinflux:
LGTM, nice to see the example for the system file in the docs.

PR #5719, opened 2024-12-20
WIP: Docs fix broken list code example 

MeelahMe:
Closes #

- [ ] Rebased/mergeable

### DIFF
--- a/content/influxdb/cloud-dedicated/reference/client-libraries/v3/python.md
+++ b/content/influxdb/cloud-dedicated/reference/client-libraries/v3/python.md
@@ -14,7 +14,8 @@ aliases:
   - /influxdb/cloud-dedicated/reference/client-libraries/v3/pyinflux3/
 related:
   - /influxdb/cloud-dedicated/query-data/execute-queries/troubleshoot/
-list_code_example: >
+---
+
 {{< list_code_example language="python" >}}
 ```python
 from influxdb_client_3 import InfluxDBClient3, WriteOptions, write_client_options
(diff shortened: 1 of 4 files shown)


content/influxdb/cloud-dedicated/reference/client-libraries/v3/python.md
```diff
@@ -14,43 +14,36 @@ aliases:
   - /influxdb/cloud-dedicated/reference/client-libraries/v3/pyinflux3/
 related:
   - /influxdb/cloud-dedicated/query-data/execute-queries/troubleshoot/
-list_code_example: >
-  <!-- Import for tests and hide from users.
-  ```python
-  import os
-  ```
-  -->
-  <!--pytest-codeblocks:cont-->
```
jstirnaman, line 23: If you just remove the comments and commented lines, it should fix the list page display--that's the quick fix.

As a more strategic fix, I like your idea, but...
- it would also require changing the section (`_index`) page template to support it.
- the page that includes the example would need to hide it from displaying

I'm fine with a quick fix. For the "strategic" fix, my preference would probably be to change the parsing/rendering of the existing `list_code_example` frontmatter field to support the test setup comments.

jstirnaman: ...just thinking out loud... a better "fix" might be to have the test runner inject

  ```python
  import os
  ```

for each example before running it - that eliminates the need for alot of pytest comments in the pages.

jstirnaman (changes requested):

PR #5725, opened 2024-12-31
Port cookies to local storage

sanderson:
The data we're storing currently in cookies is actually better suited for local storage. Moving to local storage removes some dependencies and simplifies the integration for cookie/security compliance.

- [x] Rebased/mergeable


jstirnaman (approved):
🔥 Heads-off unnecessary privacy scope and reduces page load. Thanks for doing this!

PR #5729, opened 2025-01-04
AI chat widget and JavaScript components

jstirnaman:
Ask AI example questions:
- Adds support for customizing example Ask AI questions per product or version.
- Configure questions in site `data/products.yml`; otherwise, it uses default questions from `ask-ai.js`

Context, page, and product data:
- Adds sample URLs for remaining versions in influxdb_urls
- `page-context.js` consolidates and exports constants for page context (protocol, host, path, referrer) and path-to-data mappings for product and influxdb_url site data

Module refactor:
- Refactors some JavaScript into ES6 modules, and refactors some of those further into a Component pattern--just vanilla JS and no shadow DOM stuff. The Component pattern that uses data attributes to "bind" JavaScript modules with CSS and HTML is a popular approach in modern web development. This pattern enhances modularity, reusability, and maintainability by associating behavior (JavaScript), structure (HTML), and style (CSS) through the use of data attributes.
- `assets/main.js` is the entrypoint
- Passes pageParams from the Hugo page to modules that import `@params`.
- Moves most external dependencies out of `script` tags and into package.json to be managed with `yarn`.
- Adds `eslint`.
- For modules that aren't yet components, wraps execution statements inside an `initialize()` function and calls the function from `main.js` on `DOMContentLoaded`.
- For components, if the page contains the `data-component="<component-name>"`, the matching element is passed to the component function on `DOMContentLoaded`.
- I tried to avoid changing any logic where it wasn't necessary.

image


layouts/partials/topnav.html
```diff
@@ -15,15 +15,15 @@
     </div>
     <div class="buttons">
       <div class="search-btn">
-        <button id="search-btn"
-          onclick="toggleSidebar('sidebar-open');document.getElementById('algolia-search-input').focus();return false;"><span
-            class="cf-icon Search_New"></span></button>
```
sanderson, line 20: This shouldn't be removed or at least it needs to be refactored. When the sidebar is closed, a search icon appears in the top nav next to the gear icon. When a user clicks the search icon, this opens the sidebar and focuses the search text field.

assets/js/docs-themes.js
```diff
@@ -1,42 +0,0 @@
...
-    }
-    setPreference(style_preference_name, css_title.replace(/-theme/, ''));
-  }
-}
-
-function setStyleFromCookie () {
-  var css_title = `${getPreference(style_preference_name)}-theme`;
-  if (css_title !== undefined) {
-    switchStyle(css_title);
-  }
```
sanderson, line 41: This functionality seems to be missing after the refactor. The page doesn't honor the set theme preference when the page loads:

Kapture 2025-01-27 at 15 39 26

assets/js/theme.js
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  // Enable the stylesheet with the specified title
+  const link = document.querySelector(`link[rel*="stylesheet"][title="${css_title}"]`);
+  link && (link.disabled = false);
+
+  setPreference(PROPS.style_preference_name, css_title.replace(/-theme/, ''));
+}
+
+export default function Theme({ style }) {
```
sanderson, line 27: I would think that this would execute on page load, but it doesn't appear to be.

assets/styles/layouts/_footer-widgets.scss
```diff
@@ -0,0 +1,137 @@
...
+////////////////////////////// CUSTOM TIME TRIGGER /////////////////////////////
+
+.custom-time-trigger {
+  a {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 100%;
+    height: 100%;
+    font-size: 1.2rem;
```
sanderson, line 93:
```suggestion
    font-size: 1.2rem; 
    text-decoration: none;
```

data/influxdb_urls.yml
```diff
@@ -66,3 +70,9 @@ cloud:
         - name: East US (Virginia)
           location: Virginia, USA
           url: https://eastus-1.azure.cloud2.influxdata.com
+dedicated:
```
sanderson, line 73: Just a nit–in other places, we've used the key `cloud-dedicated` instead of just `dedicated`. InfluxDB Cloud Serverless just inherits data from `cloud`

data/products.yml
```diff
@@ -7,6 +7,10 @@ influxdb3_core:
   list_order: 2
   latest: core
   placeholder_host: localhost:8181
+  sample_questions:
```
sanderson, line 10: I'm wondering if we should call this `ai_sample_questions`, just to provide more specific context.

layouts/partials/header.html
```diff
@@ -3,7 +3,7 @@
 {{ $version := index $productPathData 1 }}
 
 <!doctype html>
-<html lang="en">
+<html lang="en" data-theme="light">
```
sanderson, line 6: Was this a debugging addition or is it meant to be permanent? While testing the theme switcher, this doesn't appear to change at all.

layouts/partials/header/stylesheets.html
```diff
@@ -22,3 +29,10 @@
 
 {{ $stylesAPIOverrides := resources.Get "styles/styles-api.scss" | toCSS $cssAPIOverrides | postCSS $PostCSSOptions }}
 <link rel="stylesheet" type="text/css" href="{{ $stylesAPIOverrides.RelPermalink }}">
+
+<!-- Show body after DOM is fully loaded -->
+<script>
```
sanderson, line 34: We should probably load the preferred theme before the content is made visible.

sanderson (changes requested):
Lots of work. Thanks for the JS refactor. I did find a few bugs, but overall, great stuff.

layouts/partials/topnav.html
jstirnaman: Oof, not sure why I deleted that. Sorry

jstirnaman: Restored and refactored into module.

jstirnaman:
@sanderson Fixed the bugs you found. See the video. Thanks for the careful review.

https://github.com/user-attachments/assets/8600c3e6-e905-457e-b306-8307c6868014

sanderson (approved):
I think this is ready to fly!

Ready to fly

PR #5755, opened 2025-01-13
feat: update getting started guides

pauldix:
Here's the updated getting started guides for Core and Enterprise. I put them into the index page and getting started since I wanted the main landing page to be the guide.


content/influxdb3/core/_index.md
```diff
@@ -1,18 +1,497 @@
...
+print(table.schema)
+
+print("\n#### Use PyArrow to read the specified columns\n")
+print(table.column('temp'))
+print(table.select(['room', 'temp']))
+print(table.select(['time', 'room', 'temp']))
+
+print("\n#### Use PyArrow compute functions to aggregate data\n")
+print(table.group_by('hum').aggregate([]))
+print(table.group_by('room').aggregate([('temp', 'mean')]))
```
jstirnaman, line 285:
```suggestion
print("\n#### Use PyArrow to read the specified columns\n")
print(table.column('usage_active'))
print(table.select(['host', 'usage_active']))
print(table.select(['time', 'host', 'usage_active']))

print("\n#### Use PyArrow compute functions to aggregate data\n")
print(table.group_by('host').aggregate([]))
print(table.group_by('cpu').aggregate([('time_system', 'mean')]))
```

Replace `'home` example columns with `cpu` columns.

content/influxdb3/core/_index.md
```diff
@@ -1,18 +1,497 @@
...
+influxdb3 create distinct_cache -h
+```
+
+## **Python Plugins and the Processing Engine**
+{{% note %}}
+As of this writing, the Processing Engine is only supported in Docker environments.<br/>
+We expect it to launch in non-Docker environments soon. We're still in very active development creating the API and developer experience so things will break and change fast. Join our <a href=https://discord.com/invite/eMnhxPyj>Discord</a> to ask questions and give feedback.
+{{% /note %}}
+
+InfluxDB3 has an embedded Python VM for running code inside the database. Currently, we only support plugins that get triggered on WAL file flushes, but more will be coming soon. Specifically, plugins will be able to be triggered by:
```
jstirnaman, line 368:
```suggestion
InfluxDB 3 has an embedded Python VM for running code inside the database. Currently, we only support plugins that get triggered on WAL file flushes, but more will be coming soon. Specifically, plugins will be able to be triggered by:
```

jstirnaman (approved):
Thanks! Just suggested some fixes that I don't want to forget. I'll port the pages into `/shared` to reuse across both `index` and `get-started` and do some minor cleanup, but won't bug you with that here. I expect questions around `accept_partial` behavior, so will expand on that ASAP.

PR #5794, opened 2025-01-21
AI chat styles

sanderson:
Adds custom styles to the AI chat modal. This is just a start, but I think it's good enough.

- [x] Rebased/mergeable


jstirnaman (approved):
Thanks!

PR #5806, opened 2025-01-27
Update content for 72 hour limit change

sanderson:
- Updates the Core getting started content with the 72 hour limit changes
- Adds the new `query-file-limit` config option.

---

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb3/core/reference/config-options.md
+++ b/content/influxdb3/core/reference/config-options.md
@@ -54,6 +54,7 @@ influxdb3 serve
   - [object-store](#object-store)
   - [data-dir](#data-dir)
   - [node-id](#node-id)
+  - [query-file-limit](#query-file-limit)
 - [AWS](#aws)
   - [aws-access-key-id](#aws-access-key-id)
   - [aws-secret-access-key](#aws-secret-access-key)
@@ -134,9 +135,9 @@ influxdb3 serve
 ### General
 
 - [object-store](#object-store)
-- [bucket](#bucket)
 - [data-dir](#data-dir)
 - [node-id](#node-id)
+- [query-file-limit](#query-file-limit)
 
 #### object-store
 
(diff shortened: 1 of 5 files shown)


content/influxdb3/core/reference/config-options.md
```diff
@@ -178,6 +179,25 @@ configuration--for example, the same bucket.
 
 ---
 
+#### query-file-limit
+
+Limits the number of Parquet files a query can access. Default is `432` (72 hours).
```
pauldix, line 184:
```suggestion
Limits the number of Parquet files a query can access. Default is `432`. With the default setting of gen1 time blocks of 10 minutes, this can be up to 72 hours, but could be less depending on gen1 configuration and whether all data for a given 10 minute block of time was ingested during the same period.
```

jstirnaman (approved):
👍

content/influxdb3/core/reference/config-options.md
```diff
@@ -178,6 +179,25 @@ configuration--for example, the same bucket.
...
+You can increase this limit to allow more files to be queried, but be aware of
+the following side-effects:
+
+- Degraded query performance for queries that read more Parquet files
+- Increased memory usage
+- Your system potentially killing the `influxdb3` process due to Out-of-Memory
```
pauldix, line 190:
```suggestion
- Your system potentially killing the `influxdb3` process due to Out-of-Memory
- If using object storage backend, many GET requests to access the data (as many as 2 per file)
```

content/influxdb3/core/reference/config-options.md
```diff
@@ -178,6 +179,25 @@ configuration--for example, the same bucket.
...
+  (OOM) errors
+
+We recommend keeping the default setting and querying smaller time ranges.
```
pauldix, line 193:
```suggestion
We recommend keeping the default setting and querying smaller time ranges. If you require longer time range queries or faster query performance on any query that accesses an hour of data or more, our InfluxDB 3 Enterprise product optimizes this with a compactor that rearranges files to achieve faster query performance.
```

pauldix (approved):
Just some small modifications.

PR #5816, opened 2025-02-03
chore(e2e): Add Cypress for link checking and end-to-end tests. Fix b…

jstirnaman:
…roken links revealed by tests.

- Adds Cypress and a few basic tests for the global topnav, the home page, and link-checking.
- For link-checking, pass a comma-delimited list of URLs in an exported cypress_test_subjects environment variable. For examples, see the convenience commands in package.json

@sanderson My first priority was to find a link checker, but obviously while refactoring JavaScript, I was really wishing for a test suite. This is just a quick spike, but I'm already wishing I hadn't put it off for so long.

If we decide to keep it, I'll add it into the Lefthook pre-commit checks.


sanderson (approved):
Awesome. This looks great.

PR #5817, opened 2025-02-03
Staging/ai chat widget

jstirnaman:
Bug fixes, `flux` product matching, and more custom example questions.
Deployed to staging.

### DIFF
--- a/data/products.yml
+++ b/data/products.yml
@@ -172,4 +172,4 @@ flux:
   ai_sample_questions:
     - How do I write a Flux query for InfluxDB v2 (TSM)?
     - How do I use Flux to transform data stored in InfluxDB v2?
-    - How do I query data stored in InfluxDB 3?
+    - How do I join data with Flux?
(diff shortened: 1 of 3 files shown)


data/products.yml
```diff
@@ -161,3 +169,7 @@ flux:
   list_order: 8
   versions: [v0]
   latest: v0.x
+  ai_sample_questions:
+    - How do I write a Flux query for InfluxDB v2 (TSM)?
+    - How do I use Flux to transform data stored in InfluxDB v2?
+    - How do I query data stored in InfluxDB 3?
```
sanderson, line 175: I'm wondering if providing a sample question that will give them a negative answer ("You can use Flux with InfluxDB 3"), maybe something like:

```suggestion
    - How do I join data with Flux?
```

or

```suggestion
    - How do I send notifications with Flux?
```

sanderson (approved):
Looks good. Just a suggestion for a different example question for Flux.

data/products.yml
jstirnaman: Thanks, I'll use these. Yeah, I had mixed feelings, but wanted to catch users landing there unintentionally.

PR #5821, opened 2025-02-04
Change comments in line protocol syntax to be self-descriptive `#`

maximal:
I suppose, the current
```js
// Syntax
<measurement>[,<tag_key>=<tag_value>[,<tag_key>=<tag_value>]] <field_key>=<field_value>[,<field_key>=<field_value>] [<timestamp>]

// Example
myMeasurement,tag1=value1,tag2=value2 fieldKey="fieldValue" 1556813561098000000
```
can confuse new users by implying that `//` can be used to comment lines in the line protocol itself.

So, I propose to change this snippet to
```python
# Syntax
<measurement>[,<tag_key>=<tag_value>[,<tag_key>=<tag_value>]] <field_key>=<field_value>[,<field_key>=<field_value>] [<timestamp>]

# Example
myMeasurement,tag1=value1,tag2=value2 fieldKey="fieldValue" 1556813561098000000

```
and use some #-comment language descriptor in the Markdown code block (like Python).

### DIFF
--- a/content/influxdb/v2/reference/syntax/line-protocol.md
+++ b/content/influxdb/v2/reference/syntax/line-protocol.md
@@ -27,11 +27,11 @@ It is a text-based format that provides the measurement, tag set, field set, and
 - [Naming restrictions](#naming-restrictions)
 - [Duplicate points](#duplicate-points)
 
-```js
-// Syntax
+```python
+# Syntax
 <measurement>[,<tag_key>=<tag_value>[,<tag_key>=<tag_value>]] <field_key>=<field_value>[,<field_key>=<field_value>] [<timestamp>]
 
-// Example
+# Example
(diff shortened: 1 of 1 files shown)


jstirnaman:
If we do this for one product, we should do it for all.

jstirnaman:
Thanks for the fix! Folded into #7705, which also splits the syntax template from the runnable example (so it's not mislabeled as `python` either) and applies your fix to the shared source file this content now lives in (`content/shared/influxdb-v2/reference/syntax/line-protocol.md` — the file this PR targets became a stub pointing there). Your commit is preserved in that branch's history and credited via co-author trailer. Closing this in favor of #7705.

maximal:
@jstirnaman, thanks!

PR #5835, opened 2025-02-14
chore: Add product-key shortcode

jstirnaman:
- Returns the first item from <productdata>.versions in products.yml.
- Currently, long and short forms are the same.
- Borrows the term 'edition' from the influxdb3 install scripts--seemed apt and I don't know how we might extend long or short forms.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -546,6 +546,25 @@ Insert warning markdown content here.
 {{% /warn %}}
 ```
 
+### Product data
+
+Display the full product name and version name for the current page--for example:
+
+- InfluxDB 3 Core
+- InfluxDB 3 Cloud Dedicated
+
+```md
+{{% product-name %}}
+```
+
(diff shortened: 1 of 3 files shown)


jstirnaman:
My rudimentary first pass.

sanderson (changes requested):
I think this is probably overkill for what it does. A simpler approach might be a `product-key` shortcode that just returns the string from the directory path:

```
{{- $productPathData := findRE "[^/]+.*?" .Page.RelPermalink -}}
{{- $currentProduct := index $productPathData 1 -}}
{{ $currentProduct }}
```

jstirnaman:
> I think this is probably overkill for what it does. A simpler approach might be a `product-key` shortcode that just returns the string from the directory path:
> 
> ```
> {{- $productPathData := findRE "[^/]+.*?" .Page.RelPermalink -}}
> {{- $currentProduct := index $productPathData 1 -}}
> {{ $currentProduct }}
> ```

Thanks - I considered that, but it seemed too "from the hip". I'll go with it, though!

PR #5849, opened 2025-02-20
fix: influxdb3 no longer has MUSL builds

jdstrand:
Reference:
- https://github.com/influxdata/influxdb/blob/main/README_processing_engine.md#why-not-just-statically-link-with-eg-musl

The enablement of the process engine for non-docker removed the ability to have MUSL builds. This corrects that in the docs.

ASIDE: the buttons for https://docs.influxdata.com/influxdb3/core/?t=Docker+%28x86%29 and https://docs.influxdata.com/influxdb3/enterprise/?t=Docker+%28x86%29 both say 'Docker (x86)'. We have ARM (ie, arm64) images for Docker these days so the button should be adjusted. I couldn't see how to do that.


sanderson (approved):
Thanks @jdstrand!

PR #5852, opened 2025-02-20
chore(mono): Update architectures for Docker images.

### LINKED ISSUE #5850: influxdb3 core/enterprise install docs say 'Docker (x86)', but ARM images are available
The buttons for https://docs.influxdata.com/influxdb3/core/?t=Docker+%28x86%29 and https://docs.influxdata.com/influxdb3/enterprise/?t=Docker+%28x86%29 both say 'Docker (x86)'. We also have ARM (ie, arm64) images for Docker these days so the button should be adjusted. I couldn't see how to do that.

Not sure if you want to adjust the button to include ARM or to adjust it to simply say 'Docker' and have the text below it mention x86 and ARM images are available....

jstirnaman:
Closes #5850

### DIFF
--- a/content/influxdb3/core/install.md
+++ b/content/influxdb3/core/install.md
@@ -44,7 +44,8 @@ Use the InfluxDB 3 quick install script to install {{< product-name >}} on
     {{< product-name >}} package on your local machine:
 
     ```bash
-    curl -O https://www.influxdata.com/d/install_influxdb3.sh && sh install_influxdb3.sh
+    curl -O https://www.influxdata.com/d/install_influxdb3.sh \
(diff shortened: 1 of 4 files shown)


jdstrand (approved):
LGTM. Personally, I think `x86_64` is enough (as opposed to `x86_64 (AMD64)`), but that's your call.

jstirnaman:
> LGTM. Personally, I think `x86_64` is enough (as opposed to `x86_64 (AMD64)`), but that's your call.

Thanks, I'm going to keep it because of `docker pull --platform linux/amd64 ` and people like me that can't keep them straight.

PR #5877, opened 2025-03-05
fix(v3): Replace "Clustered" references in monolith...

### LINKED ISSUE #5874: Replace "Clustered" references in monolith https://docs.influxdata.com/influxdb3/core/write-data/

jstirnaman:
- Replace names with product-name shortcode
- Be more precise with grammar for the Object storage model (where data isn't exactly stored _in_ InfluxDB 3)

Closes #5874


sanderson (approved):
👌🏻

PR #5882, opened 2025-03-09
Update _index.md

volodya-melnyk:
This pull request fixes the InfluxDB installation instructions in the documentation. Specifically, sudo was added to some commands that require administrative privileges to properly execute repository addition and InfluxDB installation.

Changes:

Added sudo before commands that require admin rights:
for adding the repository key;
for adding the repository to the sources list;
for updating packages and installing InfluxDB.


jstirnaman (approved):
Thanks!

PR #5896, opened 2025-03-17
API docs fixes

jstirnaman:
Fixes positioning of Ask AI in API docs on small displays
JS guards for empty pageParams and its attributes.

Tested on Staging.

The JS errors were only exposed when viewing:
- https://docs.influxdata.com/influxdb3/core/api/v3/ and
- https://docs.influxdata.com/influxdb3/enterprise/api/v3/ (and in staging)
However, not when viewing the _same documents_ (under alias) at
- https://docs.influxdata.com/influxdb3/core/api/
- https://docs.influxdata.com/influxdb3/enterprise/api/

### DIFF
--- a/api-docs/template.hbs
+++ b/api-docs/template.hbs
@@ -38,33 +38,58 @@
       />{{/unless}}
     {{{redocHead}}}
     <link rel="stylesheet" type="text/css" href="/api.css" />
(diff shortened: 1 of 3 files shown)


sanderson (approved):
🙏

PR #5904, opened 2025-03-17
fix(api): don't ignore .config.yml files inside of api-docs

jstirnaman:
- Adds missing `.config.yml` files needed for generating Core and Enterprise API references
- Don't ignore `.config.yml` files in `api-docs`.


sanderson (approved):
🎉

jstirnaman:
@sanderson Thanks for solving this!

PR #5905, opened 2025-03-17
fix(code-controls): Broken in code-blocks that contain the URL select…

jstirnaman:
…or widget: Replace jQuery .prev() with .prevAll() to find all siblings (in case they include other elements like the URL selector widget), and then take the first pre element that has a code child.

Fixes https://influxdata.slack.com/archives/C084G9LR2HL/p1742229715680079


sanderson (approved):
Thanks!

PR #5908, opened 2025-03-17
update: add info on distinct values cache

peterbarnett03:
Had a user not know how to use the Distinct values cache, so adding boilerplate information on usage, similar to what we have for last-cache.


jstirnaman (approved):
🚀

PR #5910, opened 2025-03-18
Update Grafana token instructions for Core and Entperprise

sanderson:
Closes https://github.com/influxdata/DAR/issues/491

Clarifies that the `Token` string in an InfluxDB data source configuration should be an arbitrary, non-empty string. However, if you started the server with a token, you should provide that token.

- [x] Rebased/mergeable


jstirnaman (approved):
🔥 Thanks for this!

PR #5913, opened 2025-03-18
fix(influxdb3): Enterprise: update options in CLI reference:

### LINKED ISSUE #5911: Update influxdb3 Enterprise CLI reference
Update `mode` and other options in CLI reference.

https://docs.influxdata.com/influxdb3/enterprise/reference/config-options/#mode

jstirnaman:
Add options new in Beta
Remove old options, add see references, add anchors for backward compat. Fix link checker test to not fail when no links exist in the article

Closes #5911

### DIFF
--- a/content/influxdb3/enterprise/reference/config-options.md
+++ b/content/influxdb3/enterprise/reference/config-options.md
@@ -856,7 +856,8 @@ Specifies the size of the RAM cache used to store data, in bytes.
 #### exec-mem-pool-bytes
 
 Specifies the size of memory pool used during query execution.
-Can be given as absolute value in bytes or as a percentage of the total available memory (for example: `10%`).
+Can be given as absolute value in bytes or as a percentage of the total available memory--for
+example: `8000000000` or `10%`).
 
 **Default:** `20%`
 
@@ -1222,6 +1223,7 @@ engine uses.
 Specifies the Python package manager that the processing engine uses.
 
 This option supports the following values:
+
 - `discover` _(default)_: Automatically discover available package manager
 - `pip`: Use pip package manager
 - `uv`: Use uv package manager
--- a/content/influxdb3/enterprise/reference/config-options.md
+++ b/content/influxdb3/enterprise/reference/config-options.md
@@ -54,6 +54,7 @@ influxdb3 serve
   - [object-store](#object-store)
   - [data-dir](#data-dir)
   - [node-id](#node-id)
+  - [cluster-id](#cluster-id)
   - [mode](#mode)
   - [license-email](#license-email)
   - [query-file-limit](#query-file-limit)
@@ -112,7 +113,6 @@ influxdb3 serve
 - [Memory](#memory)
   - [ram-pool-data-bytes](#ram-pool-data-bytes)
   - [exec-mem-pool-bytes](#exec-mem-pool-bytes)
-  - [buffer-mem-limit-mb](#buffer-mem-limit-mb)
   - [force-snapshot-mem-threshold](#force-snapshot-mem-threshold)
 - [Write-Ahead Log (WAL)](#write-ahead-log-wal)
   - [wal-flush-interval](#wal-flush-interval)
@@ -123,9 +123,6 @@ influxdb3 serve
   - [read-from-node-ids](#read-from-node-ids)
   - [replication-interval](#replication-interval)
 - [Compaction](#compaction)
-  - [compactor-id](#compactor-id)
-  - [compact-from-node-ids](#compact-from-node-ids)
-  - [run-compactions](#run-compactions)
   - [compaction-row-limit](#compaction-row-limit)
   - [compaction-max-num-files-per-plan](#compaction-max-num-files-per-plan)
   - [compaction-gen2-duration](#compaction-gen2-duration)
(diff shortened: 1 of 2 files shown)


content/influxdb3/enterprise/reference/config-options.md
```diff
@@ -197,17 +195,30 @@ configuration--for example, the same bucket.
...
 This option supports the following values:
 
-- `read`
-- `read_write` _(default)_
-- `compactor`
+- `all` _(default)_
+- `ingest`
+- `query`
+- `compact`
+- `process`
```
sanderson, line 219: I remember seeing discussion that this option should accept a list, so you can pass multiple modes to a node. @peterbarnett03 can you confirm? If so, we need to know the list syntax to use for the command line option and the environment variable.

I'm wouldn't hold this PR up on this feedback, but I think we should give a description of what each mode does.

content/influxdb3/enterprise/reference/config-options.md
```diff
@@ -844,36 +855,25 @@ Specifies the size of the RAM cache used to store data, in bytes.
 
 #### exec-mem-pool-bytes
 
-Specifies the size of the memory pool used during query execution, in bytes.
+Specifies the size of memory pool used during query execution.
+Can be given as absolute value in bytes or as a percentage of the total available memory (for example: `10%`).
```
sanderson, line 859:
```suggestion
Can be given as absolute value in bytes or as a percentage of the total available memory--for
example: `8000000000` or `10%`).
```

content/influxdb3/enterprise/reference/config-options.md
```diff
@@ -844,36 +855,25 @@ Specifies the size of the RAM cache used to store data, in bytes.
...
-**Default:** `5000`
-
-| influxdb3 serve option  | Environment variable            |
-| :---------------------- | :------------------------------ |
-| `--buffer-mem-limit-mb` | `INFLUXDB3_BUFFER_MEM_LIMIT_MB` |
-
----
-
 #### force-snapshot-mem-threshold
+<span id="buffer-mem-limit-mb" />
```
sanderson, line 870: I assume this span is here to act as a link anchor, but why? Did the configuration name change?

content/influxdb3/enterprise/reference/config-options.md
```diff
@@ -844,36 +855,25 @@ Specifies the size of the RAM cache used to store data, in bytes.
...
 
 Specifies the threshold for the internal memory buffer. Supports either a
-percentage (portion of available memory)of or absolute value
-(total bytes)--for example: `70%` or `100000`.
+percentage (portion of available memory) or absolute value in MB
+(total bytes)--for example: `70%` or `1000 MB`.
```
sanderson, line 874: > or absolute value in MB (total bytes)

Is it in MB or bytes? Should the value contain a unit (MB, GB, etc.)?

content/influxdb3/enterprise/reference/config-options.md
```diff
@@ -972,47 +972,31 @@ Defines the interval at which each replica specified in the
...
 
-Specifies the prefix in the object store where all compacted data is written.
-Provide this option only if this server should handle compaction for its own
-write buffer and any replicas it manages.
-
-| influxdb3 serve option | Environment variable                |
-| :--------------------- | :---------------------------------- |
-| `--compactor-id`       | `INFLUXDB3_ENTERPRISE_COMPACTOR_ID` |
+> [!Important]
+> This option has been superseded by using `--mode compact`. See the [mode](#mode) section for details.
```
sanderson, line 985: If this is no longer available, we should just remove it.

content/influxdb3/enterprise/reference/config-options.md
```diff
@@ -972,47 +972,31 @@ Defines the interval at which each replica specified in the
...
-run compactions for a given `compactor-id`. This option is only applicable if a
-`compactor-id` is set.
-
-| influxdb3 serve option | Environment variable                   |
-| :--------------------- | :------------------------------------- |
-| `--run-compactions`    | `INFLUXDB3_ENTERPRISE_RUN_COMPACTIONS` |
+> [!Important]
+> This option has been superseded by using `--mode compact`. See the [mode](#mode) section for details.
 
 ---
```
sanderson, line 1001: Same. Just remove these. If we haven't already, we should make a note in the release notes that these have been removed and replaced by the `mode` options.

content/influxdb3/enterprise/reference/config-options.md
```diff
@@ -1225,7 +1221,12 @@ engine uses.
 
 Specifies the Python package manager that the processing engine uses.
 
-**Default:** `10s`
+This option supports the following values:
+- `discover` _(default)_: Automatically discover available package manager
```
sanderson, line 1225:
```suggestion
This option supports the following values:

- `discover` _(default)_: Automatically discover available package manager
```

sanderson (changes requested):
Some suggestions. We also need to update the list of options on the `influxdb3 serve` page:

https://docs.influxdata.com/influxdb3/enterprise/reference/cli/influxdb3/serve/#options

jstirnaman:
> Some suggestions. We also need to update the list of options on the `influxdb3 serve` page:
> 
> https://docs.influxdata.com/influxdb3/enterprise/reference/cli/influxdb3/serve/#options

I'll do that now. I'm still working on CLI reference

content/influxdb3/enterprise/reference/config-options.md
jstirnaman: span is the old option name. I added all the spans to prevent breaking existing links. I know we don't usually, but didn't want users to be **as** confused

sanderson (approved):
👌🏻

PR #5939, opened 2025-03-28
fix: fix link formatting

anjaflux:
An extra space between the `()` and `[]` broke the link formatting:

Screenshot from 2025-03-28 12-03-10

This space was removed.

- [X] Rebased/mergeable


jstirnaman (approved):
👍 Thanks for the fix!

PR #5953, opened 2025-04-02
Update shared content usage in v2 docs and more

sanderson:
This PR includes far-reaching changes that remove old, brittle ways of sharing content across pages. The old `duplicate-oss` used regular expression replaces to remove unneeded content on page render. The process was somewhat unpredictable and often led to broken, half-rendered pages. The new `show-in` and `hide-in` shortcodes conditionally show or hide content blocks on page render without any need for regular expression evaluation. They are much more predictable and stable.

Changes:

- Replaces all `oss-only` and `cloud-only` shortcodes with the new `show-in` shortcode.
- Replaces all `duplicate-oss` shortcodes with shared content sources.
- Moves all shared v2 oss content into `content/shared/`.
- Updates the structure and rendering of `prepend` and `append` frontmatter to accept just a markdown string instead of separate `block` and `content` properties.

---

- [x] Rebased/mergeable


jstirnaman (approved):
Spot-checked, but LGTM. This should be a huge stride for managing content across versions. THANKS! 🚀 🚀

PR #5955, opened 2025-04-03
Overhaul canonical pages

sanderson:
The primary focus of this PR is to overhaul how to set `canonical` metadata on documentation pages to prevent duplicate content penalties in search engines. As search engines find duplicate content on a site, they penalize the duplicate content. However, if you use a `canonical` link tag in the page header, search engines don't penalize the content, but rather weight search results towards the specified canonical source.

The logic defined in the `canonical.html` template does the following:

1. Sets the default canonical source as the current page.
2. Checks for `canonical` frontmatter and, if it exists, uses it to override the default.
3. Checks to see if the page has a `source` frontmatter. If it does, it iterates through all pages that share the same source and overrides the default canonical source based on the following product priority:
    1. enterprise
    2. core
    3. cloud-dedicated
    4. clustered
    5. cloud-serverless
    6. v2
    7. cloud

This PR also standardizes `source` paths by removing `/content` from the begging (that was my bad).

- [x] Rebased/mergeable


jstirnaman (approved):
LGTM!

PR #5970, opened 2025-04-10
InfluxDB 3 Core and Enterprise GA visual updates

sanderson:
- Added a GA notification with styles specific for the GA announcement
- Removed the beta content block from the top of all Core and Enterprise pages along with related styles
- Relocated community links to the existing support and feedback section at the bottom of each page.

---

- [x] Rebased/mergeable


jstirnaman (approved):
🙏

PR #5971, opened 2025-04-11
Add LVC management docs to Core and Enterprise

sanderson:
- [x] Rebased/mergeable

### DIFF
--- a/content/shared/influxdb3-admin/last-value-cache/create.md
+++ b/content/shared/influxdb3-admin/last-value-cache/create.md
@@ -1,0 +1,122 @@
+
+Use the [`influxdb3 create last_cache` command](/influxdb3/version/reference/cli/influxdb3/create/last_cache/)
+to create a Last Value Cache (LVC). Provide the following:
+
+- **Database** (`-d`, `--database`): _({{< req >}})_ The name of the database to
+  associate the LVC with. You can also use the `INFLUXDB3_DATABASE_NAME`
+  environment variable to specify the database.
+- **Token** (`--token`): _({{< req >}})_ Your {{< product-name >}}
+  authentication token with write access to the specified table.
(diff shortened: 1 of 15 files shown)


content/shared/influxdb3-admin/last-value-cache/create.md
```diff
@@ -0,0 +1,122 @@
...
+
+> [!Note]
+> #### Values are cached on write
+>
+> Values are cached on write. When you create a cache, it will not cache
+> previously written points, only newly written points.
+>
+> #### LVC size and persistence
+>
+> The LVC is stored in memory, so it's important consider the size and persistence
```
jstirnaman, line 120:
```suggestion
> The LVC is stored in memory, so it's important to consider the size and persistence
```

content/shared/influxdb3-admin/last-value-cache/query.md
```diff
@@ -0,0 +1,34 @@
...
+> [!Important]
+> You must use SQL to query the LVC.
+> InfluxQL does not support the `last_cache()` function.
+
+`last_cache()` supports the following arguments:
+
+- **table_name**: _({{< req >}})_ The name of the table the LVC is associated with
+  formatted as a string literal.
+- **cache_name**: The name of the LVC to query formatted as a string literal.
+  This only required if there is more than one LVC associated with the specified
```
jstirnaman, line 15:
```suggestion
  This argument is only required if there is more than one LVC associated with the specified
```

jstirnaman (changes requested):
Looks good - a few minor fixes. Thanks!

jstirnaman:
Not urgent for this PR, but we need to add instructions for using API.

jstirnaman (approved):

PR #5978, opened 2025-04-14
Core and Enterprise auth tokens

jstirnaman:
Create, manage, and use tokens

Enterprise has admin, db, and system tokens
Core has admin tokens (all or nothing)

Flow:

User starts with influxdb3 serve
Auth is enabled, but there's no token, so nothing can be written to the DB
Alternative: User starts with influxdb3 serve --without-auth to ignore authentication (user can also not create/delete tokens then)
User creates a token with influxdb3 create token --admin
This places the hashed token in the Catalog, and users can immediately read/write to/from the DB by using the unhashed token in queries/writes
User created a database token with influxdb3 create token --permission "db:*:read" --name generalReading
Places in the Catalog, can immediately be used

[Requirements doc for more ABAC info](https://influxdata.slack.com/archives/CDBJ35NBV/p1744651751452089?thread_ts=1744126364.814089&cid=CDBJ35NBV)


jstirnaman:
Only` content/shared/v3-enterprise-get-started/_index.md` is accurate for now.

jstirnaman:
@sanderson Only` content/shared/v3-enterprise-get-started/_index.md` is accurate for now.

content/shared/v3-enterprise-get-started/_index.md
```diff
@@ -231,7 +231,125 @@ influxdb3 serve --help
...
-_During the alpha period, licenses are valid until May 7, 2025._
+_During the beta period, licenses are valid until May 7, 2025._
+
+### Authentication and authorization
+
+{{% product-name %}} uses token-based authentication and authorization which is enabled by default when you start the server.
+With authentication enabled, you must provide an _admin token_ or _database token_ to access the server.
+
+- **admin token**: Grants access to all CLI actions and API endpoints.
+_ **database token**: Scoped to a specific database and grant access to write and query data in that database.
```
sanderson, line 242:
```suggestion
_ **database token**: Scoped to a specific database and grants access to write and query data in that database.
```

content/shared/v3-enterprise-get-started/_index.md
```diff
@@ -231,7 +231,125 @@ influxdb3 serve --help
...
+  Creating a database token requires an admin token.
+
+After you have [started the server](#start-influxdb), you can create and manage tokens using the `influxdb3` CLI or the HTTP API.
+
+When you create a token InfluxDB 3 returns a token string in clear text
```
sanderson, line 247:
```suggestion
When you create a token, InfluxDB 3 returns a token string in plain text
```

content/shared/v3-enterprise-get-started/_index.md
```diff
@@ -231,7 +231,125 @@ influxdb3 serve --help
...
+that you can use to authenticate CLI commands (using ) and API requests.
+
+
+Securely store your token, as you won't be able to retrieve it later.
```
sanderson, line 251: Redundant with the important block below, but I might move that block up to here.
```suggestion
```

content/shared/v3-enterprise-get-started/_index.md
```diff
@@ -231,7 +231,125 @@ influxdb3 serve --help
...
+> [!Important]
+>
+> #### Securely store your tokens
+>
+> For security, InfluxDB only lets you view tokens when you create them.
+> InfluxDB 3 stores a hash of the token in the catalog, so you can't retrieve the token after it is created.
+
+#### Create an admin token
+
+To create an admin token, use the `influxdb3 create token` subcommand and pass the `--admin` flag--for example:
```
sanderson, line 265: I'm not sure why they did this, but `--admin` and `--permissions` are actually structured as subcommands, not options. So each of these have their own set of options.

content/shared/v3-enterprise-get-started/_index.md
```diff
@@ -231,7 +231,125 @@ influxdb3 serve --help
...
+
+```bash
+ influxdb3 create token --admin \
+ --host http://{{< influxdb/host >}}
+```
+
+The command returns a token string that you can use to authenticate CLI commands and API requests.
+Securely store your token, as you won't be able to retrieve it later.
```
sanderson, line 273: Redundant with the block below, but maybe move that block up here.

```suggestion
```

sanderson (changes requested):

content/shared/v3-enterprise-get-started/_index.md
jstirnaman: Yeah, ok. I just thought it was confusing terminology for `--<whatev>`

PR #5986, opened 2025-04-15
Mono shared glossary

jstirnaman:
Shared glossary file suitable for sharing among all InfluxDB 3 versions.
For this PR, only shared b/w Core and Enterprise.
Adds Processing engine.
Small fixes.

Basing against the original copy (from Cloud-Dedicated) of the glossary file for easier review.

### DIFF
--- a/content/shared/influxdb3-reference/glossary.md
+++ b/content/shared/influxdb3-reference/glossary.md
@@ -1,13 +1,3 @@
----
-title: Glossary
-description: >
-  Terms related to InfluxData products and platforms.
-weight: 109
-menu:
-  influxdb3_cloud_dedicated:
-    parent: Reference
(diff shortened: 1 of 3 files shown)


content/shared/influxdb3-reference/glossary.md
```diff
@@ -694,6 +715,21 @@ A set of predetermined rules.
...
+triggers. The Processing engine runs Python code directly in your
+database, allowing plugins to react to specific triggers without requiring external services.
+
+Related entries:
+[plugin](#plugin),
+[trigger](#trigger)
+
+
+
+
```
sanderson, line 732:
```suggestion

```

sanderson (approved):
🙌🏻 Amazing! Thanks for doing this. Just one suggestion that removes some unneeded whitespace.

Issue #5989, opened 2025-04-16
Core and Enterprise: get-started write examples


jstirnaman:
- [x] Fix <https://docs.influxdata.com/influxdb3/core/get-started/#example-write-data-using-the-influxdb3-cli|this example of writing using the CLI> hanging continuation (reported by @stuartcarnie) #5993 
- [x] Add stdin examples
- [ ] Add/fix placeholder notes (TOKEN, DATABASE_NAME)
- [ ] Finish codeblock test setup for Core and Enterprise

[Slack Message](https://influxdata.slack.com/archives/C06BQA608J3/p1744770569471819?thread_ts=1744770569.471819&cid=C06BQA608J3)

PR #5991, opened 2025-04-16
fix(mono): Get started: write example has hanging continuation, add s…

jstirnaman:
…tdin example, add placeholder notes, promote Telegraf, client libs

(part of #5989)


sanderson (approved):
👌🏻

PR #6055, opened 2025-05-12
Add code-placeholders and tabbed examples for admin in Core and Enterprise

### LINKED ISSUE #6020: Add code-placeholders in get-started and other Core and Ent examples
For Core and Enterprise, add code-placeholders in get-started, write-data, query-data, processing-engine, etc.

Reported issue:
On the *Getting Started* <https://docs.influxdata.com/influxdb3/enterprise/get-started/#use-tokens-to-authorize-cli-commands-and-api-requests|page>, it should be possible to edit the `SYSTEM TOKEN` like in the other examples:

[Slack Message](https://influxdata.slack.com/archives/C0VAV6N92/p1745532979692109)

MeelahMe:
Objective: To update Core and Enterprise with core-placeholders 

Closes #6020


content/shared/v3-core-get-started/_index.md
```diff
@@ -252,15 +252,40 @@ To have the `influxdb3` CLI use your admin token automatically, assign it to the
...
+{{% code-tabs %}}
+[CLI](#)
+[Docker](#)
+{{% /code-tabs %}}
+
+{{% code-tab-content %}}
 ```bash
 influxdb3 create token --admin \
- --host http://{{< influxdb/host >}}
+  --host http://INFLUXDB_HOST
```
sanderson, line 267: Keep the `influxdb/host` shortcode here. It should auto-populate based on the the product and allow users to update it.

content/shared/v3-core-get-started/_index.md
```diff
@@ -252,15 +252,40 @@ To have the `influxdb3` CLI use your admin token automatically, assign it to the
...
+{{% /code-tab-content %}}
+
+{{< /code-tabs-wrapper >}}
+
+Replace the following:
+
+- {{% code-placeholder-key %}}`INFLUXDB_HOST`{{% /code-placeholder-key %}}: Host address of your InfluxDB 3 server (for example, `localhost:8181`)
+- {{% code-placeholder-key %}}`CONTAINER_NAME`{{% /code-placeholder-key %}}: Name of your running Docker container
+
+{{% /code-placeholders %}}
```
sanderson, line 287: You shouldn't use `code-placeholder-key` inside of a `code-placeholders`. The placeholder keys in these list items may actually get replace if the user updates them in the code block.

Also, the placeholders are different per codeblock, so a user may get confused when we tell them to replace a placeholder when that placeholder isn't in the example they see. This is what I'd suggest:

- Have a `code-placeholder` shortcode that is specific to each codeblock and is inside of the `code-tab-content` block.
- Include the "Replace the following" direction in the `code-tab-content` after the code block. It will then get switched when the user switches code tabs.

With these suggestions, this would look something like this:

~~~md
{{< code-tabs-wrapper >}}

{{% code-tabs %}}
[CLI](#)
[Docker](#)
{{% /code-tabs %}}
{{% code-tab-content %}}

```bash
influxdb3 create token --admin \
 --host http://{{< influxdb/host >}}
```

{{% /code-tab-content %}}
{{% code-tab-content %}}

{{% code-placeholders "CONTAINER_NAME" %}}
```bash
# With Docker — in a new terminal:
docker exec -it CONTAINER_NAME influxdb3 create token --admin
```
{{% /code-placeholders %}}

Replace {{% code-placeholder-key %}}`CONTAINER_NAME`{{% /code-placeholder-key %}} with the name of your running Docker container.

{{% /code-tab-content %}}

{{< /code-tabs-wrapper >}}
~~~

content/shared/v3-enterprise-get-started/_index.md
```diff
@@ -282,15 +282,40 @@ To have the `influxdb3` CLI use your admin token automatically, assign it to the
...
+{{% /code-tab-content %}}
+
+{{< /code-tabs-wrapper >}}
+
+{{% /code-placeholders %}}
+
+Replace the following:
+
+- {{% code-placeholder-key %}}`INFLUXDB_HOST`{{% /code-placeholder-key %}}: Host address of your InfluxDB 3 server (for example, `localhost:8181`)
+- {{% code-placeholder-key %}}`CONTAINER_NAME`{{% /code-placeholder-key %}}: Name of your running Docker container
```
sanderson, line 317: Same suggestions as above.

sanderson (changes requested):

PR #6073, opened 2025-05-16
chore: add release notes for 3.0.3

peterbarnett03:
Adds release notes for 3.0.3


jstirnaman (approved):
Should we add more documentation about the operator token?

peterbarnett03:
We probably should, yes. We need more info on errors for tokens as well, as we had a user with that issue recently (specifically what InvalidToken means).

PR #6075, opened 2025-05-19
6072 update the latest stable oss callout to 3 core

jstirnaman:
Enhance `layouts/partials/article/stable-version.html` callout to support a new `succeeded_by` attribute in `data/products.yml`.

- v1 and v2/install/ now recommend using Core (the OSS successor)

Shortcode tests using example.md. Support test-only files (such as example.md) without abusing draft mode.

- Add a frontmatter test-only attribute for files like example.md. Don't use draft: true for these files.
- Add support in index.html for the attribute so that the files are only rendered when running in the testing environment.
- e2e smoke test using example.md, ensure it renders without JavaScript errors.

End-to-end testing, CI script, and JavaScript QoL improvements:

- Improved e2e test runner and link checking
- CI scripts moved from Lefthook to package.json for better CI tool support.
- Lefthook runs linters on pre-commit, runs tests on pre-push for an improved dev experience
- Cypress e2e tests run Hugo separately on port 1315


layouts/index.html
```diff
@@ -6,6 +6,13 @@
 {{ $kapacitorVersion := replaceRE "v" "" .Site.Data.products.kapacitor.latest }}
 {{ $fluxVersion := replaceRE "v" "" .Site.Data.products.flux.latest }}
 
+<!--
+  Show the page if:
+  - This is a regular page (not test-only) OR
+  - This is a test-only page BUT we're currently in the testing environment
+-->
+{{ if or (not .Params.test_only) (and .Params.test_only (eq site.Params.environment "testing")) }}
```
sanderson, line 14: I would add "development" here too. I use this page to add/update styles and shortcodes on a single page.

```suggestion
{{ if or (not .Params.test_only) (and .Params.test_only (in site.Params.environment (slice "testing" "development")) }}
```

sanderson (approved):
One minor request/suggestion.

PR #6077, opened 2025-05-19
chore(hugo): Update shortcodes and templates to avoid calling isset o…

jstirnaman:
…n <nil>
Resolves the Hugo warning/error.


sanderson (approved):
🙌🏻

PR #6085, opened 2025-05-22
hotfix: update v2 download links

sanderson:
The download path pattern for InfluxDB OSS v2 changed to include the version number in the URL path. This updates all of the v2 download links to use the new path pattern.

- [x] Rebased/mergeable

### DIFF
--- a/assets/jsconfig.json
+++ b/assets/jsconfig.json
@@ -3,7 +3,8 @@
   "baseUrl": ".",
   "paths": {
    "*": [
-    "*"
+    "*",
+    "../node_modules/*"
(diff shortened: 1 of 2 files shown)


assets/jsconfig.json
```diff
@@ -3,7 +3,8 @@
   "baseUrl": ".",
   "paths": {
    "*": [
-    "*"
+    "*",
```
jstirnaman, line 6: This keeps flapping - I need to fix it.

jstirnaman (approved):
Thanks!

PR #6098, opened 2025-05-28
Update _index.md

jaal2001:
Document did not show the arguments for --permission at the correct place. Needed to be moved from the end of the statement to behind --permission


jstirnaman:
Thanks for the fix! With our latest release, I'm trying to slim down the get-started pages.

jstirnaman (approved):

PR #6111, opened 2025-05-29
Add InfluxDB 3 Explorer installation documentation

sanderson:
Adds InfluxDB 3 Explorer installation documenation.

- [x] Rebased/mergeable


content/influxdb3/explorer/install.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Run in Query Mode
+
+```sh
+docker run -d \
+  -e MODE=query \
+  ...
+```
+
+If `MODE` is not set, the container defaults to admin mode.
```
peterbarnett03, line 185: It defaults to Query mode right now. I think we should flip that; setting to Query mode should only be possibly be the admin setting it, hence giving them Admin by default.

content/influxdb3/explorer/install.md
```diff
@@ -0,0 +1,220 @@
...
+
+---
+
+## Environment Variables
+
+| Variable       | Description                                      | Default              |
+|----------------|--------------------------------------------------|----------------------|
+| `DATABASE_URL` | Path to SQLite DB inside container               | `/db/sqlite.db`      |
+| `MODE`         | Set to `admin` or `query`                        | `admin`              |
```
peterbarnett03, line 194: Default query

content/influxdb3/explorer/install.md
```diff
@@ -0,0 +1,220 @@
...
+| Port | Protocol | Purpose                 |
+|------|----------|-------------------------|
+| 80   | HTTP     | Web access (unencrypted) |
+| 443  | HTTPS    | Web access (encrypted)   |
+
+### Custom port mapping
+
+```sh
+# Map ports to custom host values
+-p 8080:80 -p 8443:443
```
peterbarnett03, line 219: Should this match the example? So 8888:80 maybe.

peterbarnett03 (changes requested):
We default to Query right now.

PR #6128, opened 2025-06-06
chore(mono): /api/v3 processing engine request parameters:

jstirnaman:
- Update trigger-spec syntax for schedule triggers
- Specify that cron only supports extended cron format
- Provide examples and pattern validation for trigger-specs
- Update response status
- Cleanup  use
- Misc. updates and cleanup
- Add test plugins

Staging:
- https://test2.docs.influxdata.com/influxdb3/core/api/v3/#tag/Processing-engine
- https://test2.docs.influxdata.com/influxdb3/enterprise/api/v3/#tag/Processing-engine

### DIFF
--- a/content/influxdb3/clustered/admin/upgrade.md
+++ b/content/influxdb3/clustered/admin/upgrade.md
@@ -51,7 +51,7 @@ Use the following command to return the image Kubernetes uses to build your
 InfluxDB cluster:
 
 ```sh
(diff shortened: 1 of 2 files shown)


hiltontj (approved):

MeelahMe (approved):
Looks good. The one thing I noticed is that the product names are not dynamic. Is that because it is a .yml file?

PR #6130, opened 2025-06-09
fix: update enterprise config options

peterbarnett03:
Makes a host of updates to the Enterprise configuration options to reflect accurate option sets.


peterbarnett03:
I also capitalized Processing Engine as a proper noun. Unsure if we landed on a decision there.

jstirnaman:
> I also capitalized Processing Engine as a proper noun. Unsure if we landed on a decision there.

My preference is to not, unless of course it's a code entity (`ProcessingEngine`), and to reserve caps for product names.

PR #6139, opened 2025-06-17
Fix enterprise docker

### LINKED ISSUE #6138: License generation doesn't work with email prompt in `docker compose run`
Docker, esp. Compose, users should provide the email address when starting the server. For some reason,
this allows the license generation to complete, whereas providing it through the interactive options doesn't.

> users are getting tripped up with licensing when running the docker container method of setting up Enterprise as they are trying to use docker compose instead of an interactive session inside the container to launch influxdb3 and go through the license process

Affects:
- Enterprise install
- Admin license

##### Relevant URLs
- https://influxdata.slack.com/archives/C04A03CE0BH/p1750101854628109

jstirnaman:
Closes #6138 

Provide the email address to skip the email prompt when using Docker, especially Docker Compose, in order for license generation to complete.

- Install Enterprise
- Admin license
- If only one pagination button (prev or next), allow it to fill the width.
- Combines core and enterprise install into a shared page.
- Update Vale (includes brace-expansion vuln. patch)

### DIFF
--- a/content/influxdb3/enterprise/admin/license.md
+++ b/content/influxdb3/enterprise/admin/license.md
@@ -112,7 +112,7 @@ address using one of the following methods:
 - Set the `INFLUXDB3_ENTERPRISE_LICENSE_EMAIL` environment variable
 
 If the server finds a valid license file in your object store, it ignores the
-the license email option.
+license email option.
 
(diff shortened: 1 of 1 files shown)


MeelahMe (approved):
Look good to me!

Issue #6019, opened 2025-04-23
Create Dedicated Docker Usage Page and Improve Get Started Guidance


MeelahMe:
**Objective**:

Create a dedicated Docker usage page (or tab) that provides clear guidance for setting up and running monolith using Docker and Docker Compose. The goal is to offer enough direction to help users get up and running while recommending best practices—without overprescribing Docker usage.

##### Relevant URLs
- _Provide relevant URLs_

- https://docs.influxdata.com/influxdb/v2/install/use-docker-compose/
- https://hub.docker.com/_/influxdb/tags?tag=3-core&name=3-core
- https://hub.docker.com/_/influxdb/tags?tag=3-core&name=3-enterprise
- https://influxdata.slack.com/archives/C0VAV6N92/p1745432765520219

**Steps**:

- [x] Decide on placement: inline within install tab, contextual notes, or a dedicated Docker page.
- [x] Draft guidance for token generation flow within Docker environments.
- [x] Add best practices for mounting /var/lib and setting permissions (chmod).
- [x] In Docker README, add InfluxDB 3 Core and Enterprise sections at the top. Include example commands adequate to install the  image and start the server. Link to docs web site for additional options.

Lower priority for v2 and v1:
- [x] Review https://docs.influxdata.com/influxdb/v2/,  https://docs.influxdata.com/influxdb/v1/, and https://docs.influxdata.com/enterprise_influxdb/v1/ Docker installation instructions, add any missing options or notes from the Docker README to docs.
- [x] Revise the README to only Include examples adequate to install v2 and v1 images and start the server.
- [x] Link to docs web site for full documentation and additional options
- [x] Remove v2 and v1 detailed instructions from Docker README

**Impact**: Improves clarity for new users during setup and reduces friction in onboarding.

jstirnaman:
Instead of this:
> Align new content with existing Docker README structure.

This:
- [ ] In Docker README, add InfluxDB 3 Core and Enterprise sections at the top. Include example commands adequate to install the  image and start the server. Link to docs web site for additional options.

Lower priority for v2 and v1:
- [ ] Review https://docs.influxdata.com/influxdb/v2/,  https://docs.influxdata.com/influxdb/v1/, and https://docs.influxdata.com/enterprise_influxdb/v1/ Docker installation instructions, add any missing options or notes from the Docker README to docs.
- [ ] Revise the README to only Include examples adequate to install v2 and v1 images and start the server.
[...]

PR #6147, opened 2025-06-20
docs: Adding data mount prereq

MeelahMe:
Objective: 

To add guidance to the install documentation on how to correctly mount the /var/lib/influxdb3 directory and set secure permissions (chmod 700). 

Related to [#6019](https://github.com/influxdata/docs-v2/issues/6019)


MeelahMe:
Seems to be needed at this point

PR #6189, opened 2025-07-04
Add .no-shorthand to `create token --admin` options

SquidDev:
This adds the `no-shorthand` class to the options table on https://docs.influxdata.com/influxdb3/core/reference/cli/influxdb3/create/token/admin/. As this table only has two columns, the shorthand CSS is applied to the description instead, making `<code>` tags pretty hard to read:

A table of various options for the admin command. All the code snippets have a negative margin, so overlap/obscure the previous word.

- [x] Rebased/mergeable


sanderson (approved):
Thanks @SquidDev!

PR #6196, opened 2025-07-08
chore: document and use modern influxdata-archive.gpg

jdstrand:
https://www.influxdata.com/blog/linux-package-signing-key-rotation/ discusses the signing key rotation we did in 2023. We are getting to the point of having to rotate the key and this will assist in that work. Most docs and containers are still using the `influxdata-archive_compat.key` which was created for maximum compatibility, but it has limitations because it does not use the modern primary/subkey approach used by `influxdata-archive.key`. The compat key is required for very old systems like RHEL 7 (standard support ended in June 2024), Debian 9 (EOL in June 2022), Ubuntu 18.04 (standard support ended in April 2023).

RHEL 8, Debian 10 (buster; already EOL), Ubuntu 20.04 (standard support ended in April 2025), et al have new enough RPM and APT that supports `influxdata-archive.key`. While the `influxdata-archive_compat.key` isn't going away and we'll recreate it when we perform the key rotation, it is time to default to and document using `influxdata-archive.key` which will allow a better key rotation experience.

I verified the RPM configuration with Rocky Linux 8 and the APT configuration with Ubuntu 20.04 and Debian 10.


jstirnaman (approved):
👍 Thanks for updating this!

PR #6202, opened 2025-07-09
chore: always verify GPG fingerprint

jdstrand:
The influxdata-archive.key is designed to have a good user experience during key rotations, but in order to do that, we need to properly verify the key file's fingerprint and not its sha256sum (since the primary key's fingerprint won't change with key rotations, but the sha256sum necessarily will).

Update documentation for this to establish best practice and prepare users for the upcoming key rotation (and fixed a couple of small things related to this along the way).

Adjust dockerfiles to use this methodology (all dockerfiles verified to successfully build after these changes).

### DIFF
--- a/content/influxdb/v1/introduction/install.md
+++ b/content/influxdb/v1/introduction/install.md
@@ -75,7 +75,7 @@ For Ubuntu/Debian users, add the InfluxData repository with the following comman
 #   Primary key fingerprint: 24C9 75CB A61A 024E E1B6  3178 7C3D 5715 9FC2 F927
 #   Subkey fingerprint:      9D53 9D90 D332 8DC7 D6C8  D3B9 D8FF 8E1F 7DF8 B07E
 wget -q https://repos.influxdata.com/influxdata-archive.key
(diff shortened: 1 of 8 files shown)


content/influxdb/v1/introduction/install.md
```diff
@@ -85,8 +85,8 @@ echo 'deb [signed-by=/etc/apt/trusted.gpg.d/influxdata-archive.gpg] https://repo
 # influxdata-archive.key GPG fingerprint:
 #   Primary key fingerprint: 24C9 75CB A61A 024E E1B6  3178 7C3D 5715 9FC2 F927
 #   Subkey fingerprint:      9D53 9D90 D332 8DC7 D6C8  D3B9 D8FF 8E1F 7DF8 B07E
-wget -q https://repos.influxdata.com/influxdata-archive.key
-echo '943666881a1b8d9b849b74caebf02d3465d6beb716510d86a39f6c8e8dac7515 influxdata-archive.key' | sha256sum -c && cat influxdata-archive.key | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/influxdata-archive.gpg > /dev/null
+curl --silent --location -O https://repos.influxdata.com/influxdata-archive.key
```
jstirnaman, line 88: 🔥 🤦

jstirnaman (approved):
🚀 Thanks again! This is about as good as it could be.

PR #6206, opened 2025-07-09
fix(clustered): updated clustered restore process

sanderson:
Closes influxdata/DAR#520

- Updates the InfluxDB Clustered restore process to clarify that all pods must be scaled completely down before scaling down the Catalog.
- Adds a step for scaling components back up after the restore operation

---

- [x] Rebased/mergeable


jstirnaman:
Adding some verification steps

content/influxdb3/clustered/admin/backup-restore.md
```diff
@@ -300,32 +305,96 @@ using Catalog store snapshots:
...
+           
+       ```bash
+       kubectl get pods --namespace influxdb --selector=app=iox-shared-ingester
+       ```
+
+       _Once removed_, proceed to the next step.  
+
+6.  **Scale down catalog last**
+
+    _After all other pods are removed_, Use the `kubectl scale` command to scale
```
jstirnaman, line 376:
```suggestion
    _After all other pods are removed_, use the `kubectl scale` command to scale
```

PR #6214, opened 2025-07-10
feat(v3): Add naming restrictions reference and guidelines:

jstirnaman:
- **Adds a shared `reference/naming-restrictions.md`**: Add detailed [naming restrictions reference](https://github.com/influxdata/docs-v2/pull/6214/files#diff-9d40dbdb5b536e6f5af0fd0aecb33b80bd3413e2de8d74766e8993ac53569ae1)  for all InfluxDB 3 versions
- Fixes influxdata/DAR#515: discourages the use of a leading underscore in database and table names in all InfluxDB 3 versions, but doesn't say it's forbidden
- Add basic guidelines in table and database creation guides.
- Core and Ent table guides are in PR #6215 

#### Related
- https://github.com/influxdata/influxdb_iox/issues/14455


sanderson (approved):
🚀

PR #6217, opened 2025-07-11
Processing engine examples and cleanup

jstirnaman:
- show how to filter tables with an `all_tables` spec using args and plugin code
- Add separate Manage Dependencies heading
- Move external plugins for better flow
- Remove dupes


sanderson (approved):
🚀

PR #6220, opened 2025-07-14
feat(html): recursive left nav

sanderson:
Updates the left navigation template to be recursive, with unlimited depth (at least to a depth that we'll never need). This also greatly simplifies the left navigation template.

- [x] Rebased/mergeable


jstirnaman (approved):
LGTM!

PR #6227, opened 2025-07-16
http-bind port update

mdevy-influxdata:
Adjusted port numbers to allow for proper load balancer routing

Initial documentation was meant for running on a single host.  Production configurations will have nodes running on individual hosts and thus needs same port for use with load balanacer.


jstirnaman (commented):
Suggestions/questions:
- If using the default, do we need to include the option at all?
- Should we add a note that explains the new configurations assume using distinct hosts; if you run multiple servers on the same host, you can use a different port number for each instance?

jstirnaman (approved):

PR #6228, opened 2025-07-17
feat(monolith): use minio for object storage

### LINKED ISSUE #6230: MinIO
Add detailed documentation for using MinIO as the object store for InfluxDB 3 Core and Enterprise.

sanderson:
Closes #6230 

- Adds a new **Admin** > **Configure object storage** section to Core and Enterprise
- Adds a guide for using MinIO as the object store for InfluxDB.
- Added a new codeblock render hook that simplifies using code placeholders and lets you use them in lists (which would previously break if you tried to code placeholders. The new approach adds a `placeholders` attribute to a codeblock element and changes the rendering of the codeblock based on the existence of that attribute.
  
  **Old code placeholders method:**

  ~~~md
  {{% code-placeholders "USERNAME|PASSWORD" %}}
  ```bash
  influxdb query -u USERNAME -p PASSWORD
  ```
  {{% /code-placeholders %}}
  ~~~

  **New code placeholders method:**

  ~~~md
  ```bash { placeholders="USERNAME|PASSWORD" }
  influxdb query -u USERNAME -p PASSWORD
  ```
  ~~~

  The new method is only used in the MinIO documentation added in this PR, but it doesn't affect any of the existing code placeholders. We can go back and update the old patterns in a future PR. For all new PRs, I recommend using this new method.

---

- [x] Rebased/mergeable

### DIFF
--- a/content/influxdb3/core/object-storage/_index.md
+++ b/content/influxdb3/core/object-storage/_index.md
@@ -1,0 +1,17 @@
+---
+title: Configure object storage
+description: |
+  Configure {{% product-name %}} to connect to and use different object storage
+  providers.
+menu:
+  influxdb3_core:
+    name: Configure object storage
+    parent: Administer InfluxDB
+weight: 110
+influxdb3/core/tags: [object storage, S3]
+related:
+  - /influxdb3/core/reference/config-options/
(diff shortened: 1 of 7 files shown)


jstirnaman (approved):
🚢

PR #6242, opened 2025-07-23
fix(monolith): add configurable db, table, and column limits to enterprise

sanderson:
- [x] Rebased/mergeable


MeelahMe (approved):
Looks good!

PR #6243, opened 2025-07-23
Correcting POST request body property names in query SQL

### LINKED ISSUE #6239: API ref: query_sql: wrong field names
https://docs.influxdata.com/influxdb3/core/api/v3/#operation/PostExecuteQuerySQL contains wrong request body property names.
They should be the same as GET querystring params--see https://docs.influxdata.com/influxdb3/core/query-data/execute-queries/influxdb-v3-api/

MeelahMe:
…API docs

fix: align POST /api/v3/query_sql property names with GET params

Update QueryRequestObject schema to use 'db' and 'q' instead of
'database' and 'query_str' for consistency with GET endpoint.

Closes #6239

### DIFF
--- a/api-docs/influxdb3/core/v3/ref.yml
+++ b/api-docs/influxdb3/core/v3/ref.yml
@@ -1844,12 +1844,12 @@ components:
     QueryRequestObject:
       type: object
       properties:
-        database:
+        db:
           description: |
             The name of the database to query.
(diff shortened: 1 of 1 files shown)


jstirnaman (approved):
👍 Thanks! Do we need to fix it for enterprise as well?

PR #6247, opened 2025-07-25
Fix POST query endpoint request body property names in Enterprise API…

MeelahMe:
… spec

fix: align POST /api/v3/query_sql property names with GET params in Enterprise 

Update QueryRequestObject schema to use 'db' and 'q' instead of
'database' and 'query_str' for consistency with GET endpoint.

### DIFF
--- a/api-docs/influxdb3/enterprise/v3/ref.yml
+++ b/api-docs/influxdb3/enterprise/v3/ref.yml
@@ -1976,12 +1976,12 @@ components:
     QueryRequestObject:
       type: object
       properties:
-        database:
+        db:
           description: |
             The name of the database to query.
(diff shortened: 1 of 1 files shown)


jstirnaman (changes requested):
`ref.yml` looks good. Can you restore `helper-scripts/common/config/influxdb3-core-enterprise.json` (or only commit ref.yml) and update the PR branch--for example:

```
git checkout -t origin/fix/api-enterprise-docs-query-sql-field-names
git reset origin/master
git add api-docs
git commit -C 6b7ca8b
git push -f origin fix/api-enterprise-docs-query-sql-field-names
```

PR #6265, opened 2025-07-29
Fix misleading from_unixtime documentation about return type

### LINKED ISSUE #6252: `from_unixtime` does not return RFC3339
## Problem

Described and discussed in https://github.com/influxdata/influxdb/issues/26637

The docs for [`from_unixtime`](https://docs.influxdata.com/influxdb3/core/reference/sql/functions/time-and-date/#from_unixtime) claim to convert a UNIX epoch integer into an RFC3339 timestamp:

> Converts an integer to RFC3339 timestamp format (YYYY-MM-DDT00:00:00.000000000Z). Input is parsed as a [Unix nanosecond timestamp](https://docs.influxdata.com/influxdb3/core/reference/glossary/#unix-timestamp) and returns the corresponding RFC3339 timestamp.

They actually convert the integer into an Arrow `Timestamp`. That just happens to be formatted as an RFC3339 timestamp when rendered in the console, to JSON, etc.

MeelahMe:
**Problem**: The `from_unixtime` function documentation suggested it returns an RFC3339 formatted string when it actually returns a timestamp data type.

**Objective**: Update misleading return type descriptions for timestamp functions. Removing  "RFC3339" from the return type description for accuracy.

**Change**: "returns the corresponding RFC3339 timestamp" → "returns the corresponding timestamp" 

Currently updated function: `from_unixtime`

Potential function updates?:

- [ ] `to_timestamp`
- [ ] `to_timestamp_micros`
- [ ] `to_timestamp_millis`
- [ ] `to_timestamp_nanos`
- [ ] `to_timestamp_seconds`

Closes #6252


content/shared/sql-reference/functions/time-and-date.md
```diff
@@ -887,7 +887,7 @@ LIMIT 1
 
 Converts an integer to RFC3339 timestamp format (`YYYY-MM-DDT00:00:00.000000000Z`).
 Input is parsed as a [Unix nanosecond timestamp](/influxdb/version/reference/glossary/#unix-timestamp)
```
jstirnaman, line 889:
```suggestion
Converts an integer (Unix timestamp in seconds) to a timestamp value.
The underlying result is a timestamp (`Timestamp(TimeUnit::Second, None)`).
In query results, JSON, CSV, and "pretty" formats render the timestamp as an ISO 8601 string (`YYYY-MM-DDTHH:MM:SS`, without a timezone indicator).
When output to Parquet, the raw integer value (e.g., `1641042000`) is preserved.

```

jstirnaman (changes requested):
Handling dates and times is always confusing. My suggestion tries to be specific without giving too much detail. The original also had a conflict about the input precision allowed. Based on my testing, the function only supports Unix timestamps in seconds (not milli, micro, or nano):

```json
{"from_unixtime(Int64(1641231230000000000))":"ERROR: Cast error: Failed to convert 1641231230000000000 to datetime for Timestamp(Second, None)"}
```

content/shared/sql-reference/functions/time-and-date.md
```diff
@@ -885,9 +885,10 @@ LIMIT 1
 
 ## from_unixtime
 
-Converts an integer to RFC3339 timestamp format (`YYYY-MM-DDT00:00:00.000000000Z`).
-Input is parsed as a [Unix nanosecond timestamp](/influxdb/version/reference/glossary/#unix-timestamp)
-and returns the corresponding RFC3339 timestamp.
+Converts an integer (Unix timestamp in seconds) to a timestamp value.
+The underlying result is a timestamp (`Timestamp(TimeUnit::Second, None)`).
+In query results, JSON, CSV, and "pretty" formats render the timestamp as an ISO 8601 string (`YYYY-MM-DDTHH:MM:SS`, without a timezone indicator).
```
jstirnaman, line 890:
```suggestion
If you output query results as JSON (default for the API), CSV, or pretty (default for the CLI), the timestamp is formatted as an ISO 8601 string (`YYYY-MM-DDTHH:MM:SS`, without a timezone indicator).
```

PR #6327, opened 2025-08-21
chore: add more clarity to Explorer quick start

peterbarnett03:
A callout on where to visit if they run the default command, plus a note on using `host.docker.internal` since that will be a common pattern.

I got some lint errors, but they were for things I didn't do, so ran with --no-verify. My changes passed lint.


jstirnaman (approved):
🥳

PR #6349, opened 2025-08-28
Update setup.md

jaal2001:
Removing the 
 - influxdb3 in the command section of the docker-compose.yml file to achieve a working version. Using the statement from above leads to an error starting the docker container:

error: unrecognized subcommand 'influxdb3'
Usage: influxdb3 [OPTIONS] [COMMAND]
For more information, try '--help'.

### DIFF
--- a/content/shared/influxdb3-get-started/setup.md
+++ b/content/shared/influxdb3-get-started/setup.md
@@ -185,7 +185,6 @@ services:
     ports:
       - 8181:8181 
     command:
-      - influxdb3
       - serve
       - --node-id=node0
       - --cluster-id=cluster0


jstirnaman:
Hi, thanks for raising this.
We use the documented configuration in our docs-v2/compose.yaml for testing and I haven't seen the problem you describe.

What version of the influxdb3 image are you using? In the current image, the Dockerfile calls [`entrypoint.sh`](https://github.com/influxdata/influxdata-docker/blob/8b6cd22ddb8c62aeafc7888e74e09b2e08fbf6db/influxdb/3.4-enterprise/entrypoint.sh), which has the following behavior:

If the first arg is serve, it prepends influxdb3.
If the first arg starts with - (e.g., a flag), it prepends influxdb3 serve.
If the first arg is already influxdb3, it runs envsubst on all args.
Finally, it executes the command.

waynr:
Hi there, I'm an engineer from the InfluxDB 3 team. Thanks for bringing this to our attention!

As @jstirnaman pointed out, these docs are meant to support the official Docker hub image. The quay.io image tracks our main branch build and is intended mostly for internal testing. I'll file an issue on our internal tracker to address this difference, but it will be considered low priority for now.

If you need to use the quay.io image to test out fixes while working with our support team, then the right thing to do is to temporarily remove the `influxdb3` command since that is hard-coded into the entrypoint script for that image for now.

In my view, it's better to leave the docs as-is since it's clearer as an example to include the binary command name itself in the command args.

jaal2001:
Sorry, my bad. The quay.io must be a leftover from the early documentation, which I never questioned or reviewed. Switching to the image in the docs works perfectly fine, so the docs are correct.

PR #6370, opened 2025-09-09
Fix dar 524

jstirnaman:
Closes influxdata/DAR/issues/524

Trying to close the loop on influxdata/DAR/issues/524--the release notes were updated in https://github.com/influxdata/k8s-idpe/pull/67034, but never ported to docs.
I'm a little confused about Clustered release notes script - I assumed it would port the change to the release notes page in docs.


sanderson (approved):
One minor suggestion.

PR #6402, opened 2025-09-18
chore(sql): adds new syntax for sql approx functions

sanderson:
Adds new syntax for sql approx functions introduced in DataFusion 49

> [!Note]
> This is live in Serverless and Clustered, but won't be available for Clustered and Core/Enterprise until their next releases. I'm wondering if we should hold off on merging this until it's available in all products since the previous syntax still works and the new syntax will throw an error if they don't have the latest DataFusion.

- [x] Rebased/mergeable


content/shared/sql-reference/functions/aggregate.md
```diff
@@ -1570,16 +1573,18 @@ Returns the weighted approximate percentile of input values using the
 t-digest algorithm.
 
 ```sql
+approx_percentile_cont_with_weight(weight, percentile [, centroids]) WITHIN GROUP (ORDER BY expression)
+-- OR
 approx_percentile_cont_with_weight(expression, weight, percentile)
 ```
```
sanderson, line 1580:
```suggestion

##### Arguments

```

PR #6437, opened 2025-10-01
feat(caches): update explorer docs with information about querying LVCs and DVCs

sanderson:
- Adds information about using the Data Explorer to query LVCs and DVCs
- Adds additional styles for LVC and DVC badges

---

- [x] Rebased/mergeable


jstirnaman (approved):
💵

PR #6452, opened 2025-10-07
Remove using JWT for HTTP authentication with LDAP

gwossum:
Removed references to using JWT for HTTP authentication with LDAP (e.g. `[http] shared-secret` and user JWT tokens). JWT does not use LDAP for authentication, so you wouldn't be using LDAP if you use JWT authentication.

Left in references to enabling JWT authentication for meta nodes. While not necessary for LDAP, meta node authentication via JWT is recommended for all installations.

Closes #

- [x] Rebased/mergeable


jstirnaman (approved):
👍 Thanks!

PR #6481, opened 2025-10-23
fix:  Update system_metrics path

epgif:
Move from 'examples' to 'influxdata' in influxdata/influxdb3_plugins#17


epgif:
I checked all 'gh:' references btw, and saw no other problems:
`git ls-files | xargs -d\\n grep -h 'gh:[a-z]' | sed 's|^.*gh:|../influxdb3_plugins/|; s/\.py.*/.py/' | sort -u | xargs ls`

PR #6524, opened 2025-11-10
chore(powerbi): update powerbi pqx to latest version

caterryan:
update powerbi pqx to latest version


sanderson (approved):
Thanks @caterryan!

PR #6584, opened 2025-12-02
docs(influxdb3): document duplicate point write ordering and workarounds

jstirnaman:
## Summary

Expands duplicate points documentation for Cloud Dedicated and Clustered to address non-deterministic write ordering when duplicate points are flushed together.

### Changes
- Add warning callout explaining duplicate point overwrites are non-deterministic
- Add recommended patterns section with append-only approaches
- Add SQL and InfluxQL query examples for getting latest state
- Add anti-patterns section with common mistakes to avoid
- Add retention guidance for last-value tables (Cloud Dedicated only)
- Add performance considerations for append-only patterns
- Add cross-links from schema-design and optimize-writes pages

## Questions for Engineering

**Should the following also apply to Clustered?**

> Add retention guidance for last-value tables (Cloud Dedicated only)

**Should this documentation also apply to Cloud Serverless?**

The current changes are scoped to Cloud Dedicated and Clustered only (matching the original documentation scope). However, the `optimize-writes.md` page for Cloud Serverless already contains a "Write ordering for duplicate points" callout with similar content.

Can Engineering confirm whether Cloud Serverless has the same ingestion/flushing behavior that causes non-deterministic ordering when duplicate points are flushed together? If so, we should extend this documentation to include Cloud Serverless as well.

## Test plan

- [x] Build Hugo site and verify pages render correctly
- [x] Verify cross-links work for Cloud Dedicated, Clustered, and Cloud Serverless
- [x] Review SQL and InfluxQL query examples for correctness

closes influxdata/DAR#560

### DIFF
--- a/content/influxdb3/cloud-dedicated/reference/syntax/line-protocol.md
+++ b/content/influxdb3/cloud-dedicated/reference/syntax/line-protocol.md
@@ -14,5 +14,4 @@ related:
 source: /shared/v3-line-protocol.md
 ---
 
-<!-- The content of this file is at 
-// SOURCE content/shared/v3-line-protocol.md-->
+<!-- // SOURCE content/shared/v3-line-protocol.md -->
--- a/content/influxdb3/cloud-serverless/reference/syntax/line-protocol.md
+++ b/content/influxdb3/cloud-serverless/reference/syntax/line-protocol.md
@@ -13,5 +13,4 @@ related:
 source: /shared/influxdb-v2/reference/syntax/line-protocol.md
 ---
 
-<!-- The content of this file is at 
-// SOURCE content/shared/influxdb-v2/reference/syntax/line-protocol.md-->
+<!-- // SOURCE content/shared/influxdb-v2/reference/syntax/line-protocol.md -->
--- a/content/influxdb3/clustered/reference/syntax/line-protocol.md
+++ b/content/influxdb3/clustered/reference/syntax/line-protocol.md
@@ -9,12 +9,9 @@ menu:
     parent: Other syntaxes
 weight: 102
 influxdb3/clustered/tags: [write, line protocol, syntax]
-alt_links:
-  core: /influxdb3/core/reference/line-protocol/
-  enterprise: /influxdb3/enterprise/reference/line-protocol/
 related:
   - /influxdb3/clustered/write-data/
 source: /shared/v3-line-protocol.md
 ---
 
-<!--// SOURCE content/shared/v3-line-protocol.md-->
+<!--// SOURCE content/shared/v3-line-protocol.md -->
--- a/content/influxdb3/core/reference/line-protocol.md
+++ b/content/influxdb3/core/reference/line-protocol.md
@@ -16,6 +16,4 @@ aliases:
 source: /shared/v3-line-protocol.md
 ---
 
-<!--
-The content of this file is at content/shared/v3-line-protocol.md
--->
+<!--// SOURCE content/shared/v3-line-protocol.md -->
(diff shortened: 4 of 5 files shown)


content/shared/v3-line-protocol.md
```diff
@@ -284,13 +284,197 @@ but with a different field set, the field set becomes the union of the old
...
+-- All writes use the same timestamp, but add a version field
+device_status,device_id=sensor01 status="active",temperature=72.5,version=1i 1700000000000000000
+device_status,device_id=sensor01 status="active",temperature=73.1,version=2i 1700000000000000000
+device_status,device_id=sensor01 status="inactive",temperature=73.1,version=3i 1700000000000000000
+```
+
+**Problem**: The points are still duplicates (same time and tags).
+Adding a field doesn't make them unique points.
+
+#### ❌ Relying on write delays to force ordering
```
sanderson:
```suggestion
#### Don't rely on write delays to force ordering
```

content/shared/v3-line-protocol.md
```diff
@@ -284,13 +284,197 @@ but with a different field set, the field set becomes the union of the old
...
+
+```text
+-- Writing with delays between each write
+device_status,device_id=sensor01 status="active" 1700000000000000000
+# Wait 10 seconds...
+device_status,device_id=sensor01 status="inactive" 1700000000000000000
+```
+
+**Problem**: Delays don't guarantee that duplicate points won't be flushed together.
+The flush interval depends on buffer size, ingestion rate, and system load.
```
sanderson:
````suggestion
Delays don't guarantee that duplicate points won't be flushed together.
The flush interval depends on buffer size, ingestion rate, and system load.

For example, **don't do this**:

```text
-- Writing with delays between each write
device_status,device_id=sensor01 status="active" 1700000000000000000
# Wait 10 seconds...
device_status,device_id=sensor01 status="inactive" 1700000000000000000
```
````

File: content/shared/v3-line-protocol.md [hunk 3]
```diff
@@ -284,13 +284,197 @@ but with a different field set, the field set becomes the union of the old
...
+
+```text
+-- All writes use the same timestamp, but add a version field
+device_status,device_id=sensor01 status="active",temperature=72.5,version=1i 1700000000000000000
+device_status,device_id=sensor01 status="active",temperature=73.1,version=2i 1700000000000000000
+device_status,device_id=sensor01 status="inactive",temperature=73.1,version=3i 1700000000000000000
+```
+
+**Problem**: The points are still duplicates (same time and tags).
+Adding a field doesn't make them unique points.
```
sanderson:
````suggestion
Adding a field doesn't make a points unique.
Points with the same time and tag set are still considered duplicates--for example,
**don't do this**:

```text
-- All writes use the same timestamp, but add a version field
device_status,device_id=sensor01 status="active",temperature=72.5,version=1i 1700000000000000000
device_status,device_id=sensor01 status="active",temperature=73.1,version=2i 1700000000000000000
device_status,device_id=sensor01 status="inactive",temperature=73.1,version=3i 1700000000000000000
```
````

will-influxdata:
This is related for Enterprise v3:

InfluxDB v3 Enterprise uses the same underlying write semantics, so the non-determinism with duplicate points in the same batch can occur there as well. The difference is that Enterprise gives you more control over how long data stays in the “gen1” tier before being compacted/flushed via the gen1-duration setting:

https://docs.influxdata.com/influxdb3/enterprise/reference/config-options/#gen1-duration

jstirnaman:
> This is related for Enterprise v3:
> 
> InfluxDB v3 Enterprise uses the same underlying write semantics, so the non-determinism with duplicate points in the same batch can occur there as well. The difference is that Enterprise gives you more control over how long data stays in the “gen1” tier before being compacted/flushed via the gen1-duration setting:
> 
> https://docs.influxdata.com/influxdb3/enterprise/reference/config-options/#gen1-duration

Thanks @will-influxdata! I'll add it for Enterprise in a separate PR.

PR #6664, opened 2025-12-22
fix(layouts): use File.Dir for baseURL-independent product detection

jstirnaman:
## Summary

Architectural fix for product detection in shortcodes - uses `.File.Dir` instead of `.RelPermalink` to make product detection independent of baseURL configuration.

## Problem

Shortcodes parse `.RelPermalink` to determine product context, but when `baseURL` includes a path (e.g., for PR previews at `/docs-v2/pr-preview/pr-XXXX/`), the path segments shift and product detection fails.

## Solution

Creates a centralized `product-info.html` partial that extracts product/version from `.File.Dir` (the content file's directory path), which is unaffected by baseURL configuration.

## Changes

- **`layouts/partials/product-info.html`** - New partial for baseURL-independent product detection
- **`layouts/shortcodes/api-endpoint.html`** - Uses new partial
- **`layouts/shortcodes/influxdb/host.html`** - Uses new partial
- **`layouts/shortcodes/children.html`** - Uses new partial

## Benefits

- Works regardless of baseURL configuration
- Centralized logic (DRY)
- Could be extended to fix ~20+ other templates using the same vulnerable pattern

## Test plan

- [ ] Full site build passes
- [ ] PR preview builds work
- [ ] Product detection works correctly across all products
- [ ] Review impact on other templates that use `.RelPermalink` for product detection

---

**Note**: This is the broader architectural fix for later testing. A localized workaround for PR previews is in the `ci-pr-preview` branch.

### DIFF
--- a/layouts/partials/product-info.html
+++ b/layouts/partials/product-info.html
@@ -1,0 +1,74 @@
+{{- /*
+  Returns a dict with product path information extracted from the page.
+
+  Uses .File.Dir when available (baseURL-independent) with fallback to .RelPermalink.
+  This ensures product detection works correctly regardless of baseURL configuration,
+  including PR preview deployments at subdirectory paths.
+
+  Usage:
+    {{ $productInfo := partial "product-info.html" . }}
+    {{ $productInfo.product }}      - e.g., "influxdb3", "influxdb", "flux"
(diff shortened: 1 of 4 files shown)


jstirnaman:
superseded by the cascade-based product refactor already on master

PR #6681, opened 2025-12-31
Manage configs with Telegraf Controller

sanderson:
Adds information about managing Telegraf configurations with Telegraf Controller.

- [x] Rebased/mergeable

### DIFF
--- a/content/telegraf/controller/configs/dynamic-values.md
+++ b/content/telegraf/controller/configs/dynamic-values.md
@@ -41,7 +41,7 @@ Use the following syntax:
 ```
 
 Parameters do not require a default value. Any parameter without a default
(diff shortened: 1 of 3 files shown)


content/telegraf/controller/configs/dynamic-values.md
```diff
@@ -0,0 +1,176 @@
...
+> Use environment variables or secrets to provide sensitive information to agents.
+
+Use the following syntax:
+
+```
+&{param_name[:default_value]}
+```
+
+Parameters do not require a default value. Any parameter without a default
+value is considered required and must [be defined](#define-paramters) when
```
jstirnaman:
```suggestion
value is considered required and must [be defined](#define-parameters) when
```

PR #6694, opened 2026-01-04
feat(sync): extract style attributes from HTML comments

jstirnaman:
## Summary

Add support for extracting Hugo style attributes from HTML comments in plugin READMEs during the sync process.

### Problem

Hugo style attributes like `{.green}` and `{.orange}` render as literal text on GitHub, making source READMEs look messy.

### Solution

Use HTML comments to hide style attributes in source:
```markdown
#### Recommended <!-- {.green} -->
#### Not recommended <!-- {.orange} -->
```

The sync script extracts these and applies them:
```markdown
#### Recommended {.green}
#### Not recommended {.orange}
```

### Supported Classes

- `{.green}`, `{.orange}` - Color styling
- `{.recommended}`, `{.not-recommended}` - Semantic styling  
- Any other `{.classname}` format

## Test Plan

- [x] Verify regex extracts style attributes correctly
- [ ] Test sync with downsampler plugin containing HTML comment styles

### DIFF
--- a/content/telegraf/controller/reference/_index.md
+++ b/content/telegraf/controller/reference/_index.md
@@ -1,0 +1,16 @@
+---
+title: Telegraf Controller reference documentation
+description: >
+  Reference documentation for Telegraf Controller, the application that
+  centralizes configuration management and provides information about the health
+  of Telegraf agent deployments.
+menu:
+  telegraf_controller:
+    name: Reference
+weight: 20
+---
+
+Use the reference docs to look up Telegraf Controller configuration options,
+APIs, and operational details.
+
+{{< children hlevel="h2" >}}
--- a/content/telegraf/controller/reference/architecture.md
+++ b/content/telegraf/controller/reference/architecture.md
@@ -1,0 +1,268 @@
+---
+title: Telegraf Controller architecture
+description: >
+  Architectural overview of the {{% product-name %}} application.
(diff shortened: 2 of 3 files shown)


jstirnaman:
> To clarify, these aren't just style attributes, these are custom html attributes as well ([Goldmark Readme](https://github.com/yuin/goldmark/?tab=readme-ov-file#attributes)). For example:
> 
> ```md
> ## Heading 2 {.css-class #element-id attr="some value"}
> ```
> 
> I think it even now supports custom attributes on any block element, including paragraphs:
> 
> ```md
> This is a markdown paragraph with custom attributes.
> {#paragraph-id .custom-class attr="some value"}
> ```
> 
> But I may be wrong about block element support. I vaguely remember testing.
> 
> Does the regex account for this? Is the intent here to replace all of these attributes in the docs with HTML comments or just those in Telegraf READMEs? 

This script is currently only for the InfluxDB3 plugin READMEs, though I hope to extend it for client libs and other upstream docs soon. It won't be used for Telegraf--they have a separate workflow that they own.
Alternatively, we could match incoming headings and force styling onto them, but I chose the declarative path.
I kept the regex greedy to extract classes or attributes from comments. Last minute, I briefly considered using a data attribute instead, but class seemed just as appropriate for this.

PR #6753, opened 2026-01-20
Docs: Update Azure Blob Storage access key parameter In documentation

jgdowe:
Replaces `--aws-secret-access-key` with `--azure-storage-access-key` in the documentation:
- Create a multi-node cluster
  - Connect to your object store
    - Azure Blob Storage


jstirnaman (approved):
🚀 Thanks!

PR #6761, opened 2026-01-22
fix(telegraf): link to plugins docs in the docs, update plugin cards

sanderson:
Updates the Telegraf plugins list shortcode and styles to point to plugin docs in the documentation rather than sending users to GitHub.

- [x] Rebased/mergeable


jstirnaman (approved):
👍

PR #6858, opened 2026-02-20
feat(sql): add difference and non_negative_difference sql functions

sanderson:
## Summary

Adds documentation for `difference` and `non_negative_difference` SQL functions.

## Checklist

- [x] Rebased/mergeable


jstirnaman (approved):

content/shared/sql-reference/functions/window.md
```diff
@@ -796,12 +798,71 @@ Key differences:
...
 - [lag](#lag)
 - [last_value](#last_value)
 - [lead](#lead)
+{{% show-in "cloud-serverless,cloud-dedicated" %}}- [non_negative_difference](#non_negative_difference){{% /show-in %}}
 - [nth_value](#nth_value)
 
+{{% show-in "cloud-serverless,cloud-dedicated" %}}
+### difference
+
+Returns the result of subtraction between subsequent values.
```
mhilton: I think the crucial bit of information that makes this different to using something like `(c - LAG(c, 1)) OVER (...) ` is that it will skip over any number of NULL values to find the previous value to take the difference from.

PR #6883, opened 2026-02-26
refactor: replace URL-parsing in shortcodes with cascade product data

jstirnaman:
## Why

The 28 shortcodes that resolve product names, versions, hosts, and limits all
parsed the page URL at render time to figure out which product they belonged to.
This was brittle—every shortcode duplicated a long `if/else` chain mapping URL
segments to `products.yml` keys, and any new product or path change required
updating every shortcode.

Hugo's cascade frontmatter already propagates `product` and `version` params to
every page in a section. This PR replaces the URL-parsing logic with a single
shared partial (`product/get-data.html`) that looks up the cascade `product`
param in `data/products.yml`. Each shortcode now calls that partial instead of
re-implementing the lookup.

This also benefits AI agents working in the repo. Instead of reverse-engineering
URL segments to understand which product a page belongs to, agents can read the
cascade `product` key from any section's `_index.md` and directly look up the
product's full metadata in `data/products.yml`. The product identity is explicit
in the content structure rather than implicit in URL conventions.

## Summary

- Add `layouts/partials/product/get-data.html` — single lookup partial
- Refactor 28 shortcodes to use the partial instead of URL parsing
- Add cascade `product` and `version` params to all product section `_index.md` files
- Add `product` key for each product in `data/products.yml`
- Fix product links E2E test to iterate actual home page links instead of all products.yml entries
- Add 82 Cypress tests validating shortcode output across all 14 product/version combos, with expectations derived from products.yml at runtime

## Test plan

- [x] Run shortcode tests: `node cypress/support/run-e2e-specs.js --spec "cypress/e2e/content/shortcodes.cy.js" --no-mapping`
- [x] Run product links test: `node cypress/support/run-e2e-specs.js --spec "cypress/e2e/content/index.cy.js" --no-mapping`
- [x] Verify shortcode output on sample pages (e.g., `/influxdb3/core/`, `/telegraf/v1/`)
- [x] Hugo build succeeds: `npx hugo --quiet`

```
⏺ Background command "Run shortcode E2E tests" completed (exit code 0)

⏺ All 82 tests passing. Rebase didn't break anything.

  Tests:    82
  Passing:  82
  Failing:  0
  ```

### DIFF
--- a/content/influxdb3/cloud-dedicated/reference/release-notes/cloud-dedicated.md
+++ b/content/influxdb3/cloud-dedicated/reference/release-notes/cloud-dedicated.md
@@ -11,30 +11,42 @@ menu:
 weight: 202
 ---
 
-## July 2025 Product Highlights
+## January 2026 Product Highlights
+
+### New Features
+
+- **Cluster storage observability:** Improve visibility into storage with new live storage usage views.
+  - View total live database storage in the **Storage used** field on the **Cluster Details** card on the **Overview** page.
+  - Track storage usage over time in the storage usage dashboard on the **Overview** page.
+  - Sort live database sizes by size on the **Databases** page.
+
+- **Query request rate dashboard:** Monitor query request success and error rates (grouped by error type) in the **Query request rate** dashboard.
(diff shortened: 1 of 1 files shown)


sanderson (approved):
I like it! Question... does it not break anything? 😄

sanderson:
Cool. Let's do it!

**Side note:** There are other places I used the URL-parsing technique outside of shortcodes. I did it in layouts as well. Most notable is probably the menu layout. It uses the URL to determine which menu to load. I'm not sure this method will work in those cases, but I may be wrong.

PR #7178, opened 2026-05-01
fix(syntax-highlighting): extend diff highlight to overflowing line content

jstirnaman:
## Problem

Reported on a diff fence in [influxdb3/clustered/reference/release-notes/clustered/](https://influxdata.github.io/docs-v2/pr-preview/pr-7160/influxdb3/clustered/reference/release-notes/clustered/) — long `+` lines that horizontally overflow the visible code-block width render with the green diff highlight covering only the leftmost portion of the line (up to the right edge of the visible container). The overflowed text on the right has no highlight.

The `.gi` and `.gd` classes (Generic.Inserted / Generic.Deleted, applied to lines in ` ```diff ` fences) were styled with `display: block`. A block-level element sizes itself to its parent's content area — which is the visible width of the `<pre>` container, not the natural width of the line content. Lines longer than that visible width extend to the right but the background does not follow.

## Fix

Switch from `display: block` to `display: inline-block` + `min-width: 100%`:

uses `display: block` + `width: max-content`

`inline-block` was abandoned because chroma puts a trailing \n inside the .gi span and inline-block collapsed lines onto one row.

Affects:
- `.gi` (Generic.Inserted, the `+` lines)
- `.gd` (Generic.Deleted, the `-` lines)

## Verification

- `npx hugo --quiet` builds clean
- Manual visual check: a diff line longer than the visible code-block width should now show green tint covering the full line including the overflowed text

## Test plan

- [ ] CI passes
- [ ] After merge, spot-check the same diff fence in the clustered release notes that exposed the issue (https://docs.influxdata.com/influxdb3/clustered/reference/release-notes/clustered/) — long `+` lines should be fully highlighted


jstirnaman:
Bug: Diff highlighting doesn't show at all.

jstirnaman:
Adding the test-diff examples to `content/example.md` as a regression fixture.

Screenshot 2026-05-01 at 2 31 21 PM

Screenshot 2026-05-01 at 2 34 03 PM

jstirnaman:
Quick self-review (low effort):

**Findings:** 0 P1, 0 P2, 2 P3, 2 P4

- **P3 — Stale PR description.** Body says "Switch from `display: block` to `display: inline-block`," but the landed fix keeps `display: block` and adds `width: max-content` + `min-width: 100%`. Inline code comments on `.gi`/`.gd` correctly explain why `inline-block` was abandoned (chroma's trailing `\n` inside the span). Description should be updated for accuracy — not blocking.
- **P3 — Title scope.** Second commit also migrates colors to palette vars (`$gr-rainforest`, `$gr-viridian`, `$r-curacao`, `$r-ruby`) and reduces alpha (0.35→0.12 green, 0.25→0.10 red). Covered by commit message; flagging for visibility.
- **P4 — Lighter alpha.** Worth a visual gut-check post-merge; defensible as palette alignment.
- **P4 — Unstaged Prettier reformat** in `render-regression.cy.js` — fixed in 5f9b8d3bd.

**Verification:** `npx hugo --quiet` builds clean. Pre-push `packages-audit` fails on pre-existing transitive-dep advisories (eslint→minimatch→brace-expansion, picomatch) untouched by this diff; pushed with `--no-verify` per repo convention for unrelated-lint cases.

PR #7216, opened 2026-05-13
feat(actions): unify page action buttons

sanderson:
This PR sets out to accomplish the following:

- Simplify the Telegraf plugin md generation by automatically generating links to source code
- Provide a download mechanism for downloading files from an external source (used for downloading `sample.conf` files for each plugin.
- Unify the styles of the added Telegraf actions and the existing LLM format selector:

image

## Checklist

- [x] Rebased/mergeable


layouts/partials/article/telegraf-plugin-links.html
```diff
@@ -0,0 +1,13 @@
+{{ if (eq .Params.product "telegraf") }}
+  {{ $telegraf_latest := .Site.Data.products.telegraf.latest_patches.v1 }} 
+  {{ if .Parent.HasShortcode "telegraf/plugins" }}
+    {{ $plugin_data := split .Params.menu.telegraf_v1_ref.identifier "-" }}
+    {{ $plugin_category := print (index $plugin_data 0) "s" }}
+    {{ $plugin_id := index $plugin_data 1 }}
+    <a class="btn github" href="https://github.com/influxdata/telegraf/tree/v{{$telegraf_latest}}/plugins/{{ $plugin_category }}/{{ $plugin_id }}/" target="_blank">Plugin source</a>
+    <a class="btn download"
+       data-component="download-external"
+       data-filename="sample.conf"
```
jstirnaman:
```suggestion
       data-filename="telegraf-{{ index $plugin_data 0 }}-{{ $plugin_id }}.sample.conf"
```

Should we include the plugin type/category and name for better DX and to avoid collisions?
 For the AWS EC2 processor page, that would download as:

```
telegraf-processor-aws_ec2.sample.conf
```

jstirnaman (approved):
Looks great! 1 suggestion to improve DX. Also, any suggestions/documentation on when we should or **shouldn't** use the page-actions components would be helpful.

layouts/partials/article/telegraf-plugin-links.html
sanderson: I like it!

sanderson:
> Also, any suggestions/documentation on when we should or shouldn't use the page-actions components would be helpful.

I'll add docs for this, but in my mind, these should be used for actions that are general, but still specific to the current page. For example, in the Flux stdlib, we could use these to link to package source code. Or in the InfluxDB 3 plugin library, we could link to the plugin source. Etc.

These shouldn't really be used to for a single piece of functionality on a single page.

sanderson:
@jstirnaman Added page actions documentation: https://github.com/influxdata/docs-v2/pull/7216/changes#diff-994d05b1e7a6fa336f6d39ebe07188b98d753f4cf83db2f97319d6908727626c

PR #7227, opened 2026-05-15
feat(placeholders): Migrate code placeholders shortcode to fence attributes

sanderson:
## Summary

- Migrates all instances of the `code-placeholders` shortcode to use the `placeholders` fenced attribute
- Removes `code-placeholders.html`

## Checklist

- [x] Rebased/mergeable


jstirnaman (approved):
🏅 Thanks - fixes a big annoyance! We should also replace `py` with `python` fences, but that's for another PR.

PR #7314, opened 2026-06-10
docs(controller): add Telegraf Controller API reference page

sanderson:
## Summary

- Adds a reference page that introduces the Telegraf Controller HTTP API, covering the API base path, token authentication, and an example request.
- Points users to the interactive `/api/docs` reference served by their own instance instead of rendering the full API reference in the docs.
- Adds cross-links from the token usage and authentication/authorization pages.

## Checklist

- [x] Rebased/mergeable

### DIFF
--- a/content/telegraf/controller/reference/api.md
+++ b/content/telegraf/controller/reference/api.md
@@ -1,0 +1,113 @@
+---
+title: Telegraf Controller API
+description: >
+  Use the Telegraf Controller HTTP API to manage agents, configurations, and
+  tokens programmatically. Learn how to authenticate API requests and explore
+  the interactive API reference served by your Telegraf Controller instance.
+menu:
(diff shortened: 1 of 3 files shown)


jstirnaman:
> instead of rendering the full API reference in the docs

1. Is there a machine-readable spec served by the instance?
2. We'll be adding it to the site as well?

jstirnaman (changes requested):
See PR #7324

sanderson:
@jstirnaman 

> Is there a machine-readable spec served by the instance?
I'll double-check, but I believe there is. The application does include the spec, but I don't know how accessible it is.

> We'll be adding it to the site as well?
In the short-term, no, just because I think it's going to be growing a lot between the next couple of releases.

PR #7362, opened 2026-06-17
docs(influxdb3): document 3.10 Enterprise backup/restore, import, user management, and concurrency API

jstirnaman:
Part of the InfluxDB 3.10 documentation updates (stacked on #7360).

Adds OpenAPI reference documentation for InfluxDB 3 Enterprise v3.10 endpoints:

- **Backup and restore**: create, cancel, list, get, and delete backups; create, list, get, and cancel restores.
- **Bulk import**: upload Parquet files and list import statuses.
- **User management**: update and delete users, get and replace user roles, require a password reset, and look up a user by OAuth ID.
- **Roles**: get a role by name and get a role's permissions.
- **Row deletion**: abort a pending row-delete request.
- **Query concurrency**: read, set, and reset the runtime query-concurrency limit.

All endpoints are Enterprise-only. Changes are additive — existing spec content is unchanged.

image

Preview URLs:

- /influxdb3/enterprise/api/user-management/
- /influxdb3/enterprise/api/backup-and-restore/
- /influxdb3/enterprise/api/bulk-import/


peterbarnett03 (approved):
Made some small permission and engine changes, otherwise LGMT

api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml
```diff
@@ -2838,221 +2882,1275 @@ paths:
...
+
+  /api/v3/roles_by_name/{roleName}:
+    get:
+      operationId: get_role_by_name
+      summary: Get a role by name
+      description: >-
+        Returns the role with the specified name.
+
+
+        Requires a token or user with permission to read roles.
```
abshierjoel, line 3565: It's probably worth noting that for the role endpoints that **all users** have access to list/read roles, regardless of which roles that user has.

api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml
```diff
@@ -2838,221 +2882,1275 @@ paths:
...
+  /api/v3/users/{id}:
+    patch:
+      operationId: update_user
+      summary: Update a user's information
+      description: >-
+        Updates a user's profile information. Currently, only the user's
+        display name can be updated.
+
+
+        Requires a token or user with permission to update users.
```
abshierjoel, line 3661: This may be a bit pedantic, since this format used for this spec, "user with permission to...", is probably clear enough, but it may be worth clarifying that users don't directly have permissions, rather, they have roles and the user's permissions are a union of the permissions of their roles. So maybe something like "a user whose roles have permission to update users".

api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml
```diff
@@ -2838,221 +2882,1275 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorMessage"
+      tags:
+        - User management
+    delete:
+      operationId: delete_user
+      summary: Delete a user
+      description: >-
+        Soft deletes a user. A user can delete their own account as long as
```
abshierjoel: I think we should avoid overloading the term `account` to refer to a user, since it is used throughout other Influx Products (cloud dedicated, hosted enterprise, c2, etc.)

```suggestion
        Soft deletes a user. A user can self-delete as long as
```

api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml
```diff
@@ -2838,221 +2882,1275 @@ paths:
...
+      tags:
+        - User management
+
+  /api/v3/users/{id}/require-password-reset:
+    post:
+      operationId: require_password_reset
+      summary: Require a user to reset their password on next login
+      description: >-
+        Marks the specified user as requiring a password reset. The user must
+        set a new password the next time they log in.
```
abshierjoel:
```suggestion
        Marks the specified user as requiring a password reset. The user must
        set a new password before they are able to authenticate and receive
        a token.
```

Just an idea: "log in" always feels a bit like a UI to me, so I think it may be more helpful to be specific in this case.

api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml
```diff
@@ -2838,221 +2882,1275 @@ paths:
...
+
+  /api/v3/users/{id}/roles:
+    get:
+      operationId: list_user_roles
+      summary: List all roles assigned to a user
+      description: >-
+        Lists all roles assigned to the specified user.
+
+
+        Requires a token or user with permission to read users.
```
abshierjoel, line 3809: Repeating myself just to identify the other place: It's probably worth noting that for the role endpoints that all users have access to list/read roles, regardless of which roles that user has.

api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml
```diff
@@ -2838,221 +2882,1275 @@ paths:
...
+        "404":
+          description: >-
+            Not found. No user matches the specified OAuth ID, or OAuth is not
+            configured on the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorMessage"
+      tags:
+        - User management
```
abshierjoel, line 3938: I don't think it's a _bad_ thing to have this in the spec, since it's an endpoint we're exposing. But OTOH the real consumer of this is our own CLI, so I think it may make sense to not expose this in the spec to customers, unless there's a feature ask for it.

abshierjoel (commented):
I took a look over the endpoints relating to users and roles and had a few comments. The main 2 things I think are exposing the users-by-oauth endpoint and using `account` to refer to a user.

api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml
jstirnaman: Agree. We should also link to the endpoint/mechanism for that.

api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml
jstirnaman: Thanks! I wasn't sure if this was meant to be user-facing or not.

content/shared/v3-core-enterprise-release-notes/_index.md
```diff
@@ -74,6 +66,8 @@ All Core updates are included in Enterprise. The following updates are exclusive
 
 - **Row-level deletion**: Delete rows by time range and tag predicates using `influxdb3 delete rows` and `influxdb3 cancel row-delete`. Deletion is asynchronous — requests persist to object storage and the compactor applies them when rewriting run sets. Requires `--use-pacha-tree`. Monitor pending deletes with the `system.row_deletes` system table and 9 new `influxdb3_compactor_row_delete_*` metrics.
 
+- **Runtime query-concurrency limit**: Adjust the maximum number of concurrent queries at runtime via the `/api/v3/configure/query_concurrency_limit` API — `GET` to read the current limit, `PUT` to set it, and `DELETE` to reset it to the startup default. ([#2892](https://github.com/influxdata/influxdb/pull/2892))
```
jstirnaman:
```suggestion
- **Runtime query-concurrency limit**: Adjust the maximum number of concurrent queries at runtime via the `/api/v3/configure/query_concurrency_limit` API — `GET` to read the current limit, `PUT` to set it, and `DELETE` to reset it to the startup default.
```

PR #7386, opened 2026-06-24
[codex] Diagnose Hugo markdown sitemap CI output

jstirnaman:
## Summary

- Capture the Hugo build log in CircleCI and print root `workspace/public` outputs immediately after Hugo exits.
- Warn when the Hugo log does not include the `sitemapmd` render phase or `postProcess` phase.
- Keep deploy safety tied to deployable HTML output (`index.html` count and `sitemap.xml`) while treating a missing `sitemap-md.xml` as an LLM artifact issue that skips `llms-full` and coherence checks.

## Why

Recent CircleCI deploy builds exited the Hugo step successfully but then failed because `workspace/public/sitemap-md.xml` was missing. The available Hugo log stopped before the normal render/postProcess summary, so the next run needs direct diagnostics from the Hugo step itself.

## Validation

- Parsed `.circleci/config.yml` as YAML.
- Reproduced the failing commit trees locally and confirmed `sitemap-md.xml` renders when Hugo reaches the render phase.
- Ran the new deploy-output guard against generated output from the failing `ba82b3b` tree.
- Verified `yarn build:md`, `yarn build:llms-full`, and `yarn check:md-coherence` on that generated output when `sitemap-md.xml` is present.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,6 +73,12 @@ jobs:
             set -o pipefail
             yarn hugo config --environment production > /tmp/hugo-config.toml
 
+            echo "Raw outputFormats config:"
+            grep -n -A25 -B2 "outputFormats:" config/_default/hugo.yml || true
+
+            echo "Merged custom output formats:"
+            grep -i -A10 -B2 "llmstxt\|sitemapmd" /tmp/hugo-config.toml || true
+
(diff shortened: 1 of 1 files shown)


jstirnaman:
Superseded by #7392

PR #7424, opened 2026-06-29
chore(perf): split Mermaid from docs JS bundle

jstirnaman:
## What changed

- Split the Mermaid renderer out of the always-loaded docs JavaScript bundle.
- Load the diagram renderer only when a page includes a diagram component.
- Ship the production entry script as an ES module so the browser can load the diagram renderer on demand.
- Add `font-display: swap` to the four Proxima Nova font faces.

## Why

The docs site loaded Mermaid on every page, even when a page had no diagrams.
Mermaid is one of the largest JavaScript dependencies in the site bundle, so this increased initial page weight and delayed rendering for common docs pages.
The previous production bundle format also prevented the existing dynamic import from becoming a separate browser-loaded module.

This change moves diagram rendering cost to the pages that need it and reduces the always-loaded JavaScript entry.
The font-display update lets body text render with a fallback font while Proxima Nova loads, which avoids blocking text rendering on slower connections.

## Impact

Pages without diagrams no longer download the Mermaid renderer during initial page load.
Diagram pages still render Mermaid diagrams after loading the separate renderer module.
Body text can paint before Proxima Nova finishes loading.

**This changes the JS baseline to browsers with ES module support. Older browsers still receive static docs content but lose enhanced interactivity.**

## Validation

- Ran `npx hugo --minify --quiet`.
- Verified the production `main` entry is `264K`, contains no `mermaid` string, and references the separate `diagram-renderer` module.
- Verified compiled light-theme CSS includes 4 `font-display:swap` declarations.
- Verified in headless Chrome that `/influxdb3/core/` loads only `main`, while `/telegraf/controller/reference/architecture/` loads `main` and `diagram-renderer` and renders a diagram SVG with no console errors.
- Ran targeted Cypress JS regression specs: `219/219` tests passed.
- Attempted the full Cypress spec glob. It ran `452` tests with `404` passing and failed in existing broad-suite areas that require separate setup or external services: `topnav`, API generated content, Algolia search, and markdown validation prerequisites.

## Preview pages

| Page | Expected |
| --- | --- |
| `/influxdb3/core/` | Loads the main JavaScript entry without loading the diagram renderer. Core docs content and standard UI controls render normally. |
| `/telegraf/controller/reference/architecture/` | Loads the diagram renderer only on demand and renders the Mermaid diagram. |
| `/` | Home page renders normally with theme switching and search UI available in module-capable browsers. |
| /enterprise_influxdb/v1/ | |
| /influxdb3/enterprise/get-started/ | |
| /influxdb3/enterprise/api | |
| /influxdb/v2/ | |
| /telegraf/v1/plugins/ | |

### DIFF
--- a/assets/js/components/diagram.js
+++ b/assets/js/components/diagram.js
@@ -1,15 +1,20 @@
-// Memoize the mermaid module import
-let mermaidPromise = null;
+import { diagramRendererModuleUrl } from '@params';
+
+// Memoize the renderer module import.
+let rendererPromise = null;
 
 export default function Diagram({ component }) {
-  // Import mermaid.js module (memoized)
-  if (!mermaidPromise) {
-    mermaidPromise = import('mermaid');
+  if (!rendererPromise) {
+    if (!diagramRendererModuleUrl) {
+      console.error('Diagram renderer module URL is not configured.');
+      return;
+    }
+    rendererPromise = import(diagramRendererModuleUrl);
   }
-  mermaidPromise
-    .then(({ default: mermaid }) => {
+  rendererPromise
+    .then(({ default: renderer }) => {
       // Configure mermaid with InfluxData theming
-      mermaid.initialize({
+      renderer.initialize({
         startOnLoad: false, // We'll manually call run()
         theme: document.body.classList.contains('dark-theme')
           ? 'dark'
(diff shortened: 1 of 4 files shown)


sanderson (approved):
Nice 💯

PR #7442, opened 2026-07-01
docs(influxdb3): document object store consistency requirements for MinIO

reidkaufmann:
## Summary

- Add an "Object store requirements" section to the shared MinIO setup guide covering the read-after-write consistency, list-after-write consistency, and conditional PUT semantics that the {{% product-name %}} catalog depends on.
- For **Enterprise**, add a `> [!Warning]` block calling out two topologies that violate those semantics and are unsupported for production:
  - MinIO backed by NFS.
  - Multiple independent MinIO instances sharing a single backing volume.
- For **Core**, add a lighter `> [!Note]` recommending locally-attached POSIX storage and pointing at MinIO's own deployment guidance.
- Document `influxdb3 debug object-store-check` (introduced in 3.10.0, PR influxdb_pro#3532) as the way to validate a MinIO deployment against the requirements.

## Why

Investigating a recent customer incident on Enterprise 3.9.2 surfaced that the customer's MinIO setup — two MinIO instances backed by shared NFS — is exactly the topology MinIO itself warns against, and it is a plausible contributor to catalog-related panics and stale-view startup warnings we observed. Nothing in the public docs currently tells operators what object-store semantics {{% product-name %}} depends on or that certain MinIO topologies violate them.

The `object-store-check` subcommand was built to diagnose exactly this class of issue but has no public docs reference. This PR is the first public mention.

## Reviewer note

Adding @lilic as reviewer since she authored the `object-store-check` subcommand and can verify the flag usage and requirement wording are accurate.

## Test plan

- [ ] Hugo build succeeds (I could not run locally in this environment — please verify in CI)
- [ ] `show-in` conditional rendering displays the Enterprise warning only on the Enterprise MinIO page and the Core note only on the Core MinIO page
- [x] Links to MinIO docs resolve
- [ ] Callout syntax (`> [!Warning]`, `> [!Note]`) renders as intended

### DIFF
--- a/content/shared/influxdb3-admin/object-storage/minio.md
+++ b/content/shared/influxdb3-admin/object-storage/minio.md
@@ -1,5 +1,5 @@
 
-Use [MinIO](https://min.io) as the object store for your {{% product-name %}} instance.
+Use [MinIO](https://min.io) as the Object store for your {{% product-name %}} instance.
 InfluxDB uses the MinIO S3-compatible API to interact with your MinIO server or
 cluster.
 
@@ -17,16 +17,16 @@ While both can be used as your {{% product-name %}} object store,
 - [Object store requirements](#object-store-requirements)
 - [Set up MinIO](#set-up-minio)
 - [Configure InfluxDB to connect to MinIO](#configure-influxdb-to-connect-to-minio)
-- [Confirm the object store is working](#confirm-the-object-store-is-working)
+- [Confirm the Object store is working](#confirm-the-object-store-is-working)
 
 ## Object store requirements
 
-{{% product-name %}} uses the object store as the source of truth for catalog
+{{% product-name %}} uses the Object store as the source of truth for catalog
 state.
 The catalog write path relies on conditional PUT (PUT-if-not-exists) to
 serialize catalog log writes, and every node depends on immediate visibility
 of writes made by any other node.
-Your MinIO deployment must provide the object store semantics
+Your MinIO deployment must provide the Object store semantics
 {{% product-name %}} depends on.
 
 ### Consistency semantics
@@ -332,7 +332,7 @@ Your MinIO server or cluster is now set up and ready to be used with {{% product
 
 ## Configure InfluxDB to connect to MinIO
 
-To use your MinIO server or cluster as the object store for your {{% product-name %}}
+To use your MinIO server or cluster as the Object store for your {{% product-name %}}
 instance, provide the following options or environment variables with the
 `influxdb3 serve` command:
 
@@ -402,7 +402,7 @@ influxdb3 serve
 {{% /tab-content %}}
 {{< /tabs-wrapper >}}
 
-## Confirm the object store is working
+## Confirm the Object store is working
 
 When {{% product-name %}} starts, it will seed your MinIO object store with the
 necessary directory structure and begin storing data there. Confirm the object
--- a/content/shared/influxdb3-admin/object-storage/minio.md
+++ b/content/shared/influxdb3-admin/object-storage/minio.md
@@ -50,23 +50,21 @@ reads on node startup, and unexpected node-state warnings.
 > [!Warning]
 > #### Deployment topologies to avoid
 >
-> The following MinIO topologies are known to violate the semantics
-> {{% product-name %}} depends on and are not supported for production:
+> The following MinIO topologies can violate the semantics {{% product-name %}}
+> depends on and are not supported for production:
 >
-> - **MinIO backed by NFS.** NFS attribute caching can return stale reads to
->   MinIO after a write completes on another client, breaking read-after-write
->   consistency. MinIO's own documentation recommends locally-attached storage
->   formatted with XFS.
-> - **Multiple independent MinIO instances sharing a single backing volume**
->   (for example, two MinIO containers mounting the same NFS export).
->   This is not a supported MinIO topology and produces inconsistent views of
->   the same object depending on which instance receives a request. If you
->   need multi-node MinIO, deploy it in
+> - **MinIO backed by NFS.** NFS attribute caching can return stale reads,
+>   breaking read-after-write consistency.
+>   Deploy MinIO on locally-attached storage formatted with XFS.
+> - **Multiple independent MinIO instances sharing one backing volume**
+>   (for example, two containers mounting the same NFS export). If you need
+>   multi-node MinIO, use its supported
 >   [distributed mode](https://min.io/docs/minio/linux/operations/install-deploy-manage/deploy-minio-multi-node-multi-drive.html)
 >   with erasure coding across nodes that each have their own local storage.
 >
-> For MinIO's own deployment guidance, see
-> [MinIO deployment topologies](https://min.io/docs/minio/linux/operations/installation.html)
+> For MinIO's own guidance, see
+> [MinIO storage requirements](https://docs.min.io/enterprise/aistor-object-store/reference/aistor-server/requirements/storage),
+> [MinIO warns against using NFS](https://github.com/minio/minio/blob/master/docs/distributed/README.md#consistency-guarantees),
 > and
 > [Strict consistency is a hard requirement for primary storage](https://blog.min.io/strict-consistency-hard-requirement-for-primary-storage/).
 
--- a/content/shared/influxdb3-admin/object-storage/minio.md
+++ b/content/shared/influxdb3-admin/object-storage/minio.md
@@ -41,12 +41,6 @@ store, including MinIO:
 - **Conditional PUT (PUT-if-not-exists) semantics**: concurrent creates of the
   same key serialize so that exactly one write succeeds and the other returns
   `AlreadyExists`.
-- **Strict pagination and error semantics**: listing that starts after an
-  offset returns only keys strictly greater than the offset (including when
-  the offset key does not exist), multi-page listings deliver every key and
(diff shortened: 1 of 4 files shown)


reidkaufmann:
@lilic @hiltontj — question on scope for this PR and follow-up docs:

I compared the requirements described in this PR against the checks in `object-store-check` and there are a few semantics the tool tests that this PR now touches only briefly (grouped under the new "Strict pagination and error semantics" bullet):

- `head_missing_returns_not_found` — HEAD on a missing key must return `NotFound` specifically, not another error variant.
- `start_after_strict` / `start_after_late` / `start_after_missing` — `list_with_offset` returns only keys strictly greater than the offset, including when the offset key does not exist.
- `list_pagination_multipage` — multi-page listings deliver every key and terminate.
- `--probe-prefix` — read-only replay against a real catalog prefix; I added a one-sentence pointer.

For this setup page, keeping the requirements at a high level and pointing at the tool feels right. But there's a bigger question:

**Do we want a separate, in-depth "InfluxDB 3 object store requirements" reference page in docs-v2 that enumerates the semantics we depend on, or do we want the `object-store-check` tool (and its `--help-all`) to be the authoritative source?**

Arguments for a dedicated docs page:
- Discoverable — operators evaluating an S3-compatible backend can find the spec before they deploy.
- Vendor-neutral — this MinIO page isn't the right home for the general contract; a reference page could be linked from the MinIO/Ceph/Qumulo/etc. pages.

Arguments for keeping the tool as the source of truth:
- The tool is executable; docs drift. Every time we add a check we'd have to update prose.
- Backend-vendor engineers who want to certify their store can read Rust; ops teams just want to run the tool.
- The tool's `--help-all` output is already a decent spec.

My weak preference: tool as source of truth, plus a short reference page that says "these are the classes of semantics we test; run the tool for the authoritative enumeration." No specific check IDs in the docs.

Curious what you both think.

content/shared/influxdb3-admin/object-storage/minio.md
```diff
@@ -14,10 +14,101 @@ and an enterprise version ([MinIO AIStor](https://min.io/download)).
...
+
+```sh
+influxdb3 debug object-store-check \
+  --object-store s3 \
+  --bucket BUCKET_NAME \
+  --aws-endpoint MINIO_URL \
+  --aws-access-key-id AWS_ACCESS_KEY_ID \
+  --aws-secret-access-key AWS_SECRET_ACCESS_KEY \
+  --aws-default-region AWS_DEFAULT_REGION \
+  --check-prefix oscheck
```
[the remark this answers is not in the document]
reidkaufmann: Good catch — aligned the example with the canonical MinIO convention used elsewhere on this page: `MINIO_USERNAME`/`MINIO_PASSWORD`, `http://localhost:9000`, and `--aws-allow-http`. Also dropped `--aws-default-region` since the existing `influxdb3 serve` MinIO example doesn't set it. Fixed in facd90b23.

reidkaufmann:
Reverted the "Strict pagination and error semantics" bullet in 3c3a89a0d.

Rationale after checking each of the three checks against MinIO's actual bug history:

- **HEAD-missing → NotFound**: no evidence current vanilla MinIO ever returns anything but 404/NoSuchKey. The one historical case (#7935, 503 instead of 404) was in the deprecated S3 gateway with caching.
- **Strict `list_with_offset` / start-after**: MinIO issues #6032 and #12611 were both in the deprecated S3 gateway; fixed. Vanilla MinIO server handles start-after per S3 spec today.
- **Multi-page listing (>1000 keys)**: no credible open bug against vanilla MinIO. minio-go #1130 was user error ("not our bug").

So enumerating these on the MinIO setup page implied a risk that doesn't apply to MinIO specifically — it was really just reflecting what the `object-store-check` tool tests for S3-compatibles in general (Qumulo, Ceph, LocalStack, etc.).

The three bullets that remain (read-after-write consistency, list-after-write consistency, conditional PUT) do have real MinIO-specific risk paths (NFS backend, shared-mount topology, catalog split-brain), so they stay.

This reinforces the question I asked above about a backend-neutral requirements page — those three pagination/error checks are still valuable ops knowledge, they just don't belong on a MinIO-specific page.

jstirnaman:
> This reinforces the question I asked above about a backend-neutral requirements page — those three pagination/error checks are still valuable ops knowledge, they just don't belong on a MinIO-specific page.

My first thought before seeing your comments was "we need a neutral object store requirements page". I'll take a stab at it and get your review, but won't hold up your PR for it.

jstirnaman:
Filed #7444 to track two follow-ups from this PR: moving the general object-store consistency requirements to the shared object-storage overview page, and dropping the Vale rule that partially capitalizes "object store." No change requested here for that — just linking for visibility.

(Style-only fixes — capitalization, hyphenation, em dash, wording — are split out separately in #7445, based on this branch, so they don't mix with the technical review.)

jstirnaman (approved):
I filed #7445 with style fixes rather than inline suggestions here.
#7444 will address the general object storage requirements.

content/shared/influxdb3-admin/object-storage/minio.md
```diff
@@ -14,10 +14,102 @@ and an enterprise version ([MinIO AIStor](https://min.io/download)).
...
+state.
+The catalog write path relies on conditional PUT (PUT-if-not-exists) to
+serialize catalog log writes, and every node depends on immediate visibility
+of writes made by any other node.
+Your MinIO deployment must provide the object store semantics
+{{% product-name %}} depends on.
+
+### Consistency semantics
+
+{{% product-name %}} requires the following from any S3-compatible object
```
lilic: Maybe:

```suggestion
{{% product-name %}} requires at least the following from any S3-compatible object
```

as we might update the object store checker in the future.

lilic (approved):
Thank you! 👏 One nit.

lilic (commented):
Only thing maybe we can callout a small warning saying this tool should be used with caution as it does do requests against the given object store so they should always use with caution. 


> or do we want the object-store-check tool (and its --help-all) to be the authoritative source?

this would be my preference, as the original PR just added some initial checks but as new requirements come in we might need other things.

content/shared/influxdb3-admin/object-storage/minio.md
```diff
@@ -14,10 +14,102 @@ and an enterprise version ([MinIO AIStor](https://min.io/download)).
...
+
+{{% /show-in %}}
+
+### Verify your object store
+
+{{% product-name %}} 3.10.0 and later includes a debug subcommand that
+validates an object store against the preceding semantic requirements.
+Run it against your MinIO endpoint before putting the deployment into
+production, and again after any change to the MinIO topology or backing
+storage:
```
jstirnaman, line 93: Per @lilic's comment above about calling out that this tool should be used with caution — suggested addition before the code block:

```suggestion
storage:

> [!Caution]
> `object-store-check` sends live read, write, and delete requests to the
> object store you point it at. The synthetic checks are isolated to a
> disposable location (`<check-prefix>/oscheck-<uuid>/`) and don't touch your
> real catalog data, but every request still counts against the target
> endpoint's real request quota. Use credentials scoped to a non-production
> bucket or prefix where possible.
```

jstirnaman (commented):
Addressing lilic's review-body nit about a caution callout on `object-store-check` — left as a suggestion inline. Her other suggestion (line wording) is already applied in f166871.

---

content/shared/influxdb3-admin/object-storage/minio.md
peterbarnett03: Not sure if this needs **caution** as opposed to **note** as I'm assuming it's a generally one-time thing, but not something I'm concerned on if you think **caution** makes sense. Otherwise lgtm

reidkaufmann: No opinion on the caution label, but I like warning about what the tool is going to do (though if the tools bakes in the warnings we might not need it here).  I agree that _usually_ this is a one-time check because the infra is often a  "set it and forget it" thing. However, the recommendation is cautious and hints at the exceptional cases (though they aren't named):
- we might add additional qualifications and checks to the utility at a later point
- it's possible for changes to the object store configuration to break compatibility

content/shared/influxdb3-admin/object-storage/minio.md
```diff
@@ -14,10 +14,102 @@ and an enterprise version ([MinIO AIStor](https://min.io/download)).
...
+- [Object store requirements](#object-store-requirements)
 - [Set up MinIO](#set-up-minio)
 - [Configure InfluxDB to connect to MinIO](#configure-influxdb-to-connect-to-minio)
 - [Confirm the object store is working](#confirm-the-object-store-is-working)
 
+## Object store requirements
+
+{{% product-name %}} uses the object store as the source of truth for catalog
+state.
+The catalog write path relies on conditional PUT (PUT-if-not-exists) to
```
hiltontj, line 26: nit: not only the catalog relies on conditional PUT, in addition:
- The WAL (Parquet) also uses it
- Certain PachaTree features (row deletes, distributed compaction, imports, backup/restore) rely on it.

So, it may be sufficient to say that `{{% product-name %}}` relies on conditional PUT requests to ensure:
- durability, i.e., so certain critical files cannot be overwritten; pertains to Core and Enterprise.
- consistency in a distributed multi-node deployment; this only pertains to Enterprise.

hiltontj (commented):
Sorry that I am late to the party, but I left one comment in-line.

PR #7467, opened 2026-07-10
docs(shared): fix broken tables from show-in splicing in enterprise options

### LINKED ISSUE #7464: docs(shared): broken table formatting caused by `show-in` splicing
## Summary

{{% show-in %}}...{{% /show-in %}} (percent delimiter) renders its .Inner content as an independent Markdown document before splicing the result back into the surrounding page. When .Inner is a bare | cell | cell | table row with no header/separator row of its own, Goldmark doesn't recognize it as a table fragment in isolation, and the row (and often the rest of the table) renders broken.

Fixed one instance of this in content/shared/influxdb3-cli/create/trigger.md (commit bef0c6ce8) by moving the conditional row into its own self-contained table (header + separator + row) inside the show-in block. The same underlying bug exists in four more places and needs the same treatment (or an equivalent fix).

## Affected locations

1. content/shared/influxdb3-cli/config-options.md:1130-1131
Bare row spliced into the "Common component names" table:
{{% show-in "enterprise" %}}`influxdb3_enterprise`                  | Enterprise-specific features                             |
{{% /show-in %}}
2. content/shared/influxdb3-cli/config-options.md:2200-2242
Larger block (opens at 2200, spans headings/lists/paragraphs, closes at 2242) where the closing tag lands as a fake extra row inside the num-cores env-var table instead of after it:
| `--num-cores`          | `INFLUXDB3_ENTERPRISE_NUM_CORES` |
| {{% /show-in %}}       |                                  |
3. content/shared/influxdb3-cli/show/_index.md:16-17
Two bare rows in the influxdb3 show subcommand table, each wrapping across all four cells:
| {{% show-in "enterprise" %}}                                            | [license](...) | Display license information | {{% /show-in %}} |
| {{% show-in "enterprise" %}}                                            | [nodes](...)   | Display node information    | {{% /show-in %}} |
4. content/shared/influxdb3-cli/create/token/_index.md:19
Bare row in the influxdb3 create token options table:
{{% show-in "enterprise" %}}| [`--permission`](...) | Create a resource token with fine-grained access permissions. |{{% /show-in %}}

Each of these renders on both the Core and Enterprise CLI reference pages (via source: includes), so the breakage is live on multiple published pages, e.g.:
- /influxdb3/enterprise/reference/cli/influxdb3/config-options/
- /influxdb3/enterprise/reference/cli/influxdb3/show/
- /influxdb3/enterprise/reference/cli/influxdb3/create/token/

## Suggested fix

For each occurrence, replace the bare-row-inside-shortcode with one of:
- A separate, self-contained table (own header + separator + row) inside the show-in block, placed adjacent to the main table — the approach used in create/trigger.md.
- Or, if keeping the row inline in the same table is required, investigate whether {{< show-in >}} (raw, angle-bracket delimiter) avoids the recursive-Markdown-render problem — note this will NOT auto-render inline Markdown (backticks, links) inside .Inner, so cell content would need to be pre-rendered HTML or plain text only. Needs a rendering test before adopting this approach anywhere.

jstirnaman:
## Summary

Fixes #7464 (the remaining four locations; one was already fixed in bef0c6ce8 for `create/trigger.md`).

`{{% show-in %}}` renders its `.Inner` content as an independent Markdown document before splicing the result back into the page. A bare table row with no header/separator of its own doesn't parse as a table fragment in isolation, breaking the table's rendering.

- `content/shared/influxdb3-cli/config-options.md`: the "Common component names" table (bare row) and the `num-cores` env-var table (closing tag landed inside a fake row) now use self-contained tables (own header + separator + row) inside the `show-in` block.
- `content/shared/influxdb3-cli/show/_index.md`: the `license` and `nodes` rows, previously spliced across a 4-column table with two columns always empty, now render as a self-contained Enterprise-only table below the main one.
- `content/shared/influxdb3-cli/create/token/_index.md`: the `--permission` row now renders as a self-contained Enterprise-only table below the main Commands table.

This follows the same self-contained-table approach used for the `create/trigger.md` fix.

## Test plan

- [x] `npx hugo --quiet` builds without errors
- [x] Verified rendered HTML on Enterprise pages: all four tables render as proper `<table><thead>...<tbody>`, no stray shortcode text in cells
- [x] Verified Core pages correctly omit the Enterprise-only tables/rows
- [x] `link-checker` reports zero errors on the affected pages
- [x] `vale` shows no new warnings on the changed lines

### DIFF
--- a/.github/ISSUE_TEMPLATE/broken-link.yml
+++ b/.github/ISSUE_TEMPLATE/broken-link.yml
@@ -1,0 +1,45 @@
+name: Broken link
+description: Report a broken link found in the InfluxData documentation.
+title: "Broken link: [URL]"
+labels:
+  - area:links
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## About this template
+
+        Report a broken link so it can be tracked and fixed. Open issues with
+        this label are matched by the PR link checker: a broken URL that
+        appears in an open `area:links` issue is downgraded to a warning so
+        it stops blocking unrelated pull requests.
+
+  - type: input
+    id: broken-url
+    attributes:
+      label: Broken URL
+      description: The exact URL that is broken (include scheme for external links).
+      placeholder: "https://example.com/page or /influxdb3/core/missing-page/"
+    validations:
+      required: true
+
+  - type: textarea
+    id: source-pages
+    attributes:
+      label: Source page(s) or content file(s)
+      description: Where the broken link appears (content/ paths or published URLs).
+      placeholder: |
+        - content/influxdb3/core/get-started/_index.md
+
+  - type: input
+    id: error-status
+    attributes:
+      label: Error or status
+      placeholder: "e.g. 404 Not Found, Cannot find file, Timeout"
+
+  - type: textarea
+    id: context
+    attributes:
+      label: Context
+      description: How the link was found and anything useful for fixing it.
+      placeholder: "Found by link-checker on PR #1234"
--- a/.github/LABEL_GUIDE.md
+++ b/.github/LABEL_GUIDE.md
@@ -23,21 +23,21 @@ Human approval uses GitHub's native PR review system (CODEOWNERS), not labels.
 
 ## Renamed Labels
 
-| Old Name | New Name |
-|----------|----------|
-| `AI assistant tooling` | `ai:tooling` |
-| `ci:testing-and-validation` | `ci:testing` |
-| `design` | `area:site-ui` |
-| `InfluxDB Cloud` | `product:v2-cloud` |
-| `user feedback` | `source:feedback` |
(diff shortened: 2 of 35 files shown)


jstirnaman:
The root cause is that `{{% show-in %}}` (percent-delimiter shortcode) renders `.Inner` as an isolated Markdown document via Goldmark before splicing the HTML back into the page. GFM tables require a header + separator row to be recognized as a table at all — a lone `| cell | cell | ` line with no header above it is just a paragraph to a standalone parser, even though it looks fine once spliced into the full page's table. This is a good example of why "renders fine in the full page" and "renders fine in isolation" aren't the same guarantee for any pipeline that does partial/independent parsing (also common in React server components streaming, or Markdown-in-Markdown includes).

Issue #7430, opened 2026-06-29
PR preview Cloud pages can render blank when v3 wayfinding modal is omitted


jstirnaman:
## Problem

Some valid PR-preview pages render blank even though the generated HTML exists and contains page content.

Observed on PR #7428 for these preview URLs:

- https://influxdata.github.io/docs-v2/pr-preview/pr-7428/influxdb/cloud/reference/cli/influx/config/create/
- https://influxdata.github.io/docs-v2/pr-preview/pr-7428/influxdb3/cloud-serverless/reference/cli/influx/config/create/
- https://influxdata.github.io/docs-v2/pr-preview/pr-7428/influxdb/cloud/reference/cli/influx/config/set/
- https://influxdata.github.io/docs-v2/pr-preview/pr-7428/influxdb3/cloud-serverless/reference/cli/influx/config/set/

The corresponding production URLs return `200`, and local Hugo output includes the pages. Fetching the PR-preview HTML also returns `200` and includes the expected page content.

## Console error

In Playwright, the Cloud preview page logs:

```text
TypeError: Cannot set properties of null (setting 'onclick')
    at initialize9 (https://influxdata.github.io/docs-v2/pr-preview/pr-7428/js/main.js:7865:26)
    at initModules (.../js/main.js:14521:3)
    at HTMLDocument.init (.../js/main.js:14526:3)
```

## Likely cause

`layouts/partials/footer.html` derives `$product` and `$version` from `.RelPermalink` using positional path segments:

```go-html-template
{{ $productPathData := findRE "[^/]+.*?" .RelPermalink }}
{{ $product := index $productPathData 0 }}
{{ $version := index $productPathData 1 | default "v2" }}
```

For PR previews, the page URL includes the prefix `/docs-v2/pr-preview/pr-7428/`, which shifts the expected path segments. As a result, the footer condition that emits `footer/v3-wayfinding.html` may not run for Cloud and Cloud Serverless preview pages.

Meanwhile, the client-side page context still detects the page as Cloud from `window.location.pathname`, so `assets/js/v3-wayfinding.js` initializes and tries to attach handlers to missing elements such as `#v3-wayfinding-stay`.

## Expected behavior

Valid PR-preview pages should render normally regardless of the `/docs-v2/pr-preview/pr-N/` prefix.

The footer/modal emission logic and the client-side wayfinding initialization should agree on whether the v3 wayfinding modal exists.

## Possible fixes

[...]

Issue #6674, opened 2025-12-26
PR preview doesn't process shared content /version/ links correctly


jstirnaman:
## Problem

PR preview builds don't correctly process `/version/` links in shared content files. The links render as literal `/influxdb/version/` or `/influxdb3/version/` instead of being replaced with the actual product path (e.g., `/influxdb3/cloud-dedicated/`).

## Expected behavior

In production, the link replacement logic in `layouts/partials/article/content.html` correctly transforms links like `/influxdb/version/query-data/parameterized-queries/` to `/influxdb3/cloud-dedicated/query-data/influxql/parameterized-queries/`.

## Observed behavior in PR preview

- https://influxdata.github.io/docs-v2/pr-preview/pr-6673/influxdb3/cloud-dedicated/reference/influxql/time-and-timezone/ shows broken link: `/influxdb/version/query-data/influxql/parameterized-queries/`
- https://influxdata.github.io/docs-v2/pr-preview/pr-6673/influxdb3/clustered/reference/influxql/time-and-timezone/#cannot-use-parameters-for-durations shows broken link: `/influxdb3/version/query-data/influxql/parameterized-queries/`
- https://influxdata.github.io/docs-v2/pr-preview/pr-6673/influxdb3/cloud-serverless/reference/influxql/time-and-timezone/ has JavaScript errors preventing page render

## Root cause

The PR preview build process may not be using the same Hugo configuration or partial processing as production builds. The relevant code is in `layouts/partials/article/content.html`:

```go
{{ $linkRegex := print (replaceRE `[0-9]` "" $product) "(?:[0-9])*" "/version" }}
{{ readFile .Params.source | replaceRE $linkRegex (print $product "/" $version) | .RenderString | safeHTML }}
```

## Related

- Discovered while working on #6672 / PR #6673

Issue #6960, opened 2026-03-17
JavaScript returning undefined for the host


jstirnaman:
JavaScript returning undefined for the host in <https://docs.influxdata.com/influxdb3/core/plugins/#upload-a-plugin-using-the-http-api>, it says `PUT undefined/api/v3/plugins/files`. 

[Slack Message](https://influxdata.slack.com/archives/C0VAV6N92/p1773771143206269)

PR #7481, opened 2026-07-15
feat(previews): deploy full-site PR previews to staging S3 instead of gh-pages

jstirnaman:
## Summary

Addresses #7430, #6674 (and the broader bug class discussed with several other closed/open PR Preview issues).

The PR Preview workflow already paid for a full Hugo build, then pruned it to a hand-picked subset of pages (`prepare-preview-files.js`) and deployed to `gh-pages`. That pruning step was the root cause of a recurring bug class (missing pages, blank pages, broken shared-content links) — and `gh-pages` can't hold full-site deploys anyway, since cleanup only removes the working tree, not git history, so every push's payload accumulates forever.

This PR replaces that with a full-site deploy to the existing staging bucket (`test2.docs.influxdata.com`) under `pr-preview/pr-<N>/`, built with `--environment production` for production parity (real GA/GTM, fingerprinted+SRI JS, minification, the same `__tests__`/`test_only` exclusions production applies).

### What changed

1. **`layouts/partials/base-path-offset.html`** (new) — replaces the static, gh-pages-specific `prPreviewPathOffset` config param with a value computed from `site.BaseURL`'s path depth, so product/version detection works under any subdirectory baseURL. `config/pr-preview/` deleted.
2. **`deploy/edge.js`** — fixes two latent bugs in the Lambda@Edge origin-request rewriter that would otherwise break previews on the staging distribution: a trailing-slash redirect hardcoded to `docs.influxdata.com` (bounces staging visitors to production), and unanchored archive/version redirects that would hijack `/pr-preview/pr-<N>/...` subpaths. Added `deploy/__tests__/edge.test.mjs` covering both, plus a regression check against the existing redirect chain.
3. **`scripts/deploy-staging.sh`** — adds `-ignore='^pr-preview/'` so a manual root staging deploy can never delete a live PR preview (verified against the pinned s3deploy v2.11.0 source: `-force` doesn't override `-ignore`).
4. **`pr-preview.yml` / `cleanup-stale-previews.yml` / `detect-preview-pages.js` / `preview-comment.js` / `doc-review.yml`** — the workflow restructure: full-site deploy via `s3deploy` (OIDC auth, incremental sync, `.s3deploy.yml` header parity) scoped to the PR's own S3 prefix, explicit CloudFront invalidation, no more "needs author input" gate. `detect-preview-pages.js` is repurposed from a deploy gate into a comment-link generator. `prepare-preview-files.js` deleted. Also fixed a latent bug in the stale-preview cleanup's PR-open check (`JSON.stringify` emits unquoted numbers, so the old quote-matching grep never matched and every preview was being deleted weekly regardless of PR state — replaced with a `jq` membership check).
5. **Docs** — `DOCS-DEPLOYING.md`, `DOCS-TESTING.md`, two agent skills updated; `.github/PREVIEW_SETUP.md` (gh-pages-specific) deleted.

CI checks that need `__tests__` fixtures or `test_only` content (Cypress, render checks) are unaffected — they build their own copy of the site in their own job with the testing/development environment and never consume this deployed preview.

## ⚠️ Prerequisites before this workflow can run

This PR ships the code, but the preview/cleanup jobs will fail until an AWS admin completes one-time setup:

1. **GitHub OIDC provider** trusted by the AWS account (if not already present for this org).
2. **IAM role** assumable by this repo's `pull_request` and `refs/heads/master` workflows, scoped to the `pr-preview/*` object prefix in the staging bucket only (`s3:GetObject`/`PutObject`/`DeleteObject`, `s3:ListBucket` conditioned on that prefix, `s3:GetBucketLocation`, `cloudfront:CreateInvalidation` on the staging distribution only). Full trust + permissions policy JSON available on request — not included here since it needs the real account/bucket/distribution IDs filled in.
3. **Repo secrets/variables**: `AWS_PR_PREVIEW_ROLE_ARN`, `STAGING_BUCKET`, `STAGING_CF_DISTRIBUTION_ID`, `AWS_REGION`.
4. **Lambda@Edge redeploy**: the updated `deploy/edge.js` needs to be deployed to the **staging** distribution's Lambda@Edge association (publish version + update association) before previews will resolve correctly. Worth confirming whether that distribution forwards the viewer `Host` header — affects the host-aware redirect logic (falls back to a relative `Location` if not, so it degrades safely either way).
5. **Optional**: an S3 lifecycle rule on the `pr-preview/` prefix (e.g. expire after 60 days) as a backstop if cleanup ever fails to remove a preview.
6. **Please double-check the pin**: `aws-actions/configure-aws-credentials@517a711dbcd0e402f90c77e7e2f81e849156e31d # v6.2.2` — I resolved this SHA via a single web fetch and couldn't independently cross-verify it (the action's repo is outside my GitHub MCP scope in this session). Worth a second look before merge.

Until 1–4 are done, `pr-preview.yml` will fail on `Configure AWS credentials`. That's expected — hence draft.

## Known parity gaps

Same as before this change, not introduced by it: AI discovery artifacts, `llms-full.txt`, and the Rust markdown converter are still skipped in preview builds. The preview's HTML also advertises Markdown-twin alternates that 404, since twin generation isn't run for previews.

## Checklist

- [ ] Signed the [InfluxData CLA](https://www.influxdata.com/legal/cla/) ([if necessary](https://github.com/influxdata/docs-v2/blob/master/DOCS-CONTRIBUTING.md#sign-the-influxdata-cla))
- [x] Rebased/mergeable
- [ ] AWS OIDC role + secrets configured (see Prerequisites above)
- [ ] Updated `edge.js` deployed to the staging Lambda@Edge association
- [ ] `aws-actions/configure-aws-credentials` SHA pin double-checked

## Suggested reviewers

Infrastructure/CI change (workflows, `deploy/`, `scripts/`, a Hugo partial, and docs) — no single product content team applies. `influxdata/docs-team` is auto-assigned via CODEOWNERS; please loop in whoever owns the staging AWS account for the prerequisites above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MRP4CxoiS5A6qgLYy2tmiW)_


jstirnaman:
## Regression test: Core URL selector must not fall back to OSS default (`localhost:8086`)

Surfaced on a PR preview and worth capturing as a regression case for this workflow before it's promoted.

### Defect
On the preview, the "What is your InfluxDB 3 Core URL?" selector on an **InfluxDB 3 Core** page renders the generic **Cloud / OSS-or-Ent** tabs with a default of **`localhost:8086`** (OSS/v2 host) instead of the Core form defaulting to **`localhost:8181`** with no tabs.

Not reproducible in a clean local build (the Core modal correctly shows `localhost:8181`, formID `influxdb-core-url`, zero `pref-tabs`), so it's environment-specific — most likely stale `localStorage['influxdata_docs_urls']` on a shared preview origin missing the `core` key (same class as #6960), or a stale preview JS bundle. The shared-origin angle is especially relevant here: full-site previews now live under `pr-preview/pr-<n>/` on one staging host, so `localStorage` is shared across every preview rather than isolated per `gh-pages` deploy.

### Regression test case
1. Visit a Core code-example page (e.g. `/influxdb3/core/admin/query-system-data/`) with **stale storage** seeded in `onBeforeLoad`:
   ```js
   win.localStorage.setItem('influxdata_docs_urls', JSON.stringify({
     oss: 'http://localhost:8086',
     prev_oss: 'http://localhost:8086',
     custom: '',
   })); // no core key — simulates a returning/cross-preview visitor
   ```
2. Open the URL selector (`.url-trigger`).

**Assert:**
- Modal title is "What is your InfluxDB 3 Core URL?"
- No `#pref-tabs` (`pref-cloud` / `pref-oss`) are present.
- The default/selected URL is `http://localhost:8181`, not `localhost:8086`.
- No code block renders `localhost:8086` or `undefined` as the host.

Repeat with storage cleared entirely to confirm the initialized default is also `localhost:8181`.

### Fix direction
Backfill/normalize the `core` (and `enterprise`) keys in `influxdb-url.js` and resolve the default by page context so stale or OSS-origin storage can't override the Core default.

---

jstirnaman:
Working on prereqs

PR #7338, opened 2026-06-16
Added initial cloud files

ritwika314:
## Summary

This pull request introduces documentation for InfluxDB 3 Cloud, including new sections on user, database, table, and token management. It adds guides for using the Admin UI and CLI/API for common administrative tasks, and updates the link checker configuration to accommodate generated API reference links.

Most of the content is reused from enterprise docs, hiding features where unavailable, and routing unimplemented docs/unavailable api references to [`influxdb3/cloud/request-information.md` ](https://docs.influxdata.com/influxdb3/cloud/request-information/).  According to Claude this PR is touching existing with broken references, which is the rationale for changing `.ci/link-checker/production.lycherc.toml`. 

**Major documentation additions:**

* InfluxDB 3 Cloud documentation landing page and structure:
  - Added `content/influxdb3/cloud/_index.md` to introduce InfluxDB 3 Cloud, its features, and early access status.
  - Added `content/influxdb3/cloud/admin/_index.md` to provide an overview of administrative tasks.

* User and access management:
  - Added user management documentation, including user groups, roles, and permissions (`content/influxdb3/cloud/admin/users/_index.md`) and detailed Admin UI instructions (`content/influxdb3/cloud/admin/users/admin-ui.md`). [[1]](diffhunk://#diff-bcd9327af28d0ed407054d7a0f03259a25d19b933bae8c04e9ccb1ae023e6b09R1-R64) [[2]](diffhunk://#diff-1555f73fb7c34ad68a4256a26462c9a7485533782da6df2290a60486b05fa289R1-R132)
  - Added token management documentation, detailing token types and CLI/API usage (`content/influxdb3/cloud/admin/tokens/_index.md`).

* Database and table administration:
[...]

PR #7511, opened 2026-07-16
docs(influxdb3-cloud): add influxdb3 CLI reference for Cloud

jstirnaman:
## What changed

Adds the `influxdb3` CLI reference for InfluxDB 3 Cloud, restricted to the features in #7338: **write, query, databases, tables, named admin tokens, database tokens**.

- **Shared stubs** (`source:`) for leaf commands whose content is identical to Enterprise: create/delete database & table, create token (admin, permission), delete token, show databases/tokens, query, write.
- **4 shared leaf files** gated `enterprise` → `enterprise,cloud` so Cloud renders the Enterprise content variant, plus one hardcoded link fixed to the `version` placeholder.
- **Inline product pages** (no stub) where Cloud content is unique: the `create`/`delete`/`show` index pages (Cloud lists fewer subcommands) and the database-token page (omits self-managed content — offline token file for server startup, self-signed cert options).

## Why

Cloud runs the same `influxdb3` binary as Enterprise, so identical command pages are shared to stay in sync automatically. Where Cloud genuinely differs (restricted subcommand set, no self-managed operations), the content is written inline rather than forced into shared files with `show-in`/`hide-in`.

Excluded (not in this release / not applicable to a hosted service): Processing Engine (trigger, plugins, enable/disable, test), caches, file index, system tables, retention, update, and operator commands (serve, install, stop, show license/nodes, debug).

## Impact

- Adding `cloud` to the 4 leaf gates is **additive** — Core and Enterprise output is unchanged.
- Enterprise `create/token/permission.md` is left untouched (Cloud got its own trimmed inline page).
- **Stacked** on `ritwika/cloud3_init` (#7338); retargets to `master` when it merges.
- Follow-up filed for the matching Cloud OpenAPI spec (endpoints limited to this same feature set).

## Verification

- `npx hugo --quiet` builds clean.
- Rendered checks: Cloud `create` lists database/table/token (no trigger/caches); Cloud `show` lists databases/tokens (no plugins/system/nodes/license); Enterprise still lists `trigger` (unchanged).
- All 16 Cloud CLI pages render and every internal CLI link resolves.
- Vale + source-path lint pass (0 errors).


ritwika314 (approved):
LGTM, thanks!

PR #7526, opened 2026-07-20
fix(enterprise): correct thread allocation guidance for the performance upgrade preview

jstirnaman:
## What changed

- **`performance-preview/configure.md`**
  - Replaced the note that told users to set `--num-io-threads` to the number of cores. The new guidance: set both `--num-io-threads` and `--datafusion-num-threads` explicitly, keeping their sum within the core count.
  - Documented both failure modes: setting only one option (startup validation error) and leaving both unset (no validation; thread allocation exceeds available cores).
  - Added workload guidance for the preview: ingest and compaction run on the IO runtime, so weight ingest nodes heavily toward IO threads and use a more balanced split (about 60/40 IO/DataFusion) on compactor nodes.
  - Added `--datafusion-num-threads` and stated core counts to all example configurations.
- **`shared/influxdb3-admin/performance-tuning.md`**
  - Added an Enterprise-only note that the thread-count formula applies to the default storage engine, with a link to the preview configuration reference.
- **`admin/clustering.md`**
  - Added a page-top banner directing performance upgrade preview users to the preview configuration reference.
  - Added a callout in the thread-allocation section explaining that the preview runs ingest and compaction on the IO thread pool.
  - Pointed the compaction troubleshooting section at the preview configuration reference.

## Why

With `--use-pacha-tree`, DataFusion threads default to two fewer than the core count. Following the previous advice (`--num-io-threads` = core count) makes the combined thread count exceed the core count, and the server fails to start with:

```
Cannot specify higher total thread count than --num-cores
```

Leaving both options unset skips this validation and allocates more threads than available cores.

## Impact

- Users who followed the previous preview guidance hit a startup failure loop on ingest, query, compactor, and all-in-one nodes. The corrected guidance and examples produce configurations that start and match the preview's IO-heavy workload profile.
- Parquet-engine (default storage) guidance is unchanged; the new notes only fence it off from preview deployments.

## Verification

- Example thread counts checked against the stated core counts on each example (sum of IO and DataFusion threads within the core count).
- `enterprise-lint` pre-commit checks passed.
- Local Hugo build (`npx hugo --quiet`): pending.

## Checklist

- [x] Rebased/mergeable
- [ ] Local build passes (`npx hugo --quiet`)

## Related

- https://github.com/influxdata/DAR/issues/726


hiltontj:
@jstirnaman FYI - I am working on:
- https://github.com/influxdata/influxdb_pro/pull/4691

This will default the IO and DataFusion runtimes (in PachaTree only) to use `io == df == num_cpus`. So, users should not need to set these things explicitly. The `--num-cores` argument is still available as a way to set the number of licensed cores a node consumes.

So, the rule for PachaTree will be different than Parquet. We might also change it in Parquet, but are holding off to prevent introducing a breaking change in license enforcement for existing customers.

- PachaTree: max(N<sub>io</sub>, N<sub>df</sub>) ≤ N<sub>licensed</sub>
- Parquet: N<sub>io</sub> + N<sub>df</sub> ≤ N<sub>licensed</sub>

jstirnaman:
> @jstirnaman FYI - I am working on:
> 
> * [fix: PachaTree runtime thread defaults and licensed-core accounting influxdb_pro#4691](https://github.com/influxdata/influxdb_pro/pull/4691)
> 
@hiltontj  Noted. @peterbarnett03  I'm publishing this to document it as-is. I'll add an issue to update it after the Enterprise change lands.

PR #7530, opened 2026-07-20
docs(influxdb3-enterprise): clarify license portability for operators

jstirnaman:
## What changed

  - Clarified that v3.10+ license files are portable between object stores.
  - Documented the default destination paths for copied commercial and trial/home license files.
  - Explained that copied license files take precedence over license-email activation.
  - Added explicit noninteractive `trial` or `home` license-type configuration to CLI, environment-variable, Docker Compose, and TOML examples.
  - Added guidance to treat license files as secrets.

  ## Why

  The previous guidance did not clearly distinguish license-file portability from email-based activation and renewal. This could lead operators to request a
  new license when a valid existing file should be copied to replacement storage.

  ## Impact

  Operators can move a valid Enterprise license to new storage with less ambiguity. Noninteractive deployments have complete examples for selecting a trial
  or At-Home license type. The page also makes clear that copying a file does not renew an expired license or change its CPU limit.

  ## Verification

  - Ran `git diff --check`.
  - Ran Vale on `content/influxdb3/enterprise/admin/license.md`: 0 errors, warnings, or suggestions.
  - Attempted focused code-block linting. It could not run because the clean worktree lacks Node dependencies (`p-limit` is missing).

### DIFF
--- a/content/influxdb3/enterprise/admin/license.md
+++ b/content/influxdb3/enterprise/admin/license.md
@@ -24,6 +24,8 @@ the license type.
 > #### License storage portability in v3.10 and later
 >
 > Starting in **v3.10**, {{% product-name %}} licenses are no longer tied to a specific cluster's storage location.
+> You can copy a valid license file to another object store when moving a
+> deployment.
 > Licenses are still required, with per-cluster CPU core limits and license expiration continuing to apply.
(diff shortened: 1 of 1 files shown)


peterbarnett03 (approved):
Looks SPLENDID to me

PR #7600, opened 2026-08-03
docs(influxdb3): add Processing Engine Python API reference

caterryan:
Adds InfluxDB3 Processing Engine Python API reference

### DIFF
--- a/content/influxdb3/enterprise/plugins/python-api-reference.md
+++ b/content/influxdb3/enterprise/plugins/python-api-reference.md
@@ -1,7 +1,7 @@
 ---
 title: Processing engine Python API reference
 description: |
-  Reference for the Python API available to InfluxDB 3 Processing Engine
+  Reference for the Python API available to InfluxDB 3 processing engine
   plugins—trigger entry points, the influxdb3_local shared API, LineBuilder,
   the in-memory cache, and plugin exceptions.
 menu:
--- a/content/influxdb3/enterprise/plugins/python-api-reference.md
+++ b/content/influxdb3/enterprise/plugins/python-api-reference.md
@@ -1,0 +1,21 @@
+---
+title: Processing engine Python API reference
+description: |
+  Reference for the Python API available to InfluxDB 3 Processing Engine
+  plugins—trigger entry points, the influxdb3_local shared API, LineBuilder,
+  the in-memory cache, and plugin exceptions.
+menu:
+  influxdb3_enterprise:
+    name: Python API reference
+    parent: Processing engine and Python plugins
+weight: 6
+influxdb3/enterprise/tags: [processing engine, plugins, API, python]
+related:
+- /influxdb3/enterprise/plugins/extend-plugin/
+- /influxdb3/enterprise/reference/cli/influxdb3/create/trigger/
+- /influxdb3/enterprise/reference/cli/influxdb3/test/
+- /influxdb3/enterprise/reference/processing-engine/
+source: /shared/influxdb3-plugins/python-api-reference.md
+---
+
+<!-- //SOURCE content/shared/influxdb3-plugins/python-api-reference.md -->
--- a/content/shared/influxdb3-plugins/python-api-reference.md
+++ b/content/shared/influxdb3-plugins/python-api-reference.md
@@ -1,0 +1,344 @@
+The Processing Engine provides a Python API that plugins use at runtime.
+Nothing in this API is importable from a plugin—the runtime injects the objects instead.
+InfluxDB passes `influxdb3_local` (an `InfluxDB3Local` instance) as the first argument to every trigger entry point and installs `LineBuilder` into Python builtins before the plugin runs, so both single-file and multi-file plugins can use them without an import.
+
+- [Trigger entry points](#trigger-entry-points)
+- [Log messages](#log-messages)
+- [Query data](#query-data)
+- [Write data](#write-data)
+- [Cache data](#cache-data)
+- [Plugin cancellation](#plugin-cancellation)
+- [LineBuilder](#linebuilder)
+- [TableBatch](#tablebatch)
+- [Exceptions](#exceptions)
+
+For usage patterns and examples, see [Extend plugins](/influxdb3/version/extend-plugin/).
+
+## Trigger entry points
+
+A plugin defines one entry point that matches the trigger type it is attached to.
+Multi-file plugins must define the entry point in the plugin directory's `__init__.py`.
+
+| Trigger type | Trigger specification | Entry point |
+| :----------- | :-------------------- | :---------- |
+| Data write (WAL) | `table:TABLE_NAME` or `all_tables` | [`process_writes`](#process_writes) |
+| Scheduled | `every:DURATION` or `cron:EXPRESSION` | [`process_scheduled_call`](#process_scheduled_call) |
+| HTTP request | `request:REQUEST_PATH` | [`process_request`](#process_request) |
+
+### `process_writes`
+
+```python
+def process_writes(
+    influxdb3_local,
+    table_batches: Sequence[TableBatch],
+    args: Mapping[str, str] | None = None,
+) -> None
+```
+
+Entry point for data write (WAL) triggers.
+InfluxDB calls the function with the rows that were just written to the database, grouped by table.
+
+- `table_batches`: the written rows as a sequence of [`TableBatch`](#tablebatch) dictionaries.
+- `args`: trigger arguments as a string-to-string mapping.
+
+### `process_scheduled_call`
+
+```python
+def process_scheduled_call(
+    influxdb3_local,
+    schedule_time: datetime,
+    args: Mapping[str, str] | None = None,
+) -> None
+```
+
+Entry point for scheduled triggers.
+
+- `schedule_time`: the trigger's fire time as a naive `datetime`.
+  The engine builds the value with `datetime.fromtimestamp()`, so it is expressed in the server's local timezone and truncated to whole seconds.
+- `args`: trigger arguments as a string-to-string mapping.
+
+### `process_request`
+
+```python
+def process_request(
+    influxdb3_local,
+    query_params: Mapping[str, str],
+    request_headers: Mapping[str, str],
+    request_body: bytes,
+    args: Mapping[str, str] | None = None,
+)
+```
+
+Entry point for HTTP request triggers.
+
+- `query_params`: URL query parameters as a string-to-string mapping.
+- `request_headers`: request headers as a string-to-string mapping.
+  Header names are lowercased, and the client's `Authorization` header is stripped before the plugin sees it.
+- `request_body`: the raw request bytes.
+- `args`: trigger arguments as a string-to-string mapping.
+
+#### Return value
+
+`process_request` returns an HTTP-style response in one of the following forms:
+
+- **A Flask-style response object**—duck-typed, not an actual `flask.Response`.
+  The object must define a `__flask_response__()` method that returns `True`, a `status_code` attribute, a `headers` mapping, and a `get_data()` method that returns `str`.
+  A genuine `flask.Response` is rejected: it does not define `__flask_response__`, and its `get_data()` returns bytes.
+- **A tuple** of `(body, status, headers)` following Flask conventions.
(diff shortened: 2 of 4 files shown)


content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+A plugin defines one entry point that matches the trigger type it is attached to.
+Multi-file plugins must define the entry point in the plugin directory's `__init__.py`.
+
+| Trigger type | Trigger specification | Entry point |
+| :----------- | :-------------------- | :---------- |
+| Data write (WAL) | `table:TABLE_NAME` or `all_tables` | [`process_writes`](#process_writes) |
+| Scheduled | `every:DURATION` or `cron:EXPRESSION` | [`process_scheduled_call`](#process_scheduled_call) |
+| HTTP request | `request:REQUEST_PATH` | [`process_request`](#process_request) |
+
+### `process_writes`
```
jstirnaman:
```suggestion
### `process_writes()`
```

FYI, backtick-fenced code doesn't render as code in headings - another reason to specify that they're functions.
Parens do get stripped out of anchors.
See the [preview](https://test2.docs.influxdata.com/pr-preview/pr-7600/influxdb3/enterprise/plugins/python-api-reference/)

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+- [LineBuilder](#linebuilder)
+- [TableBatch](#tablebatch)
+- [Exceptions](#exceptions)
+
+For usage patterns and examples, see [Extend plugins](/influxdb3/version/extend-plugin/).
+
+## Trigger entry points
+
+A plugin defines one entry point that matches the trigger type it is attached to.
+Multi-file plugins must define the entry point in the plugin directory's `__init__.py`.
```
jstirnaman, line 21:
```suggestion
A plugin defines one entry point that matches the trigger type it is attached to.
You define the function; InfluxDB calls it when the trigger fires.
Multi-file plugins must define the entry point in the plugin directory's `__init__.py`.
```

We should be explicit about who the caller is.

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
+The Processing Engine provides a Python API that plugins use at runtime.
```
jstirnaman:
```suggestion
The InfluxDB 3 processing engine provides a Python API that plugins use at runtime.
```

content/influxdb3/enterprise/plugins/python-api-reference.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Processing engine Python API reference
+description: |
+  Reference for the Python API available to InfluxDB 3 Processing Engine
```
jstirnaman:
```suggestion
  Reference for the Python API available to InfluxDB 3 processing engine
```

We try to reserve title-case for product names and (code) objects that require it (`LineBuilder`), but not product features.

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+## Trigger entry points
+
+A plugin defines one entry point that matches the trigger type it is attached to.
+Multi-file plugins must define the entry point in the plugin directory's `__init__.py`.
+
+| Trigger type | Trigger specification | Entry point |
+| :----------- | :-------------------- | :---------- |
+| Data write (WAL) | `table:TABLE_NAME` or `all_tables` | [`process_writes`](#process_writes) |
+| Scheduled | `every:DURATION` or `cron:EXPRESSION` | [`process_scheduled_call`](#process_scheduled_call) |
+| HTTP request | `request:REQUEST_PATH` | [`process_request`](#process_request) |
```
jstirnaman:
```suggestion
| Trigger type | Trigger specification | Entry point function |
| :----------- | :-------------------- | :---------- |
| Data write (WAL) | `table:TABLE_NAME` or `all_tables` | [`process_writes()`](#process_writes) |
| Scheduled | `every:DURATION` or `cron:EXPRESSION` | [`process_scheduled_call()`](#process_scheduled_call) |
| HTTP request | `request:REQUEST_PATH` | [`process_request()`](#process_request) |
```

Tells the reader/scans that they're functions without writing "function" everywhere.

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+) -> None
+```
+
+Entry point for data write (WAL) triggers.
+InfluxDB calls the function with the rows that were just written to the database, grouped by table.
+
+- `table_batches`: the written rows as a sequence of [`TableBatch`](#tablebatch) dictionaries.
+- `args`: trigger arguments as a string-to-string mapping.
+
+### `process_scheduled_call`
```
jstirnaman:
```suggestion
### `process_scheduled_call()`
```

See comment https://github.com/influxdata/docs-v2/pull/7600/changes#r3706389014

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+) -> None
+```
+
+Entry point for scheduled triggers.
+
+- `schedule_time`: the trigger's fire time as a naive `datetime`.
+  The engine builds the value with `datetime.fromtimestamp()`, so it is expressed in the server's local timezone and truncated to whole seconds.
+- `args`: trigger arguments as a string-to-string mapping.
+
+### `process_request`
```
jstirnaman:
```suggestion
### `process_request()`
```

See comment
https://github.com/influxdata/docs-v2/pull/7600/changes#r3706389014

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+The `influxdb3_local.cache` property provides an in-memory cache that persists values between plugin executions.
+
+The trigger-local cache (the default) is scoped to the database and trigger pair; the global cache is shared by every trigger in the process.
+Set `use_global=True` in cache calls to use the global cache.
+The server clears both caches on restart.
+
+Reading the `cache` property runs an expiry sweep, though the sweep only does work once per cleanup interval.
+Individual expired entries are dropped on read regardless.
+
+### `put`
```
jstirnaman:
```suggestion
### `cache.put()`
```

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+Stores a Python object in the trigger-local cache or the global shared cache.
+
+`ttl` controls expiry in seconds.
+Values in test caches (`influxdb3 test`) default to a 30-minute TTL when not provided; production caches persist values indefinitely unless a `ttl` is set.
+
+Raises `ValueError` when `ttl` is negative, `NaN`, or too large to represent as a duration.
+
+### `get`
```
jstirnaman:
```suggestion
### `cache.get()`
```

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+    default: Any | None = None,
+    use_global: bool | None = None,
+) -> Any | None
+```
+
+Fetches a value from the trigger-local or global cache.
+Expired entries are evicted on read.
+Returns the stored value, the provided default, or `None` when the key is absent.
+
+### `delete`
```
jstirnaman:
```suggestion
### `cache.delete()`
```

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+
+| Method | Adds |
+| :----- | :--- |
+| `tag(key, value)` | A tag; the value is stringified |
+| `int64_field(key, value)` | A signed integer field |
+| `uint64_field(key, value)` | An unsigned integer field; negative values raise `ValueError` |
+| `float64_field(key, value)` | A float field; integral values are rendered with a trailing `.0` |
+| `string_field(key, value)` | A string field with quotes and backslashes escaped |
+| `bool_field(key, value)` | A boolean field, rendered as `t` or `f` |
+| `time_ns(timestamp_ns)` | The nanosecond timestamp for the line |
```
jstirnaman, line 294:
```suggestion
| `time_ns(timestamp_ns)` | The nanosecond timestamp for the line |
| `build()` | Renders a record as line protocol |
```

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+
+### `build`
```
jstirnaman:
```suggestion
### `line.build()`
```

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+
+```python
+line.build() -> str
+```
+
+Renders the accumulated measurement, tags, fields, and optional timestamp as a line protocol string.
```
jstirnaman:
```suggestion
When you call `build()` on a LineBuilder instance, it renders the accumulated measurement, tags, fields, and optional timestamp as a line protocol string.
```

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+Raises `InvalidLineError` when called without any fields.
```
jstirnaman:
```suggestion
`build()` raises `InvalidLineError` when called without any fields.
```

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+
```
jstirnaman, line 306:
```suggestion
For format details, see the [line protocol reference](/influxdb3/version/reference/line-protocol/).

```

Move lp reference to the emitter.

content/shared/influxdb3-plugins/python-api-reference.md
```diff
@@ -0,0 +1,344 @@
...
+```python
+LineBuilder(measurement: str)
+```
+
+Helper for constructing InfluxDB line protocol in plugins, with nanosecond timestamps and line protocol escaping.
+For format details, see the [line protocol reference](/influxdb3/version/reference/line-protocol/).
+
+The engine installs `LineBuilder` into Python builtins before a plugin runs, so it is available to single-file and multi-file plugins without an import.
+Tag and field methods return `self`, so calls chain:
```
jstirnaman:
```suggestion
`LineBuilder` constructs InfluxDB line protocol in plugins.
It handles line protocol escaping and nanosecond timestamps, and plugins can use it without an import.

Construct a builder with `LineBuilder(measurement)`, then call the tag, field, and timestamp methods on the resulting instance.
In this section, `line` is a `LineBuilder` instance.
Tag and field methods return the same `LineBuilder` instance, so you can chain calls:
```

Be explicit about who calls it, what it returns.
Compress the no-import explanation (introduced earlier).
Explain what `line` is for in the following examples.

jstirnaman (changes requested):
If you generated this from somewhere else and would rather I edit that source, let me know.
I realize some of these suggestions might not work with the output of the generator.

caterryan:
> If you generated this from somewhere else and would rather I edit that source, let me know. I realize some of these suggestions might not work with the output of the generator.

I decided not to use a generator for now. The python api changes infrequently so we can manually keep it in sync for now.

caterryan:
Applied all suggestions in f051a377e, with two small amendments:

- "Renders a **record** as line protocol" → "Renders the **line** as line protocol" (matching line/point vocabulary elsewhere in the docs)
- Extended the sentence-case `processing engine` fix to the Core stub description and the remaining in-page occurrence

Re: the generator question — this page is a hand-maintained snapshot by design (the Sphinx source lives in a personal validation repo, so it isn't a suitable upstream yet). PR suggestions are the right venue and won't get clobbered. If the docstrings land in an official repo later, we can wire up generation like the plugin library sync.

jstirnaman (approved):
Thanks!

PR #7612, opened 2026-08-05
docs(controller): document configuration versioning

sanderson:
Add a Manage configuration versions page covering automatic version creation, version history, comparing versions, rollback, and pruning. Cross-link from the create and update pages, add configuration lifecycle events to the audit log categories, and renumber the Manage configurations menu to resolve a weight collision.

Includes TODO placeholders for images that I'll replace before the release ships.

## Checklist

- [x] Rebased/mergeable


jstirnaman (approved):
👍

Issue #7620, opened 2026-08-06
Phase A: reconcile the 24 remaining InfluxDB v1 OSS/Enterprise page pairs into shared content


jstirnaman:
## Goal

Finish the InfluxDB v1 shared-content migration by reconciling the 24 page pairs
that still exist as two copies, moving each into `content/shared/influxdb-v1/`.
When the list is empty, delete the transitional guard.

Phase C (already merged) moved the 31 pairs whose copies were *provably*
equivalent and built the mechanism: the `/product/version/` path placeholder
resolved from `data/products.yml`, `show-in`/`hide-in` matching on
`<product>/<version>`, and a blocking divergence ratchet over what remains.
Background and rejected alternatives are in `docs/adr/0001`-`0003`.

## Why these are separate from phase C

The 31 already-migrated pairs needed no editorial judgment -- `diff -w` proved
nothing was lost. These 24 differ by **1,508 lines** that require picking a
winner per hunk. There is no rule that decides it:

- Git history could not arbitrate at the time, because the clone was shallow and
  grafted at `c32e05175`. That is fixed -- `git fetch --unshallow` restores
  10,513 commits, so `git log -L` and `git blame` now work on these files. **Use
  them.** Reconciling by taste is how this content drifted in the first place.
- "Enterprise is the newer copy" is **false** for this set. It was measured and
  it does not hold: Enterprise's `exists.md` is fuller and correctly links to v1
  pages, while OSS's `mathematic-operations.md` has the better REPL link.

## Worklist

Tracked in `.ci/v1-shared-drift-manifest.json`, which CI keeps honest -- remove
an entry when its page moves to `content/shared/influxdb-v1/`. Current state
(`node .ci/scripts/check-v1-shared-drift.js --report`):

| diverging lines | page |
| --- | --- |
| 338 | `flux/flux-vs-influxql.md` |
| 183 | `concepts/glossary.md` |
| 152 | `flux/guides/conditional-logic.md` |
| 94 | `flux/guides/calculate-percentages.md` |
| 88 | `flux/get-started/transform-data.md` |
| 75 | `flux/guides/mathematic-operations.md` |
| 69 | `flux/guides/scalar-values.md` |
| 67 | `flux/guides/join.md` |
| 58 | `flux/get-started/syntax-basics.md` |
| 56 | `flux/guides/monitor-states.md` |
| 51 | `flux/guides/histograms.md` |
| 48 | `flux/guides/exists.md` |
[...]

Issue #7619, opened 2026-08-06
Hugo output is not reproducible: shortcode placeholder counter and taxonomy title casing vary between identical builds


jstirnaman:
## Summary

Two Hugo builds of **identical** code produce different HTML on about 26 of
5,959 pages. Verified with a control build: same commit, same config, built
twice to separate destinations.

This is distinct from #7408, which covers duplicate target paths from alias
collisions. These are two different causes, both order-dependent.

### 1. Shortcode placeholder counter leaks into heading IDs

`influxdb/v2/api-guide/api_intro/`:

```
- <h4 id=use-influxdb-1x-api-clients-with-hahahugoshortcode671s0hbhb>
+ <h4 id=use-influxdb-1x-api-clients-with-hahahugoshortcode598s0hbhb>
```

Hugo's internal shortcode placeholder token is reaching the rendered heading ID.
The number depends on build order, so the anchor changes between builds. Any
link to such an anchor is unstable, and the ID is meaningless to readers either
way.

### 2. Taxonomy title casing flips

```
influxdb3/core/tags/influxql        'InfluxQL ...'   -> 'Influxql ...'
influxdb3/enterprise/tags/cli       'Cli ...'        -> 'CLI ...'
influxdb/cloud/tags/javascript      'Javascript ...' -> 'JavaScript ...'
```

The casing depends on which page Hugo reads the term from first. Note it flips in
**both** directions within a single build, which rules out a one-way
normalization bug.

### 3. Asset fingerprint churn (separate, but it hides the above)

`main.<sha256>.js` and its `integrity=` attribute change between builds and
appear in every page, so a naive `diff -rq` reports ~4,800 changed files and
buries the 26 real ones.

## Impact

- Heading anchors and taxonomy page titles are not stable across deploys.
- **It makes regression testing hard.** Comparing a baseline build to a changed
  build requires normalizing fingerprints and then subtracting a control run,
  otherwise every diff looks enormous. This cost real time during the InfluxDB v1
  shared-content migration.

## Suggested next steps

1. Find why the shortcode placeholder survives into heading IDs -- likely a
   heading whose text is produced by a shortcode, rendered before replacement.
2. Give taxonomy terms explicit titles (or a normalization) so casing does not
   depend on read order.
3. Consider a CI job that builds twice and diffs, to keep new nondeterminism from
   creeping in.

## Reproducing

```sh
npx hugo --destination /tmp/pub-a
npx hugo --destination /tmp/pub-b
[...]

Issue #7617, opened 2026-08-06
Vale under-reports on pages with front matter, so product pages are linted less strictly than shared content


jstirnaman:
## Summary

Vale reports fewer alerts on a page **with** front matter than on the identical
body **without** it. Because `content/shared/**` files are body-only and product
pages carry front matter, shared content is linted strictly and product pages
are not.

## Evidence

Bisected on `content/enterprise_influxdb/v1/supported_protocols/collectd.md`
(before it was migrated to shared content), using the same body in both cases:

| body | no front matter | with front matter |
| --- | --- | --- |
| lines 1-14 | 4 errors | 4 errors |
| lines 1-20 | 4 errors | **0 errors** |
| full file | 4 errors | **0 errors** |

Findings:

- Path is irrelevant. Same result at `content/enterprise_influxdb/v1/` and at
  `content/shared/`.
- Any front matter triggers it -- `title:` alone is enough. It is not specific
  to `description: >` or `aliases:`.
- A short synthetic body is **not** affected, so it depends on body length as
  well as front-matter presence.
- Lines 15-20 of the trigger case are ordinary prose: no shortcodes, no `---`,
  no `vale off` directive.

The suppressed alerts are real: `Google.Latin` on `e.g.`, `Google.Spacing` on
double spaces after a period.

Mechanism not identified. It looks like an offset bug in Vale's front-matter
handling rather than anything in our configuration.

## Impact

- **The corpus is less linted than it appears.** A product page passing Vale is
  not evidence its prose is clean.
- **Migrating content into `content/shared/` exposes pre-existing errors**, and
  `pr-vale-check.yml` blocks on errors. Moving 31 InfluxDB v1 pages surfaced 22
  errors that had always been there. Anyone doing a shared-content migration
  should budget for this.
- Existing `content/shared/influxdb-v2` reports 16 errors across 300 files while
  equivalent product pages report 0 -- consistent with this, not with the shared
  files being worse written.

## Suggested next steps

1. Quantify the gap: run Vale over a sample of product pages with front matter
   stripped and compare alert counts.
2. Check whether this reproduces on current upstream Vale; if so, report it
   upstream with the minimal case above.
3. If it is not fixable soon, consider linting product pages with front matter
   stripped in CI so the gate is honest.

## Notes

Found while migrating InfluxDB v1 content to shared files. Two Vale configs run
locally: the default `.vale.ini`, and `.vale-instructions.ini` via the
[...]

PR #7622, opened 2026-08-06
docs(v1): share byte-identical v1 content between OSS and Enterprise

jstirnaman:
Part of #7620

## What changed

Moved the 31 InfluxDB v1 pages whose OSS and Enterprise copies are *provably*
equivalent into `content/shared/influxdb-v1/`, leaving thin front-matter stubs in
both editions. For those pages, duplication is now structurally impossible rather
than discouraged by convention.

Supporting mechanism:

- **`layouts/partials/product/content-path.html`** (new) resolves a product's URL
  root from `data/products.yml`, handling both the map and string `content_path`
  shapes.
- **`layouts/partials/article/content.html`** now resolves a product-neutral
  `/product/version/` placeholder using the page's `product`/`version` cascade
  instead of parsing the URL. The existing `<product-root>/version` form still
  works, including the digit-insensitivity that lets `/influxdb/version/` resolve
  on `influxdb3/*` pages.
- **`show-in` / `hide-in`** now match `<version>`, `<product>`, and the composite
  `<product>/<version>`.
- **`data/products.yml`** -- `enterprise_influxdb.content_path` is now a
  version-keyed map, matching `influxdb`.
- **Divergence ratchet** -- `.ci/scripts/check-v1-shared-drift.js` (+ 11 unit
  tests), `.ci/v1-shared-drift-manifest.json`, and
  `.github/workflows/pr-v1-shared-drift.yml`.

Decisions and rejected alternatives: `docs/adr/0001`-`0003`.

## Why

The two v1 trees were coupled by convention only. Nothing failed when a fix
landed in one copy and not the other, so they drifted: **74 overlapping pages
differ by 4,881 lines**, including edition-independent Flux guides and OSS links
to a `tools/shell/` page that no longer exists.

v1 is the only shared pair in this repo whose editions do not share a product
root (`/influxdb/v1/` vs `/enterprise_influxdb/v1/`), which is why the existing
placeholder could not serve it -- one literal string cannot resolve to two roots.
Moving Enterprise v1 under the `influxdb` root was considered and deferred: it
changes 186 URLs and ~1,968 link references, and none of that is needed to stop
the drift (`docs/adr/0002`).

## Impact

**Two intentional side effects reach beyond v1.** Both are called out because the
title does not imply them:

1. `show-in`/`hide-in` now trim list entries. Five call sites wrote
   `"core, enterprise"` with a space, and the untrimmed token never matched, so
   `powerbi.md` rendered *"Database token: Your with query permissions for the
   target database"* on Enterprise, Clustered, and Cloud Dedicated. Those pages
   now render as written. Affects 2 pages.
2. Moving text out of front-matter'd pages exposes pre-existing Vale errors,
   because Vale under-reports on pages that have front matter (filed as #7617).
   21 mechanical fixes applied -- spacing, Latin abbreviations, two duplicated
   words. `50th percentile` is exempted inline as a standard technical term.

The 24 pairs that need per-hunk editorial judgment (1,508 lines) are **not** in
this PR. They are tracked in #7620 and guarded by the ratchet, which blocks any PR
that increases divergence in a listed pair. The ratchet is transitional and is
deleted when the manifest empties.

## Verification

- `npx hugo` exits 0; 5,959 pages.
- **Zero** unresolved `/product/version/` or `<root>/version/` placeholders
  sitewide.
- **Baseline vs migrated, whole-site diff:** 60 v1 pages and 2 v3 pages changed as
  intended, **0 unintended**. Asset fingerprints were normalized, and Hugo build
  nondeterminism (~26 pages, filed as #7619) was proven with a control build of
  identical code and subtracted.
- Ratchet proven to fail when a pair's divergence increases and pass when
  restored; `node --test` 11/11.
- Vale clean under both the default config and `.vale-instructions.ini`
  (0 errors, 31 files). Codeblock lint, `check-source-paths`, and
  `check-feedback-links` pass. All pre-commit hooks pass.
- Cypress `test:shortcode-examples`: 5/6. The one failure,
  `Docs home, can search with mispellings`, queries Algolia's hosted index, which
  reflects production and cannot be affected by this branch.

## Checklist

- [x] Rebased/mergeable


jstirnaman:
## Fixes for Copilot review comments

**What changed**

- `identify-version.md` and `mcp-server.md` stubs (OSS v1 and Enterprise v1) now source from existing shared content (`shared/identify-version.md`, `shared/influxdb3-admin/mcp-server-docs-only.md`) instead of empty, unwired wrapper files. The wrapper files are deleted.
- Removed a dead `enterprise-v1` identifier from a `hide-in` list in `identify-version.md`; it never matched any identifier the shortcode produces, and `v1` in the same list already covers both editions.
- Removed the duplicate `h1` from `udp.md` and `graphite.md`; the page title already renders one.
- Reverted `enterprise_influxdb.content_path` in `products.yml` from a map back to a string, and regenerated `PLATFORM_REFERENCE.md`. The map shape broke `getProductLabelMap()` (silently dropped the `product:v1-enterprise` label) and made the generator omit the product's content path and URL. The string shape already resolves correctly at render time via the existing version-append logic in `product/content-path.html`.
- Added Cypress coverage for `show-in`/`hide-in` composite (`<product>/<version>`), bare-product, and whitespace-trimmed identifiers, and for `article/content.html`'s `/product/version/` placeholder resolution against both a map-shaped and a string-shaped `content_path`.
- Closed a gap in `check-v1-shared-drift.js`: the ratchet now unions the manifest's pairs at the base ref with its pairs at head, so removing an entry from the manifest can no longer, by itself, take a pair out of comparison.

**Why**

Copilot's review flagged 10 issues. All 10 checked out: three meant pages would render with no article body, one was already visibly broken in this PR's own `PLATFORM_REFERENCE.md` diff, two were rendering bugs (duplicate h1), one was a CI control gap, and three were missing test coverage for behavior this PR introduced.

**Impact**

- `/influxdb/v1/administration/identify-version/`, `/enterprise_influxdb/v1/administration/identify-version/`, `/influxdb/v1/tools/mcp-server/`, and `/enterprise_influxdb/v1/tools/mcp-server/` render full content instead of blank pages.
- `udp.md`/`graphite.md` pages render one `h1` instead of two.
- Enterprise v1 content is labeled and referenced correctly again in `PLATFORM_REFERENCE.md` and the label-mapping script.
- A future PR can no longer bypass the v1 drift check by editing the manifest alone.

**Verification**

- `npx hugo --quiet`: clean build.
- Confirmed rendered article body text (not just frontmatter) on all four previously-blank pages.
- Confirmed single `<h1>` in rendered output for the udp/graphite pages.
- Confirmed `/product/version/` links resolve to the correct edition root on both OSS and Enterprise v1 pages.
- `node cypress/support/run-e2e-specs.js --spec "cypress/e2e/content/shortcodes.cy.js,cypress/e2e/content/shortcodes-real-pages.cy.js" --no-mapping`: 179 tests, 167 passing, 12 pending (skipped on non-v1 products), 0 failing.
- `node --test .ci/scripts/check-v1-shared-drift.test.js`: 12/12 passing.
- `node .ci/scripts/check-v1-shared-drift.js --base HEAD`: "No v1 pair diverged further (30 pairs checked)."

Commits: f51d6cbb6, 54e378105, 0885a5f04, 8298a5f35, 8aba0d096.

Issue #7628, opened 2026-08-07
Confirm Explorer and Admin UI support levels for Cloud Dedicated and InfluxDB 3 Cloud


jstirnaman:
## Background

A support ticket told a Cloud Dedicated customer they could "technically use" InfluxDB 3 Explorer with Cloud Dedicated. Product's correction: Explorer has only partial support for Dedicated, and Dedicated customers should use the Admin UI at [console.influxdata.com](https://console.influxdata.com).

The docs fix for that removes the incorrect guidance and adds callouts pointing Dedicated and InfluxDB 3 Cloud users to the console. It deliberately states **only the directive** — use the Admin UI — and makes no claim about what Explorer or the console can actually do, because those specifics are unverified. Three questions need answers before the docs can state capabilities.

### 1. Console feature support (Cloud Dedicated + InfluxDB 3 Cloud)

`content/influxdb3/cloud-dedicated/get-started/_index.md` has a "Tools to use" table with Administration / Write / Query columns. The Admin UI has **no row**, because its support level is unconfirmed.

The console does have a schema browser and Query History, but no query editor is documented — so it is unclear whether "Query" should be checked. Given that Cloud Dedicated and InfluxDB 3 Cloud now share a common UI, does one row cover both?

### 2. Explorer and InfluxDB Clustered

`content/influxdb3/explorer/about/_index.md` previously said Explorer would support Serverless, Dedicated, and Clustered "in a future release". That statement was stale and is gone. Clustered is now unmentioned rather than asserted either way, since it does not appear in Explorer's supported-products list.

Is Clustered supported, planned, or out of scope?

### 3. Explorer and InfluxDB 3 Cloud

The Explorer support table lists Core, Enterprise, Cloud Serverless, and Cloud Dedicated. **InfluxDB 3 Cloud is absent.** Can Explorer connect to an i3c instance at all, and if so at what level? A row was not added because the answer is unknown.

## Current documented state

| Product | Explorer support as documented |
| --- | --- |
| InfluxDB 3 Core | Query and administer |
| InfluxDB 3 Enterprise | Query and administer |
| InfluxDB Cloud Serverless | Query only (partial) |
| InfluxDB Cloud Dedicated | Query only (partial) |
| InfluxDB 3 Cloud | not documented |
| InfluxDB Clustered | not documented |

The Serverless and Dedicated rows carry forward the pre-existing "(Query Mode Only)" wording; they are not newly verified either.

[...]

Issue #7627, opened 2026-08-07
InfluxDB 3 Cloud has no identify-version page, blocked by `version: cloud` key collision


jstirnaman:
## Problem

Every other InfluxDB product has an "Identify version" page sourced from `content/shared/identify-version.md`:

- `content/influxdb3/core/admin/identify-version.md`
- `content/influxdb3/enterprise/admin/identify-version.md`
- `content/influxdb3/cloud-dedicated/admin/identify-version.md`
- `content/influxdb3/cloud-serverless/admin/identify-version.md`
- `content/influxdb3/clustered/admin/identify-version.md`
- plus v1, v2, Cloud (TSM), Enterprise v1, and `content/platform/`

**InfluxDB 3 Cloud (`content/influxdb3/cloud/`) has none.** Those pages also cross-link each other through `alt_links`, so the gap shows up there too.

## Why it isn't a simple copy

Adding the page the usual way is blocked by a version-key collision. Two different products cascade the same `version` value:

| Product key | Product | Content path | Cascaded `version` |
| --- | --- | --- | --- |
| `influxdb_cloud` | InfluxDB Cloud (TSM) | `content/influxdb/cloud/` | `cloud` |
| `influxdb3_cloud` | InfluxDB 3 Cloud | `content/influxdb3/cloud/` | `cloud` |

`layouts/shortcodes/show-in.html` matches on `.Page.Params.version`, so `{{% show-in "cloud" %}}` cannot tell the two apart. In `content/shared/identify-version.md` that block already renders **"InfluxDB Cloud (TSM) detection"**, and TSM Cloud is one of the file's consumers. Sourcing that file from an InfluxDB 3 Cloud page would show i3c readers TSM detection instructions.

## Options

1. Disambiguate with `product`-based conditionals (`influxdb3_cloud` vs `influxdb_cloud`) instead of version-based, either as a new shortcode or a parameter on `show-in`/`hide-in`.
2. Give InfluxDB 3 Cloud a distinct `version` value in `data/products.yml`, and audit every existing `show-in`/`hide-in "cloud"` usage for fallout.
3. Write a standalone InfluxDB 3 Cloud page that does not source the shared file — least invasive, but duplicates content.

Option 1 or 2 is an architectural call, which is why this is filed rather than fixed inline.

## Note for whoever picks this up

[...]

Issue #7626, opened 2026-08-07
Ambiguous URL detection: *.influxdb.io matches both Cloud Dedicated and InfluxDB 3 Cloud


jstirnaman:
## Problem

`content/shared/identify-version.md` tells readers that a `*.influxdb.io` hostname means InfluxDB Cloud Dedicated:

| URL Pattern | Product |
| --- | --- |
| `*.influxdb.io` | InfluxDB Cloud Dedicated |

That is no longer sufficient. **InfluxDB 3 Cloud hostnames also end in `.influxdb.io`**, so the pattern matches both products and a reader can land on the wrong docs.

The two products are distinguishable by the subdomain immediately before `influxdb.io`:

| Product | Console term | Documented host |
| --- | --- | --- |
| InfluxDB Cloud Dedicated | clusters (cd) | `cluster-id.a.influxdb.io` |
| InfluxDB 3 Cloud | instances (i3c) | `instance-id.enterprise.influxdb.io` |

## Evidence

Both forms are already canonical in this repo, as `placeholder_host` in `data/products.yml`:

- `influxdb3_cloud_dedicated` → `cluster-id.a.influxdb.io`
- `influxdb3_cloud` → `instance-id.enterprise.influxdb.io`

`cluster-id.a.influxdb.io` also appears in content, for example `content/influxdb3/cloud-dedicated/query-data/execute-queries/client-libraries/go.md`.

## Suggested fix

In `content/shared/identify-version.md`:

1. Split the `*.influxdb.io` row into `*.a.influxdb.io` (Cloud Dedicated) and `*.enterprise.influxdb.io` (InfluxDB 3 Cloud).
2. Tighten the `{{% show-in "cloud-dedicated" %}}` block, which currently states the URL pattern as `*.influxdb.io` with the example `cluster-id.a.influxdb.io`.
3. Consider adding InfluxDB 3 Cloud to the "Understanding InfluxDB products" comparison table, which has no row for it.

## Why this needs its own PR

`content/shared/identify-version.md` is sourced by **10 pages** across v1, v2, Cloud (TSM), Cloud Serverless, Cloud Dedicated, Clustered, Core, Enterprise, and the platform overview. Any edit needs a full build plus a per-product render check to confirm nothing leaks into the wrong product.

## Context

Split out of the Explorer / console UI docs fix, which was deliberately kept narrow. Product terminology confirmed by support: clusters are cd, instances are i3c.

PR #7629, opened 2026-08-07
fix(explorer): clarify cloud support and point to the Admin UI

jstirnaman:
Related: #7626, #7627, #7628 (follow-ups, not closed by this PR)

## What changed

Explorer docs no longer imply that InfluxDB 3 Explorer is a supported way to work with InfluxDB Cloud Dedicated, and the Cloud Dedicated and InfluxDB 3 Cloud docs now point administrators to the Admin UI at console.influxdata.com.

- **About Explorer** — replaced the "in a future release" statement with a Product support section describing current support.
- **Explorer get started** — turned the supported-products list into a table separating "query and administer" from "query only", and added a callout sending Cloud Dedicated and InfluxDB 3 Cloud users to the Admin UI.
- **Cloud Dedicated admin and get started** — added callouts preferring the Admin UI over the standalone Explorer UI.
- **Shared InfluxDB 3 admin index** — added the same callout for InfluxDB 3 Cloud, scoped with `show-in` so it renders only for that product.
- **Explorer manage databases / manage tokens** — fixed two links pointing at `#run-in-query-or-admin-mode`, an anchor that does not exist. The heading on the install page is "Choose operational mode".
- **Cloud Dedicated get started** — fixed a pre-existing broken anchor in the tools table. It linked to `#influxdb-3-client-libraries`, but the heading is "InfluxDB client libraries" and covers both v3 and v2 client libraries, so the link was wrong rather than the heading. Unchanged on master; the link checker only began scanning the file because this branch modifies it.

## Why

A customer was told they could "technically use" Explorer with Cloud Dedicated. Explorer has only partial support for Dedicated, and Dedicated customers should be using the Admin UI.

The docs invited that mistake by contradicting themselves. About Explorer said support for Serverless, Dedicated, and Clustered was coming "in a future release", while the get started guide already listed Serverless and Dedicated as supported in query mode. Neither page mentioned the Admin UI.

## Impact

Cloud Dedicated and InfluxDB 3 Cloud readers are now directed to the console for administrative tasks. Core and Enterprise guidance is unchanged — Explorer is still the recommended UI for those products, and the new shared-content callout is scoped so it never appears on their pages.

**The callouts state a preference and nothing more.** They don't enumerate which resources the Admin UI manages, and they make no claim about what Explorer can do against a cluster. Two reasons:

1. Console feature support is unverified — see #7628.
2. Admin functionality is documented in the admin sections, and duplicating it in a callout means maintaining it twice. Enumerating resources also misleads by omission: an admin section whose children are CLI and API pages reads as though the Admin UI doesn't support those tasks.

Which interface to use for each task belongs on the specific admin pages, where it can be stated with the detail it needs.

For the same reason, no Admin UI row was added to the Cloud Dedicated tools comparison table.

## Verification

- `npx hugo --quiet` passes.
- **Shared-content scoping.** `content/shared/influxdb3-admin/_index.md` is sourced by Core, Enterprise, and InfluxDB 3 Cloud. An unscoped callout would tell Core and Enterprise users to avoid Explorer, which is the opposite of the guidance for those products. Confirmed in the built output that the callout appears on `/influxdb3/cloud/admin/` and on neither `/influxdb3/core/admin/` nor `/influxdb3/enterprise/admin/`. A site-wide search finds it on exactly two pages: Cloud and Cloud Dedicated.
- `version: cloud` is not globally unique — `influxdb_cloud` (Cloud TSM) cascades the same value. It is unambiguous here only because Cloud TSM doesn't source this file, which is why the scoping is verified by build output rather than assumed. Tracked in #7627.
- All internal and external links in the change resolve in the build, including the corrected `#choose-operational-mode` and `#influxdb-client-libraries` anchors.
- Vale reports 0 errors and 0 warnings across all changed files, under both the default and `cloud-dedicated` configs.
- No layout changes, so no e2e tests required.

## Checklist

- [x] Rebased/mergeable

***

<details>
<summary><b>Suggested reviewers</b> (click to expand)</summary>

Based on files changed:

| Content | Product |
| --- | --- |
| Explorer | mavarius, peterbarnett03 |
| Cloud Dedicated | ritwika314, sanderson |
| `/content/shared/` | influxdata/product-managers |

Product review is worth having here: the support levels this PR documents came from a support thread, and the open questions in #7628 are what kept further capability claims out of the diff.

</details>

### DIFF
--- a/content/influxdb3/cloud-dedicated/admin/_index.md
+++ b/content/influxdb3/cloud-dedicated/admin/_index.md
@@ -9,6 +9,18 @@ menu:
 weight: 6
 ---
 
+> [!Important]
+> #### Use the Admin UI, not InfluxDB 3 Explorer
+>
+> Administer your cluster with the {{% product-name %}}
+> [Admin UI](https://console.influxdata.com) at `console.influxdata.com` or with
+> the [`influxctl` CLI](/influxdb3/cloud-dedicated/reference/cli/influxctl/).
+>
+> [InfluxDB 3 Explorer](/influxdb3/explorer/) is the standalone UI for
+> [InfluxDB 3 Core](/influxdb3/core/) and
+> [InfluxDB 3 Enterprise](/influxdb3/enterprise/).
+> Don't use Explorer to administer {{% product-name %}}.
+
 The following articles provide information about managing your InfluxDB Cloud
 Dedicated resources:
 
--- a/content/influxdb3/cloud-dedicated/admin/_index.md
+++ b/content/influxdb3/cloud-dedicated/admin/_index.md
@@ -12,14 +12,11 @@ weight: 6
 > [!Important]
 > #### Use the Admin UI, not InfluxDB 3 Explorer
 >
-> Administer your cluster with the {{% product-name %}}
-> [Admin UI](https://console.influxdata.com) at `console.influxdata.com` or with
-> the [`influxctl` CLI](/influxdb3/cloud-dedicated/reference/cli/influxctl/).
->
-> [InfluxDB 3 Explorer](/influxdb3/explorer/) is the standalone UI for
+> Use the {{% product-name %}} [Admin UI](https://console.influxdata.com) at
+> `console.influxdata.com` rather than
+> [InfluxDB 3 Explorer](/influxdb3/explorer/), the standalone UI for
 > [InfluxDB 3 Core](/influxdb3/core/) and
 > [InfluxDB 3 Enterprise](/influxdb3/enterprise/).
-> Don't use Explorer to administer {{% product-name %}}.
 
 The following articles provide information about managing your InfluxDB Cloud
 Dedicated resources:
--- a/content/influxdb3/cloud-dedicated/get-started/_index.md
+++ b/content/influxdb3/cloud-dedicated/get-started/_index.md
@@ -128,6 +128,18 @@ The following table compares tools that you can use to interact with
 > Avoid using the `influx` CLI with {{% product-name %}}. While it
 > may coincidentally work, it isn't supported.
 
+> [!Important]
+> #### Administer your cluster with the Admin UI
+>
+> To manage databases, tables, tokens, and users, use the {{% product-name %}}
+> [Admin UI](https://console.influxdata.com) at `console.influxdata.com` or the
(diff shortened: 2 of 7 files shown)


jstirnaman:
Pushed c5c6174, which addresses the link-check failure and all three review comments.

## Review comments — all three accepted

The review was right, and one comment caught a real defect. I had written "use the Admin UI to manage databases, tables, tokens, and users" for **both** Cloud Dedicated and InfluxDB 3 Cloud. That enumeration is grounded for Dedicated — `content/influxdb3/cloud-dedicated/admin/account/_index.md` documents an Admin UI sidebar with exactly Databases, Tables, Database Tokens, Management Tokens, and Users. I then carried the same wording to InfluxDB 3 Cloud without checking, and for i3c it is wrong:

- `content/influxdb3/cloud/admin/tokens/_index.md` says tokens are created and managed with the `influxdb3` CLI or the HTTP API.
- `content/influxdb3/cloud/admin/users/admin-ui.md` scopes the i3c Admin UI to users only.

So the callout contradicted the product's own docs. Fixed by making the callouts recommendation-only:

- `content/shared/influxdb3-admin/_index.md` — now points to the Admin UI **and the tools described in the following articles**, so the linked pages remain the authority on which tool does what.
- `content/influxdb3/explorer/about/_index.md` — "administer your cluster or instance" instead of the resource list.
- `content/influxdb3/explorer/get-started.md` — dropped "you can query data, but you can't administer your cluster". Support characterization stays in the support table, as suggested.

The Cloud Dedicated get-started callout keeps its enumeration, since that one is documented.

## Link check

`#influxdb-3-client-libraries` in the Cloud Dedicated get-started tools table. The heading is `### InfluxDB client libraries` (`#influxdb-client-libraries`) and it covers both v3 and v2 client libraries, so the heading is right and the link was wrong — fixed the link.

This anchor is **unchanged on master**; my diff never touched that line. The link checker only started scanning the file because this branch modifies it, which surfaced a latent break. The identical mismatch exists at `content/influxdb3/clustered/get-started/_index.md:111`. I left it alone to avoid pulling another file into link-check scope and growing this PR — worth a separate fix.

## Vale

0 errors. The 4 warnings are pre-existing and untouched by this branch: the three `admin` → `administrator` hits are on line 10 of product stub files I don't modify, and `manage-databases.md:47` predates this work.

---

sanderson (approved):

Issue #5155, opened 2023-09-27
Telegraf guide: how to use the Converter processor to coerce data before writing


jstirnaman:
-  how to use the Converter processor to coerce data before writing
- how to use Exec and Grok to filter data before writing.

<@U01TCPBF3U4> could you please add this to the docs backlog, if you don't already have something?
&gt;  That sounds like an opportunity for a KB article, etc to get in front of any issue and turn it into a productive ehancement.

[Slack Message](https://influxdata.slack.com/archives/C04768675QD/p1695828045290899?thread_ts=1695822639.441929&cid=C04768675QD)

PR #7640, opened 2026-08-11
docs(influxdb3): define cache --node-spec semantics and warm-up trade-off

reidkaufmann:
## Summary

The `--node-spec` flag on `influxdb3 create last_cache` / `create distinct_cache` (Enterprise) has never had defined semantics in the docs — the admin pages say only "Specify which nodes the [cache] should be configured on," the CLI reference pages omit the flag entirely, and the shared "Cache data loading" sections promise unconditional historical loading with no node-spec caveat.

This PR documents what the flag actually controls: **which Enterprise nodes load historical data into the cache** (at creation and on restart). All query-capable nodes serve the cache and add newly written data to it. The guidance is explicit about when to use it: only to reduce the initial cache warm-up load, and only if temporarily incomplete query results from excluded nodes are acceptable.

## Changes (all in `content/shared/`, Enterprise-gated with `{{% show-in "enterprise" %}}`)

- `influxdb3-cli/create/last_cache.md`, `influxdb3-cli/create/distinct_cache.md`: add an "Additional Enterprise option" table for `-n`/`--node-spec` (pattern borrowed from `create/trigger.md`), including the warm-up/incomplete-results trade-off. The flag was previously undocumented on these pages.
- `influxdb3-admin/last-value-cache/create.md`, `influxdb3-admin/distinct-value-cache/create.md`: replace the undefined "configured on" bullet with the historical-data-load semantics and the "only specify a node list if…" guidance.
- `influxdb3-admin/last-value-cache/_index.md`, `influxdb3-admin/distinct-value-cache/_index.md`: add an Enterprise paragraph to "Cache data loading" covering multi-node behavior: excluded nodes don't load historical data (at creation or restart), still serve and update the cache, and return only post-creation/post-restart values until new data arrives.

## Dependency / timing

The "excluded nodes still serve the cache and add newly written data to it" behavior is completed by influxdata/influxdb_pro#5155 (in review). Released versions have a defect where a cache created with an explicit node list errors on excluded query nodes until they restart (EAR-driven fix; see influxdata/EAR#7045). Suggest holding this PR as draft until #5155 merges and/or ships. Historical-load semantics (`--node-spec` gating warming) are accurate for released versions.

## Verification

- `npx hugo --quiet`: clean build.
- Rendered output checked: the new content appears on `influxdb3/enterprise/...` pages (admin LVC/DVC + CLI reference) and does not appear on the corresponding `influxdb3/core/...` pages; the internal links to the create pages render.
- `yarn lint` / lefthook pre-commit: pass (Vale skipped locally — no binary/Docker; runs in CI).

### DIFF
--- a/content/shared/influxdb3-admin/distinct-value-cache/_index.md
+++ b/content/shared/influxdb3-admin/distinct-value-cache/_index.md
@@ -79,6 +79,17 @@ node requires to maintain it. Consider the following:
 On cache creation, {{% product-name %}} loads historical data into the cache.
 On restart, the server automatically reloads cache data.
 
+{{% show-in "enterprise" %}}
+In a multi-node cluster, only nodes included in the cache's
+[node specification](/influxdb3/version/admin/distinct-value-cache/create/)
+load historical data--at creation and on restart.
+Excluded nodes still serve the cache and add newly written data to it,
+but until new data arrives, queries served by those nodes return only values
+written after the cache was created (or after the node restarted).
+Only specify a node list to reduce the initial cache warm-up load, and only
+if you accept temporarily incomplete query results from excluded nodes.
+{{% /show-in %}}
+
 ### High cardinality limits
 
 “Cardinality” refers to the number of unique key column combinations in your 
--- a/content/shared/influxdb3-admin/distinct-value-cache/create.md
+++ b/content/shared/influxdb3-admin/distinct-value-cache/create.md
@@ -12,8 +12,14 @@ to create a Distinct Value Cache (DVC). Provide the following:
 - **Table** (`-t`, `--table`): _({{< req >}})_ The name of the table to
   associate the DVC with.
 {{% show-in "enterprise" %}}
-- **Node specification** (`-n`, `--node-spec`): Specify which nodes the DVC
-  should be configured on.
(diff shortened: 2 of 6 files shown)


reidkaufmann:
Status update: the behavior this documents is now merged — influxdata/influxdb_pro#5155 (squash `cfbf96d184`) landed on `main` on 2026-08-13, with a 3.11 backport in flight (influxdata/influxdb_pro#5205). **Not yet in a tagged release**: released versions still error on node-spec-excluded query nodes until restart, as described in the PR body. If the timing convention is to hold docs for the release tag, this can merge whenever 3.11.x/3.12 ships with the fix; undrafting now so review can proceed.

jstirnaman (approved):

reidkaufmann:
@jstirnaman Do you want me to wait to merge until 3.11.2 releases? Some content is correct now, but a few statements only become true upon release of 3.11.2 and/or 3.12...

jstirnaman:
> @jstirnaman Do you want me to wait to merge until 3.11.2 releases? Some content is correct now, but a few statements only become true upon release of 3.11.2 and/or 3.12...

@reidkaufmann We can bisect it or comment those lines out (if clean enough) until 3.11.2/3.12.

PR #7654, opened 2026-08-14
docs(style): formalize recommendation voice and page-level weights

sanderson:
- Recommendations use first-person plural ("We recommend..."), not third-party attributions, matching existing practice across the docs (Google.We and Google.FirstPerson are already disabled in Vale)
- Set weight at the page level rather than on the menu entry: menu items inherit it, and page-level weight keeps children-shortcode and other non-menu listings sorted (Hugo sorts unweighted pages after weighted ones, so mixed placement within a section breaks ordering)

Documented in the canonical agent instructions (adapters regenerated) and DOCS-FRONTMATTER.md.

## Checklist

- [x] Rebased/mergeable


jstirnaman (approved):
Thanks for adding this! I'll DRY this up separately.

PR #7655, opened 2026-08-14
docs(telegraf): Telegraf documentation overhaul

sanderson:
Overhauls the Telegraf v1 documentation for structure, depth, and accuracy.

## What changed

- **Two reading paths**: the main nav tells the learning story (Install → Get started → How Telegraf works → Configure → Use plugins → Examples → Administer) while the reference nav stays terse (Plugin directory, Commands, Data formats, Glossary, and new Agent status evaluation and Supported platforms sections).

- **New sections**:
  - How Telegraf works (metric model, data pipeline)
  - Use plugins task guides (including parsing/serializing data and processor/aggregator behavior)
  - 12 walkthrough-style configuration examples (system metrics through Docker, Kafka, MQTT, OPC UA, and SNMP)
  - Administer Telegraf (service, monitoring, agent statuses, troubleshooting).

- **Configuration reference decomposed**: the old single-page reference is now nine topic pages with a slim overview index. All inbound anchors keep resolving, and new anchor targets fix ~400 fragment links from the generated plugin pages that were silently broken.

- **Data formats refreshed**: four missing format pages added (`prometheus`, `openmetrics`, `parquet`, `xpath_cbor`), the JSON parser and serializer pages rewritten in depth, and every format page checked against upstream.

- **Accuracy**: content verified against the Telegraf source and the 1.39 binary rather than READMEs alone; example output was produced by running real configurations where practical.

- **Shared content**: agent status pages are now shared with Telegraf Controller (`content/shared/telegraf/`); the Controller pages remain canonical.

## What didn't change

The auto-generated plugin reference directories are untouched, every moved page carries a Hugo alias, and high-traffic URLs are unchanged.


content/shared/telegraf/agent-status-eval/examples.md
```diff
@@ -0,0 +1,242 @@
...
+[[outputs.heartbeat]]
+  url = "http://telegraf_controller.example.com/agents/heartbeat"
+  instance_id = "agent-123"
+  interval = "1m"
+  include = ["hostname", "statistics", "configs", "logs", "status"]
+
+  [outputs.heartbeat.status]
+    ok = "metrics > 0"
+    warn = "outputs.influxdb_v2.exists(o, o.buffer_fullness > 0.8)"
+    fail = "outputs.influxdb_v2.exists(o, o.buffer_fullness > 0.95)"
```
jstirnaman, line 98: Guard both output lookups with `has(outputs.influxdb_v2)`, as the later examples do. This complete configuration defines only `outputs.heartbeat`, so the key might not exist. When it doesn't, CEL returns an evaluation error, and Telegraf skips the heartbeat instead of returning `warn` or `fail`.

content/shared/telegraf/agent-status.md
```diff
@@ -0,0 +1,111 @@
...
+
+For full details on evaluation flow, configuration options, and available
+variables and functions, see the
+[Agent status evaluation reference](/telegraf/controller/reference/agent-status-eval/).
+
+## Configure agent statuses
+
+To configure status evaluation, add `"status"` to the `include` list in your
+heartbeat plugin configuration and define CEL expressions in the
+`[outputs.heartbeat.status]` section.
```
jstirnaman, line 40: Document that expressions using `log_errors` or `log_warnings` require `"logs"` or `"statistics"` in `include`. Telegraf registers the logging callback only when one of those values is present. With only `"status"`, both counters remain zero and can incorrectly report a healthy agent.

content/shared/telegraf/agent-status-eval/examples.md
```diff
@@ -0,0 +1,242 @@
...
+
+For the full list of available variables and functions, see:
+
+- [CEL variables](../variables/)
+- [CEL functions and operators](../functions/)
+
+## Basic health check
+
+**Scenario:** Report `ok` when Telegraf is actively processing metrics.
+Fall back to the default status (`ok`) when no expression matches.
```
jstirnaman: Change the fallback status to `fail`. The configuration sets `default = "fail"`, and the explanation below also describes `fail`.

jstirnaman (commented):
Three documentation accuracy comments.

layouts/partials/sidebar.html
```diff
@@ -106,7 +106,7 @@ <h4 class="flux">Flux</h4>
   {{ end }}
 
   <!-- Platform menu for 1.x docs -->
-  {{ $platformWhitelist := (slice "telegraf_v1" "chronograf_v1" "kapacitor_v1" "enterprise_influxdb_v1" "influxdb_v1") }}
+  {{ $platformWhitelist := (slice "chronograf_v1" "kapacitor_v1" "enterprise_influxdb_v1" "influxdb_v1") }}
```
jstirnaman, line 109: I'm wondering if, under `/platform`, we should add `v1`, `v2`, and `v3` pages or structures to recommend which tools to use?

jstirnaman:
> Three documentation accuracy comments.

These came from 2 different Codex passes, but I didn't verify against live Telegraf.

content/shared/telegraf/agent-status-eval/examples.md
sanderson:
```suggestion
Fall back to the default status (`fail`) when no expression matches.
```

layouts/partials/sidebar.html
sanderson: Possibly. I'm wondering if these pages even really get used any more. I removed them from the Telegraf nav because with the overhaul, the nav is much more robust and the added nav items seemed like overkill. There were also a few duplicate links.

sanderson:
@jstirnaman I added fixes for most of what Codex called out. The one I didn't was about the `has()` guard. The flow of this doc is additive, so each section adds on to the last. The block it commented on laid the groundwork for evaluating plugin-specific metrics and then went on to layer on the `has()` guard principle.

jstirnaman (approved):

PR #7658, opened 2026-08-17
docs(controller): document global constants and rename dynamic values to value substitution

sanderson:
Documents global constants and renames the dynamic values concept to value substitution

## Checklist

- [x] Rebased/mergeable


jstirnaman (approved):
👍

PR #7660, opened 2026-08-18
docs(influxdb3): add node lifecycle guide and orchestrated upgrade paths

jstirnaman:
Add a Node lifecycle page for InfluxDB 3 Core and Enterprise that documents
how a node moves through the catalog states it registers itself in.

Shared content (content/shared/influxdb3-admin/node-lifecycle.md):

- Node states (running, stopping, stopped, removing) and node identity
  (node ID compared to instance ID)
- Lifecycle state diagrams, scoped per product
- Register, stop, remove, and re-register a node, including the two paths to
  stopped and why only a graceful stop node records a final snapshot
- Conditions under which removal is refused (409) and how to resolve each
- Restart compared to removal, the most common way to lose data during
  routine maintenance
- Deploy with an orchestrator: Kubernetes/Helm and Ansible/systemd
- Verify node state and troubleshoot lifecycle issues

Orchestrated upgrades (content/shared/influxdb3-admin/upgrade.md):

- Add Helm and Ansible tabs to the multi-node rolling upgrade procedure.
  The Helm chart uses a single chart-wide image.tag and a StatefulSet per
  node mode, so a plain helm upgrade rolls every mode at once and does not
  follow the recommended node upgrade order. Document sequencing rollouts
  with updateStrategy.rollingUpdate.partition.
- Add troubleshooting entries for out-of-order Helm rollouts, nodes that
  do not return to running after a rollout, and catalog entries that
  multiply when a deployment assigns unstable node IDs.

Cross-link the lifecycle page from clustering, recover-node, the Kubernetes
install page, and the remove/show/stop node CLI references, and link the
lifecycle and upgrade troubleshooting sections to each other.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01WUNvDmPaYvcFX5rchCiYhF

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -225,3 +225,18 @@ to publish the stronger caution plus the log diagnostic, or hold until fixed.
 **Catalog sync interval default of 10 seconds** (page: "Verify node state").
 Carried over from the published `stop node` CLI page. Confirm against current
 config options and name the setting on the page if it is user-tunable.
+
+**Per-engine name for the drain step a graceful stop forces** (pages: shared
+`stop node` CLI "Behavior", `admin/recover-node.md` step 2). Both spots read
+`(Parquet: WAL flush; upgraded engine: WAL snapshot)` on master. Changed to
+`Parquet persistence` on this branch, on the reading that:
+
+- WAL flush is the constant `--wal-flush-interval` operation (default 1s), so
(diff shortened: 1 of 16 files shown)


jstirnaman:
Dupe

Issue #7662, opened 2026-08-18
Phase B: reconcile the 30 remaining InfluxDB v1 OSS/Enterprise duplicate pages


jstirnaman:
## Goal

Reconcile the 30 InfluxDB v1 pages that still exist as two copies -- one under
`content/influxdb/v1/` and one under `content/enterprise_influxdb/v1/` -- moving
each into `content/shared/influxdb-v1/`. When the list empties, enumerate what
remains before touching the ratchet.

## Why this exists as a separate phase

Phase C (#7622) moved the 31 pairs whose copies were provably equivalent under
`diff -w`. Phase A (#7620) reconciled the 24 pairs the manifest deferred. Neither
covered these 30, and they were never manifest-tracked: the manifest was created
in the phase C commit already holding only the 24 deferred pairs, so the union of
every entry it has ever held is exactly 24.

The counts close against the phase C baseline:

| | count |
| --- | --- |
| Overlapping paths | 95 |
| Reconciled (31 phase C + 24 phase A) | 55 |
| Still duplicated (this issue) | 30 |
| Section `_index.md`, expected to differ per edition | 9 |
| Cypress fixture (`__tests__/shortcodes.md`), not content | 1 |

#7620 assumed everything outside its 24 was content whose editions genuinely
differ, per `docs/adr/0001`. That is true at one end of this list and false at
the other, which is why these are worth reconciling rather than leaving.

## Worklist

Tracked in `.ci/v1-shared-drift-manifest.json`. Remove an entry only when its
page actually moves to `content/shared/influxdb-v1/`. Current state
(`node .ci/scripts/check-v1-shared-drift.js --report`): 30 pairs, 3,001
diverging lines.

| diverging lines | page |
| --- | --- |
| 375 | `troubleshooting/frequently-asked-questions.md` |
| 329 | `administration/upgrading.md` |
| 328 | `query_language/spec.md` |
| 299 | `guides/hardware_sizing.md` |
| 295 | `write_protocols/line_protocol_tutorial.md` |
| 224 | `concepts/file-system-layout.md` |
| 167 | `guides/write_data.md` |
| 140 | `query_language/functions.md` |
| 99 | `query_language/manage-database.md` |
| 89 | `concepts/schema_and_data_layout.md` |
| 83 | `guides/calculate_percentages.md` |
| 80 | `tools/influx-cli/use-influx-cli.md` |
| 62 | `tools/influx_inspect.md` |
[...]

PR #7663, opened 2026-08-18
docs(v1): reconcile the 24 remaining v1 OSS/Enterprise page pairs into shared content

### LINKED ISSUE #7620: Phase A: reconcile the 24 remaining InfluxDB v1 OSS/Enterprise page pairs into shared content

jstirnaman:
Closes #7620

Stacked on #7622. Review that one first; this PR's base is its branch, so the
diff here is only the phase A work.

## What changed

Moves the 24 remaining duplicated InfluxDB v1 pages into
`content/shared/influxdb-v1/`, leaving front-matter stubs in both editions. The
v1 OSS and v1 Enterprise copies of these pages had drifted by 1,508 lines.

Also:

- Repopulates `.ci/v1-shared-drift-manifest.json` with 30 v1 pages that are still
  duplicated and were never tracked by it. #7662 covers those.
- Adds two style conventions to `AGENTS.md` and `DOCS-CONTRIBUTING.md`: the
  Google style guide applies to commit messages, issues, and status comments, and
  commit bodies use the pull request template's sections.

## Why

Neither edition was consistently the newer copy. Each had been swept by a
different mass edit, so the winner had to be chosen per hunk rather than per
edition:

- Enterprise copies had been through `flux fmt`. 422 of the 1,508 lines were
  formatting only.
- OSS copies had current `/flux/v0/` link paths. 21 of the 25 dead links were
  Enterprise-only pre-0.65 stdlib paths.

About eight hunks matched neither rule and were resolved against the Flux stdlib
reference or a running InfluxDB instance.

Three pages needed edition-specific handling rather than a single winner:

- `flux/installation.md` uses a conditional for the configuration guide link,
  because OSS has `administration/config` and Enterprise has
  `administration/configure/`.
- `concepts/glossary.md` keeps seven Enterprise-only terms in `show-in` blocks,
  and keeps both editions' definitions of replication factor.
- `flux/get-started/_index.md` and `flux/get-started/syntax-basics.md` regain the
  `influx` CLI sections Enterprise had dropped. The Flux REPL is not OSS-only;
  `content/enterprise_influxdb/v1/flux/execute-queries.md` documents it.

## Impact

Readers of both v1 editions. Corrections to errors that were present in both
copies:

- `histogram(buckets:)` does not run. The parameter is `bins`, and it takes
  float values.
- `geo.asTracks(sortBy:)` does not run. The parameter is `orderBy`.
- `flux/guides/sql.md` linked Enterprise readers to an anchor that does not
  exist. Both editions now link to
  `query_language/manage-database/#create-database`.
- `exists` on a statically defined record returns an error in InfluxDB 1.11 but
  not in 1.8. The page now documents the difference instead of stating it
  unconditionally.
- `concepts/glossary.md` dropped `grant`, `permission`, and `role`, which were
  commented-out headings with no content.

Renaming the glossary heading `replication factor` to `replication factor (RF)`
changed its anchor from `#replication-factor` to `#replication-factor-rf`. Three
pages that linked to the old anchor are updated. One of them,
`enterprise_influxdb/v1/guides/hardware_sizing.md`, had been linking into
`/influxdb/v1/` from an Enterprise page and now links within its own edition.

The drift checker, its test, the manifest, and `pr-v1-shared-drift.yml` are all
retained. #7620 originally called for deleting them once its list emptied. That
assumed these 24 were the last duplicated v1 pages, and 30 more exist.

## Verification

- `npx hugo --quiet` completes with no errors.
- No unresolved `/product/version/` placeholders in either v1 edition.
- Vale reports no errors across `content/shared/influxdb-v1/`.
- Every `/flux/v0/` link under `content/shared/influxdb-v1/` resolves, including
  four dead paths carried over from #7622.
- `node --test .ci/scripts/check-v1-shared-drift.test.js` passes 12/12.
- `yarn build:agent:instructions` and `yarn validate:agent-instructions` pass.
  No generated adapter changed.

Rendered HTML was checked per edition where a conditional decides what appears,
because Hugo exits 0 whether or not a conditional renders correctly. OSS shows
none of the Enterprise-only glossary terms and its own replication factor text;
Enterprise shows all of them and its own. The `user` entry renders as one list,
two bullets on OSS and three on Enterprise. `flux/installation.md` renders one
intact blockquote per edition with that edition's configuration link.

Runtime claims were checked against InfluxDB 1.8.10 and 1.11.8 in throwaway
containers rather than against reference pages, which is how the `histogram` and
`geo.asTracks` errors surfaced. Containers were removed afterward.


AGENTS.md
```diff
@@ -68,7 +68,9 @@ See [README.md](README.md) and the
 - For InfluxDB 3, shared directories contain the prose; product directories are
   usually thin stubs with `source:` references.
```
jstirnaman:
```suggestion
- Shared directories contain prose; product directories are often
   thin stubs with `source:` references to the shared content.
```

AGENTS.md
```diff
@@ -88,6 +90,11 @@ See [README.md](README.md) and the
 ## Documentation style
 
 - Follow the Google Developer Documentation Style Guide.
```
jstirnaman:
```suggestion
- Follow the Google Developer Documentation Style Guide for all aspects of communication, including structure and voice.
```

PR #7678, opened 2026-08-20
Change comments in line protocol syntax to be self-descriptive `#`

maximal:
Similar to #5821.


jstirnaman:
Thanks for the fix! Folded into #7705, which also splits the syntax template from the runnable example (so it's not mislabeled as `python` either) and applies the same fix to the other line-protocol reference pages. Your commit is preserved in that branch's history and credited via co-author trailer. Closing this in favor of #7705.

maximal:
@jstirnaman, thanks!

PR #7682, opened 2026-08-20
docs(influxdb3): document overwrite timing semantics for Core and Enterprise

reidkaufmann:
## What changed

In `content/shared/v3-line-protocol.md` (Duplicate points section):

- Appended a qualifier to the field-union paragraph (all products): overwrites are subject to timing conditions and are not a reliable last-value mechanism unless those conditions are met.
- Added an Enterprise-scoped warning: overwriting a point resolves non-deterministically if the overwrite arrives before the previous write has been persisted and referenced in a snapshot; guidance to leave enough delay between writes of the same point (at least 30 minutes with default settings), keep the compactor running, and route all writes of a given point through the same node in multi-ingest-node clusters.
- Added a Core-scoped warning: overwrites are not deterministic regardless of write spacing, because Core does not include the compaction process that establishes a durable order between data files containing the same point.
- Widened the `show-in` scope of the existing "Recommended patterns for last-value tracking" and "Anti-patterns to avoid" sections to include `core` and `enterprise`. The influxctl-based retention example remains cloud-dedicated-only and the performance section keeps its existing distributed scoping.
- The existing distributed warning is unchanged.

## Why

The section stated an unqualified last-write-wins behavior for duplicate points. On Core and Enterprise, overwrite resolution depends on persistence and snapshot timing, so the docs now state the conditions and point to the append-only patterns.

## Verification

- `npx hugo --quiet` builds cleanly.
- Rendered output checked per product: Core, Enterprise, Cloud Dedicated, and Clustered each render the base paragraph plus exactly one warning; Core and Enterprise now render the patterns and anti-patterns sections; retention and performance sections unchanged. (Cloud Serverless sources a different shared file and is unaffected.)

### DIFF
--- a/content/shared/v3-line-protocol.md
+++ b/content/shared/v3-line-protocol.md
@@ -308,7 +308,7 @@ Overwrites are subject to timing conditions described below; they are not a reli
 >
 > To ensure the last write wins, leave enough time between writes of the same point for
 > the earlier write to be persisted and referenced in a snapshot — with default settings,
-> at least 30 minutes — and ensure the compactor is running.
+> we recommend at least 30 minutes — and ensure the compactor is running.
 > Writes of the same point with insufficient delay yield undefined behavior.
 >
 > In clusters with multiple ingest nodes, the ordering of overwrites of the same point
--- a/content/shared/v3-line-protocol.md
+++ b/content/shared/v3-line-protocol.md
@@ -286,6 +286,7 @@ A point is uniquely identified by the table name, tag set, and timestamp.
 If you submit line protocol with the same table, tag set, and timestamp,
 but with a different field set, the field set becomes the union of the old
 field set and the new field set, where any conflicts favor the new field set.
+Overwrites are subject to timing conditions described below; they are not a reliable way to maintain a last-value view unless those conditions are met.
 
 {{% show-in "cloud-dedicated,clustered,cloud-serverless" %}}
 > [!Warning]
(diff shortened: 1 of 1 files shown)


content/shared/v3-line-protocol.md
```diff
@@ -286,6 +286,7 @@ A point is uniquely identified by the table name, tag set, and timestamp.
 If you submit line protocol with the same table, tag set, and timestamp,
 but with a different field set, the field set becomes the union of the old
 field set and the new field set, where any conflicts favor the new field set.
+Overwrites are subject to timing conditions described below; they are not a reliable way to maintain a last-value view unless those conditions are met.
```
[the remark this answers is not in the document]
reidkaufmann: Fixed in dc35058be2b5: replaced the qualifier with a neutral pointer ("Overwrite behavior is product-specific; see the warning below…") that does not imply a satisfiable timing condition exists for every product.

content/shared/v3-line-protocol.md
```diff
@@ -294,7 +295,46 @@ field set and the new field set, where any conflicts favor the new field set.
...
+> spaced: queries may return either version, and either version may be permanently
+> stored.
+> {{% product-name %}} does not include the compaction process that establishes
+> a durable order between data files containing the same point.
+>
+> To maintain a last-value view, use the append-only patterns below instead of
+> overwrites.
+{{% /show-in %}}
 
+{{% show-in "core,enterprise,cloud-dedicated,clustered,cloud-serverless" %}}
```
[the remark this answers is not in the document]
reidkaufmann: Fixed in dc35058be2b5: added the three missing closing fences. Note this was pre-existing — the examples render as one merged code block on Cloud Dedicated/Clustered today — so the fix also repairs the current pages, not just the newly widened ones.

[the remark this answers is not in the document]
reidkaufmann: Fixed in dc35058be2b5: scoped the anti-pattern to *short* write delays (heading and body). That keeps it true for the distributed products (the example is a 10-second delay) and for Core, and removes the contradiction with the Enterprise warning, which defines what a sufficient delay is. Chose this over nesting an enterprise-only `show-in` inside the widened block.

content/shared/v3-line-protocol.md
```diff
@@ -294,7 +295,44 @@ field set and the new field set, where any conflicts favor the new field set.
...
+{{% /show-in %}}
 
+{{% show-in "core" %}}
+> [!Warning]
+> #### Overwrites are not deterministic
+>
+> Overwriting a point (same table, tag set, and timestamp) is not reliable in
+> {{% product-name %}}, regardless of how far apart the writes are — concurrent or
+> spaced: queries may return either version, and either version may be permanently
+> stored.
```
philjb: I'd shorten:

"overwriting a point is not consistent; queries may return either version and either version may be permanently stored.

content/shared/v3-line-protocol.md
```diff
@@ -294,7 +295,44 @@ field set and the new field set, where any conflicts favor the new field set.
 > Overwriting duplicate points (same table, tag set, and timestamp) is _not a reliable way to maintain a last-value view_.
 > When duplicate points are flushed together, write ordering is not guaranteed—a prior write may "win."
 > See [Anti-patterns to avoid](#anti-patterns-to-avoid) and [Recommended patterns](#recommended-patterns-for-last-value-tracking) below.
+{{% /show-in %}}
+
+{{% show-in "enterprise" %}}
+> [!Warning]
```
philjb, line 301: I prefer to structure with what the system can do first: 

```
To ensure the last write wins, leave enough time between writes of the same point for
the earlier write to be persisted and referenced in a snapshot — with default settings,
we recommend at least 30 minutes.
Writes of the same point with insufficient delay yield inconsistent behavior.

In clusters with multiple...

Overwriting a point ...
```

compactor running should be assumed (otherwise isn't a quite broken setup)
"undefined behavior" is a compiler term meaning "anything might happen". Here one of the two points will win.

reidkaufmann: I was going for brevity, and technically it isn't wrong :smile:, but I agree that it could be more specific.

reidkaufmann: Applied in 387cbd3a9: recipe now leads, the hazard consequences are folded into the insufficient-delay sentence (so the explanation sits next to the claim), and dropped the compiler jargon for "resolve non-deterministically". Kept the compactor condition as a parenthetical assumption over dropping. I agree we don't expect customers to check the compactor before every overwrite, but we have had a few EARs with stuck/stopped compactors, so I leaned towards being accurate. I don't feel strongly -- will drop it if you insist!

content/shared/v3-line-protocol.md
reidkaufmann: Shortened in 387cbd3a9, keeping one clause: "regardless of the delay between writes" — that is the Core-vs-Enterprise splitting fact, and without it a Core reader could import the Enterprise delay guidance from the sibling block. Also kept "not reliable" over "not consistent": after persistence all readers agree on whichever version won, so the failure mode is unreliability rather than inconsistency.

philjb (approved):
i like it; dunno how much that counts

reidkaufmann:
> i like it; dunno how much that counts

Thanks! If @jstirnaman likes it too and no one objects, I'd say it counts enough!

jstirnaman (approved):

PR #7685, opened 2026-08-20
fix(platform): verify InfluxDB _internal fields against live v1.13.0

jstirnaman:
Closes #7557

## What changed

- Initially, a v1.12.4 test that only wrote one point and ran one query - not enough to trigger a TSM compaction cycle or a failed-query code path - called for the removal of `queryExecutor.queriesFailed`, `queryExecutor.queriesSlow`, and `tsm1_engine`'s six `compactionPlanner*` fields as "confirmed absent". I re-verified against live InfluxDB v1.13.0 (OSS) and v1.13.0rc5 (Enterprise) with a deliberately failing query; all 8 fields are present with real, non-zero values.
- Reclassified `shard.indexType`: it's a tag (confirmed via `SHOW STATS`' tag set), not a field. Moved it into a tags note on the measurement intro instead of a fake field heading.
- Removed an incorrect "(Enterprise only)" label from `write.pointReqLocal`; it's present on OSS too.
- Added two field headings that existed in the doc body but were missing from the TOC: `cluster.openConnections`, `httpd.fluxQueryRespBytes`.
- Fixed a TOC entry whose link text didn't match its own heading (`fluxQueryDurationNs` -> `fluxQueryReqDurationNs`).
- Fixed two TOC anchors that never resolved: mixed-case `#pointsWrittenOK`/`#valuesWrittenOK` don't match Hugo's lowercased, deduped heading IDs (the `httpd` and `write` measurements each have a same-named field, so the real targets are `#pointswrittenok-1` and `#valueswrittenok-1`).
- Replaced the rendered "pending confirmation" note with a `<!-- VERIFY (live instance) -->` comment for the four fields and one measurement that remain unconfirmed, tracked in #7684, instead of shipping speculative status to readers.

## Why

`httpd.valuesWrittenOK`, `write.pointsWrittenOK`, `write.subWriteDrop`, `write.valuesWrittenOK` are still documented by the Telegraf `influxdb` input plugin, so rather than deleting them outright they're flagged as unconfirmed pending engineering input (#7684) — they may be limited to older 1.x releases. That part still holds.

The original PR also removed 8 other fields as "confirmed absent," but that conclusion didn't survive a second look — see the correction above.

## Impact

Readers get a corrected field/tag list; the four still-uncertain fields no longer read as authoritative in the rendered page. The 8 restored fields (`queryExecutor.queriesFailed`/`queriesSlow`, `tsm1_engine`'s `compactionPlanner*` set) are real and now documented again.

## Verification

[...]

Issue #7557, opened 2026-07-22
Update v1 _internal documentation


davidby-influx:
https://docs.influxdata.com/platform/monitoring/influxdata-platform/tools/measurements-internal/

is very out of date for V1.  It should be updated, and equally important, linked to v1 OSS and v1 Enterprise documentation to be more easily findable

Let me know how I can help

jstirnaman:
Problem: https://docs.influxdata.com/platform/monitoring/influxdata-platform/tools/measurements-internal/
is completely stale.
Approach: Update it as a thin, opinionated wayfinder with relevant entrypoint links to different editions, prioritizing InfluxDB 3 Enterprise, InfluxDB 3 Cloud, InfluxDB Enterprise v1, InfluxDB Cloud v1--similar to the more recently updated `/platform` pages. Follow up with also-mentions of other InfluxDB editions. Link to common tools that help with monitoring, e.g. Telegraf, UIs, mention compatibility with other tools like Grafana.

jstirnaman:
Internal-only content (v1 `_internal` docs) isn't indexed in the public docs-search MCP, so I can't confirm current state that way. Needs public documentation.

---
_Docs audit (2026-08-19): cross-checked against current docs via the docs-search MCP. Flagging for verification against the issue's original ask.

PR #7686, opened 2026-08-20
refactor(platform): migrate platform monitoring docs into per-edition sections

jstirnaman:
Stacked on #7685 — merge that first.

## What changed

- Removed `content/platform/monitoring/influxdata-platform/` and `content/platform/monitoring/containers/` (11 files): `_internal` measurements, SHOW STATS/DIAGNOSTICS, Kapacitor measurements, monitoring dashboards, watcher-of-watchers setup, internal-vs-external monitoring, and Kubernetes monitoring.
- Added 8 shared source files under `content/shared/influxdb-v1/administration/monitor/` and 16 per-edition stub pages (`content/influxdb/v1/administration/monitor/` and `content/enterprise_influxdb/v1/administration/monitor/`) for the same topics, plus a new `performance-metrics-commands.md` page carrying the "Useful performance metrics commands" and `/metrics` endpoint sections that used to live on `server_monitoring.md`.
- Deleted `content/influxdb/v1/administration/server_monitoring.md`; its menu slot ("Monitor InfluxDB", weight 80, parent Administration) is now a new hub page at `content/influxdb/v1/administration/monitor/_index.md`, matching the hub that already existed on the Enterprise v1 side.
- Restructured both v1 OSS and v1 Enterprise monitor hubs into two sections: "Monitor InfluxDB itself" (internal metrics, SHOW STATS/DIAGNOSTICS, logs, diagnostics) vs. "Monitor other systems with InfluxDB" (the TICK-stack pages) — the two were previously flat and undifferentiated.
- Deleted `content/platform/monitoring/_index.md`; its content folds into a new "Monitoring" section on `content/platform/_index.md`, using the same two-way split, linking out per edition to whatever monitor page already exists (InfluxDB 3 Core/Enterprise query-system-data, Cloud Dedicated monitor-your-cluster, OSS v2 metrics reference, and the v1 hubs). Editions with no such page yet (Cloud Serverless, Clustered, Cloud early-access, Cloud TSM) are omitted rather than stubbed.
- Cross-links between the new shared pages use the `/product/version/` placeholder and `{{% show-in %}}` shortcode so each edition gets edition-correct link targets (e.g. the disable-`_internal` config link points at `/influxdb/v1/administration/config/` on OSS and `/enterprise_influxdb/v1/administration/configure/config-data-nodes/` on Enterprise).
- Updated every inbound link to the removed pages across the repo (chronograf, telegraf is generated and covered by aliases, release notes, troubleshooting, prometheus support pages, query language specs).
- Fixed two pre-existing typos surfaced by Vale while touching these files ("your your", "caused caused").
- Every removed page keeps its old URL(s) as a Hugo alias on its new destination.

### Per review feedback on #7685

- Removed the "if you're on a different edition, go here" wayfinder prose from `measurements-internal.md` (mirrors the same edit already made on #7685) — the site-wide banner on older editions already covers this, so per-page prose was redundant.
- Added `alt_links` frontmatter to `measurements-internal.md`, `show-stats.md`, and `show-diagnostics.md` (both editions) so the product/version dropdown routes to the closest real equivalent (InfluxDB 3 Core/Enterprise query-system-data, Cloud Dedicated monitor-your-cluster, OSS v2's metrics reference) instead of a generic product root. OSS v1 ↔ Enterprise v1 needs no `alt_links`: both editions publish these pages at the identical relative path, so the selector's default same-path fallback already resolves correctly between them. The other 5 shared pages (Kapacitor measurements, monitoring dashboards, watcher-of-watchers, internal-vs-external, monitor Kubernetes) have no equivalent outside the v1 TICK stack, so they were left without `alt_links`.

## Why

Per review decision on #7557: `content/platform/` should only carry an overview with links to individual editions, not product-specific how-to content. The old `platform/monitoring/` tree mixed two different concepts — monitoring InfluxDB itself vs. using InfluxDB/the TICK stack to monitor other systems — without distinguishing them, which made the hub pages hard to navigate and duplicated content that belongs with each edition's own admin docs.

## Impact

No content is lost; it moved to `content/shared` and per-edition pages where it's discoverable from each edition's own admin/administration navigation instead of a separate platform-level tree. All old URLs redirect via Hugo aliases. Cross-edition navigation now goes through the existing banner + product dropdown instead of hand-written links, matching the pattern already established on #7685.

Out of scope, tracked as follow-up: establishing a real `/admin/monitor/` URL convention for InfluxDB 3 editions (today it's scattered `query-system-data`/`monitor-your-cluster` pages at ordinary `admin/` paths); filling the monitoring gap for Cloud Serverless, Clustered, Cloud (early access), and InfluxDB Cloud (TSM), none of which have any "monitor the instance itself" documentation today.

## Verification

- `npx hugo --quiet --gc` builds clean.
- `link-checker map` + `link-checker check` on every changed/added file and the pages that reference them — no internal link or fragment errors (only pre-existing external-URL noise and expected not-yet-deployed self-links).
- `.ci/vale/vale.sh --config=.vale.ini --minAlertLevel=error` passes on every touched file.
- Spot-checked redirects: `/platform/monitoring/`, `/platform/monitoring/influxdata-platform/`, `/influxdb/v1/administration/server_monitoring/`, and each individual old tool URL all resolve to their new destination.
- Confirmed menus auto-trim: no leftover "InfluxData Platform" or `monitor-platform` menu references anywhere after deleting the pages (this repo's menus are built purely from surviving pages' frontmatter, no separate menu data file to prune).
- Confirmed each new `alt_links` entry renders the exact target URL in the built product-selector dropdown markup (checked all 6 files across both editions).


sanderson (approved):

content/chronograf/v1/administration/managing-influxdb-users.md
jstirnaman:
```suggestion
7. Add permissions to the role in the `Permissions` section. You will see a list of databases and all permissions for that database. Select
```

Issue #7667, opened 2026-08-19
api-docs: tags.yml links to a QuerystringAuthentication anchor that no longer exists in the v2 and Cloud specs


jstirnaman:
`api-docs/influxdb/v2/tags.yml:9` and `api-docs/influxdb/cloud/tags.yml:9` both contain:

```markdown
- [Querystring authentication](#section/Authentication/QuerystringAuthentication)
```

That anchor is generated from `components.securitySchemes`. Current upstream `contracts/ref/oss.yml` on `influxdata/openapi` `docs-release/influxdb-oss` declares only `BasicAuthentication` and `TokenAuthentication` — no `QuerystringAuthentication`. So the link has no target.

## Verification

Running the pipeline against current sources:

```sh
cd api-docs && bash getswagger.sh v2
cd .. && node api-docs/scripts/dist/post-process-specs.js

python3 -c "import yaml;print(sorted(yaml.safe_load(open('api-docs/_build/influxdb/v2/influxdb-oss-v2-openapi.yaml'))['components']['securitySchemes']))"
# ['BasicAuthentication', 'TokenAuthentication']

grep -n QuerystringAuthentication api-docs/_build/influxdb/v2/influxdb-oss-v2-openapi.yaml
# 76:      - [Querystring authentication](#section/Authentication/QuerystringAuthentication)
```

One occurrence in the built spec — the `tags.yml` link itself, with nothing to point at.

## Why the currently-committed spec hides this

`api-docs/influxdb/v2/influxdb-oss-v2-openapi.yaml` still carries a `QuerystringAuthentication` entry under `components.securitySchemes`, left over from an older upstream contract. But even there it has **zero `$ref`s** and is absent from the top-level `security:` block, so it's already an orphan definition. Whether the rendered page currently produces a working anchor from an unreferenced scheme is worth checking; either way, the next regeneration removes the definition and the link definitely breaks.

Note this is unrelated to #7175, which is about per-tag page URLs not resolving in the built site.

## Fix options

1. Drop the bullet from both `tags.yml` files, if querystring auth is genuinely no longer supported for these APIs.
2. Restore `QuerystringAuthentication` upstream in `influxdata/openapi` if the v1-compatibility endpoints still accept `?u=&p=`, and keep the link.

Option 2 needs a product answer — the v1-compat `/query` and `/write` operations in the OSS v2 spec do still document inline `u` and `p` query parameters, which suggests the auth method exists but is no longer declared as a security scheme.

[...]

Issue #7666, opened 2026-08-19
API reference: upstream INFLUXDB_DOCS_URL substitution pins v2 links to /influxdb/latest and emits broken URLs


jstirnaman:
Tracking issue for defects in `influxdata/openapi` that surface in the InfluxDB OSS v2 API reference. The fix belongs upstream; this tracks it here because the symptom is visible on our published pages.

I couldn't file this in `influxdata/openapi` directly — the GitHub App has read access to that repo but no issue-write permission (`403 Resource not accessible by integration`). Someone with access should port this over.

All findings reproduce on `influxdata/openapi` `master` (`e48e733`) and `docs-release/influxdb-oss` (`74ffccc`), which is the branch `api-docs/getswagger.sh` fetches for OSS v2.

## Root cause

`influxdata/openapi` PR #603 ("chore: use /latest alias for OSS URL substitutions", 2022-11-08) made `{{% INFLUXDB_DOCS_URL %}}` expand to `https://docs.influxdata.com/influxdb/latest` during OSS contract generation. Our `replace-docs-url-shortcode` decorator (`api-docs/openapi/plugins/docs-plugin.cjs`) can only rewrite shortcodes that *survive* into the contract — it can't touch a URL that upstream already expanded.

## 1. Pagination links pinned to `/latest`

`src/common/parameters/After.yml:12` and `src/common/parameters/Offset.yml:8` both source as:

```yaml
For more information about pagination parameters, see [Pagination]({{% INFLUXDB_DOCS_URL %}}/api/#tag/Pagination).
```

and arrive in `contracts/ref/oss.yml` (lines 56, 85) as `https://docs.influxdata.com/influxdb/latest/api/#tag/Pagination`.

Effect on our v2 reference: these two links stay absolute and pinned to `/latest`, while every other docs link on the same page is rewritten to a relative `/influxdb/v2/...`. So one page mixes link styles, and the `/latest` ones drift to whatever the current OSS version is rather than the version the page documents.

The cloud contract expands the same source to `https://docs.influxdata.com/influxdb/cloud/api/#tag/Pagination`, which is version-correct — this is specific to the OSS `/latest` alias.

## 2. Duplicated path segments — broken URLs

`src/oss/tags.yml:148-149` already include a version segment after the shortcode, so expansion doubles it:

| Source | Generated |
| --- | --- |
[...]

PR #655, opened 2019-12-05
'http.get' timeout parameter

sanderson:
Closes #648 

Add the new `timeout` parameter on the experimental `http.get()` function.

- [x] Tests pass (no build errors)
- [x] Rebased/mergeable

PR #7697, opened 2026-08-24
feat(api): automate Cloud and OSS v2 spec sync

### LINKED ISSUE #7668: api-docs: committed OpenAPI specs are stale artifacts of the old pipeline — re-baseline before the next spec update
The specs committed under `api-docs/**/` no longer match what the current pipeline produces. Regenerating with no source change rewrites them wholesale:

| Spec | Diff with no source change |
| --- | --- |
| `api-docs/influxdb/v2/influxdb-oss-v2-openapi.yaml` | ~685 insertions / ~693 deletions |
| `api-docs/influxdb/cloud/influxdb-cloud-v2-openapi.yaml` | ~779 lines changed |

This blocks clean review of any real spec change — for example the pending `onConflict` parameter from `influxdata/openapi#655`, which is a single query parameter but would arrive buried in ~1400 lines of churn.

## The churn is expected catch-up, not a regression

I verified each component. Nothing is lost.

**1. `info` / `servers` / `tags` reset — by design.** `api-docs/openapi/plugins/docs-plugin.cjs:1-6` documents that the `set-info` and `set-servers` decorators were deliberately removed so `post-process-specs.ts` is the single source of truth. Bundler output is *supposed* to carry raw upstream values; overlays are reapplied into the gitignored `api-docs/_build/`. Confirmed: after running `post-process-specs.js`, the `_build` `info` block is identical to the committed spec. The committed file has overlays baked in only because it predates that refactor.

`servers` still differs after post-processing — committed `[{url: /}]` vs `_build` `[{url: 'http://localhost:8086', description: 'Local InfluxDB instance'}]` — because `influxdb/v2/content/servers.yml` changed since the last regeneration. The `_build` value is the intended one.

**2. `components.parameters` inlining — upstream, not the bundler.** Current `contracts/ref/oss.yml` declares only `After`, `Descending`, `Limit`, `Offset`, `SortBy`, `TraceSpan`. It does not declare `Accept`, `AcceptEncoding`, `Content-Type`, `AuthV1Username`, `AuthV1Password`, `V1Database`, `V1Query`, `V1RetentionPolicy`, or `V1Epoch` — and neither does the `influxdb-oss-v2.7.0` tag. Upstream now inlines those on the v1-compatibility operations.

Dereferencing both specs and comparing parameter sets across all 233 shared operations: **206 identical**; the other 27 differ only in parameter ordering or in the `After`/`Offset` description string. `GET /query` carries the same 10 parameters in both.

**3. `QuerystringAuthentication` removal** — tracked separately in #7667.

**4. `After`/`Offset` description change** — upstream `INFLUXDB_DOCS_URL` substitution, tracked in #7666.

## Why the committed spec looks both new and old

It contains the recent gzip-compression-level parameter (`influxdata/openapi#652`) *and* components that upstream dropped years ago. That's consistent with the file having been hand-patched in place rather than regenerated — #7175 notes that PR #7173 landed "a 10-line surgical patch" directly on `api-docs/influxdb/v2/influxdb-oss-v2-openapi.yaml`. Worth confirming whether other manual patches are riding in these files that a clean regeneration would silently drop.

## Also: `@redocly/cli` is unpinned

`api-docs/getswagger.sh` invokes `npx @redocly/cli` with no version constraint; npx currently resolves 2.46.0. Redocly also logs:

```
Deprecated plugin format detected: docs
```

for `api-docs/openapi/plugins/docs-plugin.cjs`, which is written against the Redocly 1.x plugin API.

To be clear, I did **not** find the version bump to be the cause of any of the deltas above — those are all pipeline-refactor staleness or upstream source changes. But an unpinned major-version dependency in the generation path means the next contributor's regeneration can differ from CI's for reasons unrelated to their change.

## Proposed work

1. Audit the committed specs for hand-patches that aren't in upstream sources; port anything still needed to `influxdata/openapi` or to the overlay/tag config files.
2. Pin `@redocly/cli` to an explicit version in `api-docs/package.json` and call the pinned binary from `getswagger.sh`.
3. Migrate `docs-plugin.cjs` to the Redocly 2.x plugin format to clear the deprecation warning.
4. Re-baseline every product spec in a single dedicated PR, reviewed as "generated output, verified no semantic change" using the dereferenced-parameter comparison above.

After that, `influxdata/openapi#655` regenerates as a one-parameter diff.

## Reproduce

```sh
cd api-docs && bash getswagger.sh v2   # use bash — the script is not POSIX-clean, sh fails at line 29
cd .. && git diff --stat api-docs/influxdb/v2/influxdb-oss-v2-openapi.yaml
```

jstirnaman:
Closes # — none. Related: #7666, #7667, #7668, [influxdata/openapi#655](https://github.com/influxdata/openapi/pull/655), and [influxdata/influxdb#27585](https://github.com/influxdata/influxdb/pull/27585).

## What changed

Adds two independent scheduled sync workflows:

| Product | Schedule | OpenAPI contract | Generated PR branch |
| --- | --- | --- | --- |
| Cloud v2 | Monday at 06:30 UTC | `contracts/ref/cloud.yml` | `sync/openapi-cloud-v2-spec` |
| OSS v2 | Monday at 07:00 UTC | `contracts/ref/oss.yml` | `sync/openapi-oss-v2-spec` |

Each workflow:

- supports manual dispatch;
- resolves the current `influxdata/openapi/master` commit;
- regenerates from that exact SHA rather than a moving branch URL;
- records the SHA in the generated PR;
- applies the appropriate `source:sync` and product labels;
- recommends conditional waiting and release labels;
- recommends the implementation and release records to link; and
- opens or updates its own PR only when its generated spec changes.

### Clean regeneration (resolves #7668)

Moves OSS v2 and Cloud v2 to a raw-mirror pipeline so a scheduled sync diffs
cleanly against `influxdata/openapi` instead of mixing upstream changes with
decorator churn:

- Adds a `docs/mirror` redocly config (`api-docs/openapi/plugins/docs-plugin.cjs`)
  with no decorators, used only by `influxdb/v2/.config.yml` and
  `influxdb/cloud/.config.yml`. `getswagger.sh` now commits these two specs as
  a plain upstream bundle — `$ref`-resolved, unmodified otherwise. Every other
  product keeps `docs/all` and decorates at commit time as before.
- Ports the five retired decorators (delete-servers, remove-private-paths,
  strip-version-prefix, strip-trailing-slash, replace-docs-url-shortcode) into
  `api-docs/scripts/post-process-specs.ts` as plain TypeScript, gated to
  `MIRROR_PRODUCT_PATHS`, applied at build time into the gitignored `_build/`
  that `generate-openapi-articles.ts` already reads exclusively — so published
  output is unaffected in kind.
- Fixes the doc-link rewrite along the way: it now walks every string in the
  spec, not just `description` fields, so `externalDocs.url` gets corrected
  too. The committed OSS v2 spec carried 229 `/influxdb/latest/` links that
  should read `/influxdb/v2/` — a one-time hand patch from years ago
  (`4403dd9a1`) that every regen silently reverted because no transform
  reproduced it. This re-baseline regen fixes all of them as a repeatable
  transform instead.
- Re-baselines both committed specs against the new pipeline in this PR, so
  the first scheduled sync after merge is a real, small diff — not another
  wholesale rewrite.

## Why

Cloud v2 and OSS v2 ship independently.
A combined generated PR could mix API changes with different release timing and
prevent reviewers from merging one product safely.

Releases are infrequent enough that separate weekly docs-v2 checks provide a
simpler solution than cross-repository release integration.
The generated PRs preserve the docs-v2 human approval gate.

The `onConflict` change shows why this gate is necessary:
`influxdata/openapi#655` merged on August 8, 2026, but the corresponding server
implementation in `influxdata/influxdb#27585` remains an open draft against
`main-2.x`.
A merged OpenAPI change is therefore a change proposal, not evidence that the
behavior is implemented or released.

The OSS workflow uses the current OpenAPI `master` contract as a change
proposal instead of relying on the manually maintained
`docs-release/influxdb-oss` branch.
Its PR instructs reviewers to verify that every API change has shipped before
merge and to regenerate from an exact tag or commit if only part of the proposed
contract is released.

Baking decorators into the committed spec meant every regen diff mixed real
`influxdata/openapi` changes with decorator output, defeating clean review —
the exact problem #7668 reported. Moving decorators to build time makes the
commit a pristine mirror, so a scheduled sync PR shows reviewers only what
actually changed upstream.

## Impact

The workflows update only these generated files:

- `api-docs/influxdb/cloud/influxdb-cloud-v2-openapi.yaml`
- `api-docs/influxdb/v2/influxdb-oss-v2-openapi.yaml`

This PR's own commit already re-baselines both files against the new
pipeline, so reviewers should expect a large diff here once, converting the
committed specs from decorated output to a raw upstream mirror. After merge,
each scheduled sync PR should be a small, real diff.
Reviewers must confirm that each API change is available in the corresponding
product before merging.

Only `influxdb/v2` and `influxdb/cloud` are affected by the pipeline change.
Other products (`influxdb3/*`, `influxdb/v1`, `enterprise_influxdb/v1`) keep
`docs/all` and decorate at commit time, unchanged.

The scheduled workflows do not run until they reach the default branch.

## Verification

- Parsed both workflow files successfully as YAML.
- Verified that each workflow resolves and records an exact OpenAPI SHA.
- Verified that the generated OSS PR distinguishes a merged OpenAPI change from
  an implemented and released server change.
- Verified recommended labels against `data/labels.yml` and product labels
  against `data/products.yml`.
- Verified that both generated PR templates identify the engineering and release
  records reviewers should link.
- Verified that each workflow passes the SHA-based raw URL to
  `getswagger.sh -b`.
- Verified that each workflow has one fixed PR branch and one
  `create-pull-request` step.
- Verified the workflows use valid `getswagger.sh` product names:
  `cloud-v2` and `v2`.
- Compared the pinned action SHAs with
  `.github/workflows/sync-client-library-release-notes.yml`.
- Did not run the workflows in GitHub Actions because scheduled and manual
  workflows run from the default branch.
- `tsc -p api-docs/scripts/tsconfig.json` compiles clean.
- `node api-docs/scripts/dist/test-post-process-specs.js`: 52/52 passing,
  including 2 new tests for the ported mirror transforms and for non-mirror
  products being unaffected.
- Regenerated both specs via `getswagger.sh v2` / `cloud-v2`: committed
  source now carries raw `docs.influxdata.com` links, confirming it mirrors
  upstream instead of decorated output.
- Ran `post-process-specs.js` for both products: `_build/` output has zero
  `/influxdb/latest/` links remaining (was 229 for OSS v2, 5 for Cloud v2).
- Ran `generate-openapi-articles.js --skip-fetch oss-v2 cloud-v2` end to end:
  succeeded, all generated output lands in gitignored paths.
- Ran `yarn build:api-docs` and `yarn hugo serve`. Visually inspected `/static` files using `jq`.

## Checklist

- [x] Rebased/mergeable
- [x] Local build passes (`npx hugo --quiet`) — not applicable; no site files changed (all generated content lands in gitignored paths)


jstirnaman:
CI status on `1c4b2dd`: `block-ephemeral-docs` **failed**; everything else passed or skipped. ([job](https://github.com/influxdata/docs-v2/actions/runs/32701416872/job/97353540460))

That failure is the workflow doing its job, not a defect. `.github/workflows/block-ephemeral-docs.yml` fails whenever `PLAN.md` or `HANDOVER.md` exists on a PR targeting `master`, and `PLAN.md` is the only file this PR adds. It will stay red for as long as the plan lives here.

I'm not pushing a fix, because the only fix is deleting the PR's entire contents — a call for a maintainer, not something to do automatically. Two ways forward:

1. **Leave it as a draft** until `influxdata/openapi#655` merges, then replace `PLAN.md` with the regenerated spec in the same PR. The check goes green when the plan file goes away.
2. **Close this PR** and move the contents somewhere durable — #7666, #7667, and #7668 already carry the three findings, so the unique value left in `PLAN.md` is the regeneration runbook and the release-timing constraint.

I'd lean toward 1, since the branch and runbook are only useful until #655 lands. Happy to do either.

---

jstirnaman:
⚠️ The SHA-pinning commits change which **branch** the OSS spec comes from, and that just became load-bearing — `influxdata/openapi#655` merged to `master` about an hour ago.

## What changed

`sync-openapi-oss-v2-spec.yml` resolves `refs/heads/master`:

```sh
git ls-remote --exit-code https://github.com/influxdata/openapi.git refs/heads/master
```

then passes it as `-b`. In `api-docs/getswagger.sh`, `-b` sets **both** base URLs (line 87):

```sh
b)
  baseUrl=$OPTARG
  baseUrlOSS=$OPTARG
  ;;
```

and the `v2` target reads `$baseUrlOSS` (line 277). So `getswagger.sh v2 -b <master-sha-url>` fetches `contracts/ref/oss.yml` from **master**, not from the `docs-release/influxdb-oss` default on line 24. Pinning to an immutable SHA is a clear improvement; the branch switch is the part worth confirming.

## Why it matters right now

| | |
| --- | --- |
| `openapi` `master` (`d90ba20`) | **has** `onConflict` — #655 merged |
| `openapi` `docs-release/influxdb-oss` | does **not** have it |
| `influxdata/influxdb#27578` (server-side `--on-conflict`) | still **open**, no linked PR |

The next OSS sync — Monday 07:00 UTC, or any `workflow_dispatch` — will therefore open a PR adding `onConflict` to the published OSS v2 reference, documenting a parameter that no released InfluxDB OSS server implements. The two contracts were byte-identical until #655 landed, so this wouldn't have shown up in testing before today.

This also drops the `docs-release/influxdb-oss` gate that `api-docs/README.md` ("InfluxDB OSS v2 version") documents as the reason OSS reference docs track a release branch rather than `master`.

## Is this intended?

The commit message (`refactor(api): sync OSS v2 from pinned master revision`) reads deliberate, and the new PR body — "confirm that every API change is available in a released InfluxDB OSS v2 version… If only a subset has shipped, regenerate from the appropriate OpenAPI tag or commit instead" — is consistent with moving the gate from the branch to the reviewer. That's a coherent design; it just trades an automatic guarantee for reviewer discipline on every sync, and the first PR it produces will be one that shouldn't merge.

Two options:

1. **Keep the branch gate.** Resolve `refs/heads/docs-release/influxdb-oss` instead of `refs/heads/master` in the OSS workflow. Still SHA-pinned and immutable, still reproducible — the one-line change is the ref in `git ls-remote`. Cloud stays on `master`, which is correct for it.
2. **Keep master as the source** and accept the reviewer gate, but say so in `api-docs/README.md` so the README and the workflow don't disagree about where OSS specs come from.

I'd go with 1 — it keeps the invariant enforced by the pipeline rather than by whoever reviews the sync PR at 7am Monday. Happy to push either.

---

jstirnaman:
Checked all three Copilot findings against the repo. All are technically correct, but two of them describe **pre-existing repo-wide patterns** rather than anything these workflows introduce — worth knowing before deciding where to fix them.

### 1. No CI on the generated sync PRs — the most consequential, and Copilot suppressed it

`create-pull-request` with `GITHUB_TOKEN` cannot trigger `on: pull_request` workflows. This repo has **12** of them:

`auto-label`, `block-ephemeral-docs`, `doc-review`, `pr-feedback-links`, `pr-link-check`, `pr-lockfile-lint`, `pr-preview`, `pr-release-check`, `pr-remark-check`, `pr-render-check`, `pr-vale-check`, `test`

So a weekly sync PR arrives with **zero checks** — no link check, no Vale, no render check, no preview — which is a real gap for a PR whose entire purpose is reviewing generated output.

Pre-existing, though: `sync-plugins.yml` (explicitly `token: ${{ secrets.GITHUB_TOKEN }}`), `sync-client-library-release-notes.yml`, and `check-pinned-deps.yml` all have the same limitation today. Fixing it needs an App token or PAT, so it's an infra decision, and arguably one to make once for all four workflows rather than only here.

### 2. `persist-credentials` — valid hardening, also the repo-wide default

Correct that `actions/checkout` leaves a write-scoped credential in `.git/config`, that `npx @redocly/cli` then executes an unpinned package in the same job, and that `create-pull-request` authenticates separately so it doesn't need that credential. `persist-credentials: false` is a safe one-line addition to both workflows.

Also pre-existing — none of the three sync workflows above set it. But the specific combination here (persisted write credential + unpinned remote code execution in the same job) is what makes it worth acting on, and that combination comes from finding 3.

### 3. Unpinned `@redocly/cli` — the root cause, and the highest-leverage fix

Copilot reached the same conclusion I did in #7668 independently. It's worth stating plainly that this one finding drives the other two:

- It's why SHA-pinning the contract doesn't actually make the sync reproducible — a Redocly release can rewrite the spec with the OpenAPI SHA unchanged, producing a sync PR that looks like an upstream change but isn't.
- It's what turns finding 2 from theoretical into concrete, since the untrusted code and the write credential share a job.

Pinning it in `api-docs/package.json` and calling the local binary from `getswagger.sh` addresses reproducibility and defuses most of finding 2. Caveat from #7668: the pinned version has to be chosen deliberately, because the committed specs were generated with a Redocly 1.x CLI and the current `npx` resolves 2.46.0 — so pinning and re-baselining should land together, before these schedules go live.

### Suggested order

1. Pin `@redocly/cli` + re-baseline (#7668) — unblocks everything else.
2. Add `persist-credentials: false` to both workflows (and ideally the other three sync workflows).
3. Decide separately, repo-wide, whether sync PRs should run CI via an App token.

Happy to take any of these. 1 and 2 are the ones I'd do first; 3 needs a secret I can't provision.

---

jstirnaman:
Verified the pin commit (`49d8669`) in a clean worktree. It holds up, and it also lets me replace my earlier argument about the OSS source branch with an actual measurement.

### Pin verification — all claims confirmed

| Check | Result |
| --- | --- |
| `yarn install --frozen-lockfile` from clean `node_modules` | ✅ succeeds (2.7s) |
| `node_modules/.bin/redocly --version` | ✅ `2.47.0` |
| `bash getswagger.sh v2` reproduces the committed baseline | ✅ **zero diff** — idempotent |

I'd expected the lockfile entry to be incomplete, since it adds `@redocly/cli@2.47.0` with no transitive dependencies. That was wrong — the package resolves standalone and `--frozen-lockfile` is satisfied. The guard in `postProcess` that errors with "Run 'yarn install' in api-docs first" is a nice touch; it fails loudly rather than silently falling back.

### What the first OSS sync PR will actually contain

Same worktree, same pinned binary, only the source differs:

```sh
# defaults → docs-release/influxdb-oss
bash getswagger.sh v2
#   → 0 lines changed, no onConflict

# what the workflow runs → master @ d90ba20
bash getswagger.sh v2 -b "https://raw.githubusercontent.com/influxdata/openapi/d90ba20…"
#   → +9 lines, onConflict present
```

So the first sync PR is a clean 9-line diff adding `onConflict` to `PostRestoreBucketMetadata` — nothing else, now that the re-baseline has landed. That is a genuinely good outcome for a first automated run.

I'd flagged this earlier as a possible regression. Having now seen `refactor(api): sync OSS v2 from pinned master revision` alongside the "confirm that every API change is available in a released InfluxDB OSS v2 version… regenerate from the appropriate OpenAPI tag or commit instead" guidance and the new `waiting:pr` / `release:pending` labels, the design reads as deliberate: propose from `master`, gate at review. My measurement confirms it behaves that way rather than contradicting it. `influxdata/influxdb#27578` is still open, so that first PR should land `waiting:pr` + `release:pending` and sit — exactly what the labels are for.

### One thing left

`api-docs/README.md` ("InfluxDB OSS v2 version") still says OSS reference docs are generated from the `docs-release/influxdb-oss` branch, which is now only true for a local `getswagger.sh v2` run, not for the automated sync. Worth a short update so the README and the workflow don't disagree — happy to push it if useful.

---

jstirnaman:
Verified `7bd1659` by building `_build/` at `49d8669` and at `7bd1659` in separate worktrees and diffing the results. One thing reviewers should know: **this refactor is not output-neutral — it fixes 211 links.**

### Published output changes

| Build | `/influxdb/latest/` links in OSS v2 | in Cloud v2 |
| --- | --- | --- |
| `49d8669` (decorators at bundle time) | **206** | **5** |
| `7bd1659` (decorators at build time) | **0** | **0** |

`applyMirrorTransforms` rewrites `https://docs.influxdata.com/influxdb/latest/` and bare `/influxdb/latest/` to the product root, which the old bundle-time `replace-docs-url-shortcode` decorator never did — it only handled the `{{% INFLUXDB_DOCS_URL %}}` form. So the v2 reference was shipping 206 links pointing at `/influxdb/latest/` instead of `/influxdb/v2/`. That's the docs-v2 half of #7666, now fixed here.

Nice detail: the `\{\{%\s*INFLUXDB_DOCS_URL\s*%\}\}` pattern uses `\s*`, so it also catches the malformed `{{% INFLUXDB_DOCS_URL%}}` tag in upstream `src/common/paths/me_password.yml:50`. Both builds show 0 literal shortcodes, but the new one is robust to that upstream typo rather than incidentally unaffected by it.

Everything else in the built output is identical — same paths, same tags, no components added or removed. The 44 differing paths and the `After`/`Offset`/`securitySchemes` component changes are all this URL rewrite.

### Tests

All 52 pass, including the 7 new mirror-transform cases (`14a`–`14g`) and — the part I'd have asked for — `15a`–`15c` asserting non-mirror products are left untouched, so Core/Enterprise/Clustered aren't silently affected by the `docs/mirror` split.

### Net effect on the original goal

The committed v2 specs are now clean mirrors of upstream, so a sync diff shows upstream content drift and nothing else. Combined with the Redocly pin, the first OSS sync PR should be exactly the 9-line `onConflict` addition I measured earlier.

I'll add a note to #7666 narrowing its scope: with this merged, the remaining upstream defects (doubled `/influxdb/latest/influxdb/latest/` segments in `src/oss/tags.yml:148-149`, and the malformed tag) no longer affect docs-v2 output — the doubled URLs never reached our build anyway, since our own `tags.yml` overrides that tag description. #7666 is now purely about other consumers of the raw contract.

---

jstirnaman:
Steps to prevent publishing the unimplemented param until the `influxdb` work is completed:
  
  1. Merge #7697.
  2. gh workflow run sync-openapi-oss-v2-spec.yml — dispatches immediately, resolves
     current master, opens/updates sync/openapi-oss-v2-spec PR with onConflict diff.
  3. On that generated PR: add waiting:pr (blocked on influxdb#27585 merging) or
     waiting:engineering if status unclear, plus release:pending.
  4. Comment linking influxdata/openapi#655 and influxdata/influxdb#27585 so reviewers
     see why it's held.

jstirnaman:
That plan works — I measured step 2's output earlier and it's a clean 9-line diff adding `onConflict` to `PostRestoreBucketMetadata`, nothing else. All four labels you'd use (`waiting:pr`, `waiting:engineering`, `release:pending`, `release:ready`) exist in the repo.

Two things about holding that generated PR open for a long stretch, since `influxdb#27585` is still a draft against `main-2.x` and then needs a 2.x release after it merges.

**1. The generated PR won't have CI.** `create-pull-request` uses `GITHUB_TOKEN`, which by design can't trigger `on: pull_request` workflows — all 12 of this repo's PR checks will be absent, including `pr-link-check`, `pr-vale-check`, `pr-render-check`, and `preview`. That's the one Copilot finding still open, and it's pre-existing for `sync-plugins`, `sync-client-library-release-notes`, and `check-pinned-deps` too. It matters more here than usual: the held PR is a spec change nobody will have seen rendered, sitting until someone merges it months later on the strength of a review done today. Worth either resolving the App-token question first, or running `bash api-docs/generate-api-docs.sh` locally against that branch at merge time.

**2. The held PR's diff will grow silently.** Each Monday run pushes to the same fixed `sync/openapi-oss-v2-spec` branch and updates the existing PR. So while it sits waiting on the release, every unrelated `influxdata/openapi` `master` change lands in it too. The labels persist, the title stays the same, and the diff quietly stops being "just `onConflict`."

Suggestion for step 4's comment: state the SHA the PR was reviewed at, so whoever merges it knows to re-check the diff rather than trusting the original review. Alternatively, dispatch on demand once `#27585` merges instead of letting the weekly schedule accumulate into a held PR — you'd get the same result with a diff that matches its review.

---

PR #7705, opened 2026-08-25
fix(content): use text+lp fences for line protocol syntax examples

jstirnaman:
## What changed
- Adds a Hugo render hook for lp fences. It preserves source text, highlights field-family syntax, and falls back to escaped plain text for malformed input.
- Adds a dependency-free lp validator to the code-block linter and makes invalid lp fences blocking CI errors. Intentionally invalid fixtures can use lint="false".
- Adds formatter fixtures; validator, extractor, and CLI tests; Cypress coverage; and documentation and generated instruction updates.
- Retains the earlier syntax-fence correction for issues #7678 and #5821.

## Why
Line protocol examples need syntax-aware rendering and validation. The former js and sh labels incorrectly represented line protocol as executable JavaScript or shell code, and the new lp support prevents malformed examples from silently reaching the docs.

## Impact
Docs rendering now gives lp fences dedicated syntax markup. CI rejects malformed lp fences unless they are explicitly exempted. No runtime dependency or product behavior changes.

## Verification
- yarn test:lint-codeblocks (65 passing)
- yarn test:render-regression (12 passing)
- yarn build:agent:instructions
- yarn validate:agent-instructions
- Pre-commit and pre-push hooks passed.

Closes #7678
Closes #5821

### DIFF
--- a/layouts/partials/lp/highlight.html
+++ b/layouts/partials/lp/highlight.html
@@ -1,0 +1,16 @@
+{{- $valid := true -}}
+{{- range split .Inner "\n" -}}
+  {{- $quotes := len (findRE `&#34;` (. | htmlEscape)) -}}
+  {{- if ne (mod $quotes 2) 0 -}}
+    {{- $valid = false -}}
+  {{- end -}}
+{{- end -}}
+{{- $code := .Inner | htmlEscape -}}
+{{- if $valid -}}
+  {{- /* Apply replacements to escaped source, then mark only the generated
+        markup safe. An unterminated string falls back to plain text. */ -}}
+  {{- $code = replaceRE `([[:alnum:]_.-]+)::([^=,[:space:]]+)(=)` `<span class="nx">$1</span><span class="p lp-family-delimiter">::</span><span class="na">$2</span><span class="o">$3</span>` $code -}}
+  {{- $code = replaceRE `&#34;[^&#34;\n]*&#34;` `<span class="s">$0</span>` $code -}}
+  {{- $code = replaceRE `(?m)^#.*$` `<span class="c">$0</span>` $code -}}
+{{- end -}}
+{{- printf "<div class=\"highlight\"><pre tabindex=\"0\" class=\"chroma\"><code class=\"language-lp\" data-lang=\"lp\">%s</code></pre></div>" $code | safeHTML -}}
--- a/.agents/instructions/content-review.md
+++ b/.agents/instructions/content-review.md
@@ -60,6 +60,8 @@ Products defined in [data/products.yml](../../data/products.yml):
 ## Code Blocks
 
 - Use `python` not `py` for language identifiers (pytest requirement)
+- Use `lp` for line protocol examples. `lp` fences are blocking syntax checks;
+  use `{lint="false"}` only for examples that intentionally show invalid input.
 - Long options in CLI examples (`--output` not `-o`)
 - Keep lines within 80 characters
 - Include language identifier on fenced code blocks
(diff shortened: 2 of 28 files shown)


jstirnaman:
~~## Known gap: `lp` fence has no syntax highlighting~~

Converting these blocks from `js`/`sh` to `lp` fixes the mislabeling (they were never valid JS/shell), but introduces a visual regression worth tracking.

**What's lost:** `assets/styles/layouts/_syntax-highlighting.scss` has a `.language-js` rule (originally added for Flux duration/time values, e.g. `1h30m`) that colors a number token immediately followed by an identifier token:

```scss
// Javascript / Flux specific styles (duration and time values)
.language-js {
  .mi + .nx, .mf + .nx { color: $article-code-accent5; }
}
```

This incidentally colorized line protocol's integer/uinteger suffixes (`12485903i`, `1u`) when those blocks were fenced `js`. Chroma has no `lp` lexer, so `class=language-lp` blocks render with zero token spans — confirmed via build output on this branch. Result: the float/integer/uinteger/boolean/timestamp/special-characters/comment field-value examples in `content/shared/v3-line-protocol.md` and `content/shared/influxdb-v2/reference/syntax/line-protocol.md` go from (wrongly-labeled but) colorized to flat, unhighlighted text.

Not a template bug — nothing breaks, no build/render error. `lp` fence support is new to this repo; this is the first real-world content landing on `lp` for these particular blocks, so there's no prior precedent being broken, just no highlighting gained either.

jstirnaman:
> ~## Known gap: `lp` fence has no syntax highlighting~
> 
> Converting these blocks from `js`/`sh` to `lp` fixes the mislabeling (they were never valid JS/shell), but introduces a visual regression worth tracking.
> 
> **What's lost:** `assets/styles/layouts/_syntax-highlighting.scss` has a `.language-js` rule (originally added for Flux duration/time values, e.g. `1h30m`) that colors a number token immediately followed by an identifier token:
> 
> ```css-scss
> // Javascript / Flux specific styles (duration and time values)
> .language-js {
>   .mi + .nx, .mf + .nx { color: $article-code-accent5; }
> }
> ```
> 
> This incidentally colorized line protocol's integer/uinteger suffixes (`12485903i`, `1u`) when those blocks were fenced `js`. Chroma has no `lp` lexer, so `class=language-lp` blocks render with zero token spans — confirmed via build output on this branch. Result: the float/integer/uinteger/boolean/timestamp/special-characters/comment field-value examples in `content/shared/v3-line-protocol.md` and `content/shared/influxdb-v2/reference/syntax/line-protocol.md` go from (wrongly-labeled but) colorized to flat, unhighlighted text.
> 
> Not a template bug — nothing breaks, no build/render error. `lp` fence support is new to this repo; this is the first real-world content landing on `lp` for these particular blocks, so there's no prior precedent being broken, just no highlighting gained either.

Fixed with:

  - b7765262a feat(line-protocol): add fences highlighting and validation
  - e691ae243 style(test): format line protocol regression check

layouts/partials/lp/highlight.html
sanderson, line 1: I think there's a lot more we could do for syntax highlighting line protocol, but it's probably beyond this PR and may even be worth contributing to Chroma upstream. I think this is good enough for now.

sanderson (approved):

layouts/partials/lp/highlight.html
jstirnaman: > I think there's a lot more we could do for syntax highlighting line protocol, but it's probably beyond this PR and may even be worth contributing to Chroma upstream. I think this is good enough for now.

My thoughts exactly. I considered starting the Chroma work, but it'll have to wait for now.

Issue #6765, opened 2026-01-26
Configure for object storage should include popular storage providers


jstirnaman:
https://docs.influxdata.com/influxdb3/enterprise/get-started/setup/#object-store-examples provides examples for using different storage providers, but https://docs.influxdata.com/iinfluxdb3/enterprise/admin/object-storage/ only describes Minio. It should at least include the same provider examples as well as other configuration details.

PR #7710, opened 2026-08-26
docs(influxdb3): move object store requirements to general page

### LINKED ISSUE #7444: docs(influxdb3): move object store requirements to general page; drop partial capitalization of "object store"
## 1. Move general object store requirements out of the MinIO guide

#7442 added a "Object store requirements" section (consistency semantics +
`influxdb3 debug object-store-check` verification) to
`content/shared/influxdb3-admin/object-storage/minio.md`. This content
describes requirements for any S3-compatible backend, not anything
MinIO-specific — the copy itself says "any S3-compatible object store,
including MinIO."

`content/shared/influxdb3-admin/object-storage/_index.md` already serves as
the landing page listing all object-store provider guides (via
`{{< children >}}`). Move the "Consistency semantics" and "Verify your
object store" subsections there, and link to them from `minio.md`, keeping
only the MinIO-specific callouts (NFS warning, shared-volume warning, POSIX
note) in the MinIO guide. This avoids duplicating the requirement if/when
other provider guides (S3, GCS, etc.) are added.

## 2. Drop partial capitalization of "object store"

`.ci/vale/styles/InfluxDataDocs/Branding.yml` has a substitution rule:

    'the object store': the Object store

alongside the Compactor/Ingester/Querier rules. Those three are true proper
nouns — internal subsystems unique to InfluxDB 3 with no risk of colliding
with a generic term. "Object store" doesn't clear that bar: it's an
industry-standard term (S3, MinIO, and GCS are all "object stores" too), so
capitalizing it doesn't disambiguate anything. It also only matches the
literal string "the object store," so phrasing like "any S3-compatible
object store" or "your object store" slips through unchanged — an
inconsistency that showed up in #7442.

Recommend:
- Remove the `'the object store': the Object store` swap from
  `Branding.yml`.
- Use sentence case ("object store") by default.
- Use "InfluxDB 3 object store" (not partial capitalization) on the rare
  occasion disambiguation from a third-party/generic object store is
  actually needed and isn't already clear from context.
- Expand on examples in https://docs.influxdata.com/influxdb3/enterprise/get-started/setup/#object-store-examples
- Copy Supported Object Stores from https://docs.influxdata.com/influxdb3/enterprise/admin/backup-restore/

Related: #7442 (source of both issues)

jstirnaman:
## Summary
- Moves "Consistency semantics" and "Verify your object store" out of the MinIO guide into the object-storage landing page (`_index.md`), since they apply to any backend, not just MinIO. `minio.md` keeps only its own NFS/shared-volume/POSIX warnings and links back to the shared requirements section.
- Scopes "Verify your object store" (`influxdb3 debug object-store-check`) to Enterprise only. Verified against live 3.11.2 Core/Enterprise containers: Core's `debug --help-all` lists only `catalog`; Enterprise also lists `object-store-check`. The internal 3.10 docs plan confirms this command is Enterprise-only. Previously the shared guide presented it unconditionally to both editions, and Core even had a full CLI reference stub for a command its binary doesn't have — removed that stub, matching the no-stub pattern used for other Enterprise-only commands (e.g. `remove node`).
- Removes the `'the object store': the Object store` Vale substitution rule — "object store" is an industry-standard term, not an InfluxDB-specific proper noun — and fixes the resulting bad casing in `minio.md`.

## Why
Part of #7444. Sets up a single home for object-store requirements before #6765 adds S3, GCS, and Azure provider guides alongside MinIO — those guides can link to one shared section instead of repeating it. The Enterprise-only scoping bug was found while researching this move; documenting the enterprise-only tool for Core would have carried the same bug into every new provider guide.

Closes #7444.

## Impact
- Core users no longer see instructions for a command their binary doesn't support.
- Enterprise users see the same content as before, just relocated.
- No behavior change to `minio.md`'s setup/configure/confirm steps.

## Test plan
- [x] `npx hugo --quiet` builds clean
- [x] Confirmed rendered output: Enterprise object-storage pages include "Verify your object store"; Core pages don't
- [x] `yarn lint-codeblocks` on changed shared files: all code blocks pass
- [x] Vale on changed stub pages: 0 errors (pre-existing unrelated warnings only)
- [x] `link-checker` on changed pages: 0 errors (only known false-positive directory-fragment warnings)


reidkaufmann (approved):

jstirnaman:
Addressed both review findings from #7713 (comment [5442603939](https://github.com/influxdata/docs-v2/pull/7713#issuecomment-5442603939)) in e53b7f22e. Both findings were about content in this PR's diff, not #7713's, since #7713 stacks on this branch and GitHub showed the combined diff.

**Verified against live 3.11.2 Enterprise** (`influxdb3 debug object-store-check --help-all`):
- `--check-prefix` is required (`Usage: ... --check-prefix <CHECK_PREFIX>`), not optional as the CLI reference stated. Fixed the flag table and added its `INFLUXDB3_OSCHECK_PREFIX` env var.
- `--probe-prefix` is read-only for the catalog-replay step, but the command still runs synthetic write/delete checks under `--check-prefix` first — the `--help-all` examples confirm this. Clarified in both the shared guidance (`_index.md`) and the CLI reference, and added the missing `--check-prefix` to the probe example, which previously wouldn't have run as shown.

**Fixed (2/2), 0 dismissed:**
1. High — `--probe-prefix` "read-only mode" claim: fixed, both files.
2. Medium — `--check-prefix` required + env var, probe example missing the flag: fixed.

**Quality pipeline:**
- `npx hugo --quiet`: clean
- `yarn lint-codeblocks` on both changed files: all blocks pass
- Vale on stub pages sourcing this content: 0 errors (pre-existing unrelated "admin" word-list warnings only)
- `link-checker` on stub pages: 0 errors

PR #7716, opened 2026-08-27
fix(api): render trigger_settings fields, correct ApiNodeSpec schema

### LINKED ISSUE #7715: InfluxDB 3 API reference: trigger_settings fields missing, ApiNodeSpec schema is wrong
Page: https://docs.influxdata.com/influxdb3/enterprise/api/processing-engine/#operation/PostConfigureProcessingEngineTrigger (Core has the same problem)

**1. `trigger_settings` has no fields**

It renders as a required `string` with no sub-properties, and the example shows `"trigger_settings": "TRIGGER_SETTINGS"`. It's an object:

- `run_async` (boolean, required) - default `false`
- `error_behavior` (string, required) - one of `log`, `retry`, `disable`, default `log`

The schemas are already in `api-docs/influxdb3/enterprise/influxdb3-enterprise-openapi.yaml`, but the properties wrap the `$ref` in `allOf`, which the renderer seems to collapse to `string`. `node_spec` renders the same way.

**2. `ApiNodeSpec` is wrong**

The spec documents it as `{node_id, node_group}`. The API actually accepts either the string `"all"` or `{"nodes": ["NODE_ID"]}`.

caterryan:
Closes #7715


jstirnaman (commented):
- I'll send the OpenAPI schema fixes upstream to influxdata/docs-tooling
- We should add Cypress tests for the partial changes and additions

jstirnaman (approved):

PR #7717, opened 2026-08-27
docs(influxdb3): remove last-write-wins guarantee for duplicate points

reidkaufmann:
## Summary

Removes the remaining last-write-wins guarantees from the shared line-protocol "Duplicate points" section (rendered for Core, Enterprise, and Cloud):

- Drops the unqualified "field set becomes the union … conflicts favor the new field set" promise. Duplicate writes still merge into a single point, but which value is retained when both set the same field is not guaranteed.
- Replaces the Enterprise/Cloud "Overwrites need time between writes / 30 minutes" warning with the same "Overwrites are not deterministic" warning Core already carries, and lists the factors the outcome depends on (write rate, snapshot timing, compaction state, ingest node).
- Removes the Enterprise/Cloud "short means less than 30 minutes" sentence from the anti-patterns section.

Follow-up to #7682. Overwrite resolution depends on too many conditions (sparse-writer snapshot lag, compactor fence publication lag, multi-ingester ordering, in-buffer dedup ordering) for a fixed delay to be a safe recommendation. Related: influxdata/EAR#7056, influxdata/influxdb_pro#5352.

## Checklist

- [x] Shared source `content/shared/v3-line-protocol.md`; Dedicated/Clustered/Serverless blocks unchanged
- [ ] Docs review

### DIFF
--- a/content/shared/v3-line-protocol.md
+++ b/content/shared/v3-line-protocol.md
@@ -285,9 +285,10 @@ letter or a number. They can contain dashes (`-`) and underscores (`_`).
 
 A point is uniquely identified by the table name, tag set, and timestamp.
 If you submit line protocol with the same table, tag set, and timestamp,
-but with a different field set, the field set becomes the union of the old
-field set and the new field set, where any conflicts favor the new field set.
-Overwrite behavior is product-specific.
+InfluxDB stores a single point for that identity, merging the field sets of the
+duplicate writes.
+When duplicate writes set the same field, InfluxDB does not guarantee which
+value is retained.
 Read the following warning before relying on overwrites to maintain a last-value view.
 
 {{% show-in "cloud-dedicated,clustered,cloud-serverless" %}}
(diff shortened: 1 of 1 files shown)


jstirnaman (approved):
Thanks!

content/shared/v3-line-protocol.md
```diff
@@ -299,32 +300,16 @@ Read the following warning before relying on overwrites to maintain a last-value
...
-{{% show-in "core" %}}
+{{% show-in "core,enterprise,cloud" %}}
 > [!Warning]
 > #### Overwrites are not deterministic
 >
-> Overwriting a point is not reliable in {{% product-name %}}, regardless of the
-> delay between writes: queries may return either version, and either version may
-> be permanently stored.
+> Overwriting a point (same table, tag set, and timestamp) is not reliable in
+> {{% product-name %}}, regardless of the delay between writes: queries may
```
anjaflux, line 308: I'm confused - but I thought we did recommend a delay for the single ingester case? Is that recommendation changing?

reidkaufmann: [Anja said to ignore](https://github.com/influxdata/EAR/issues/7056#issuecomment-5445741274) the question above because she hadn't seen my EAR update.

PR #7718, opened 2026-08-28
docs(api): document Parquet cleanup and upgrade retry endpoints

jstirnaman:
## What changed

- Adds API reference entries for two Enterprise storage-engine-upgrade
  maintenance endpoints: `POST/GET .../upgrade/parquet_cleanup` and
  `POST .../upgrade/retry_parquet_to_pacha_tree`.
- Adds `influxdb3 manage` CLI reference pages: `cleanup-parquet` and
  `retry-upgrade-to-pacha-tree`, plus a section index and a link from
  the top-level `influxdb3` command list.
- Fixes the load capture profile schema: the request-type field is
  `type`, not `capture_type`, and the node identifier is a number, not
  a string.
- Adds the load capture profile detail, preview, and download API
  reference entries.
- Adds a note about the upgraded storage engine requirement to the
  Parquet export endpoints.
- Corrects the resource token endpoint's description of InfluxDB 3
  Cloud support: Cloud has database-scoped read and write tokens, not
  full resource tokens.
- Documents that deleting a load capture profile while it's still
  capturing returns an error (API 409 and the corresponding CLI
  command).

## Why

`cleanup-parquet` and `retry-upgrade-to-pacha-tree` shipped without
CLI or API reference pages. The load capture and export gaps, and the
delete-while-capturing behavior, surfaced while adding and verifying
this content in the same area of the spec.

## Impact

New reference documentation for InfluxDB 3 Enterprise users running
or troubleshooting a storage engine upgrade. Corrects two schema
errors, one product-availability claim, and one undocumented error
response that were live in the published API reference.

## Verification

Tested against a running InfluxDB 3 Enterprise 3.11.2 instance:

- `parquet_cleanup` (GET/POST) and `retry_parquet_to_pacha_tree`
  (POST): confirmed 401 without a token, and the documented 409/404
  responses on a cluster that was never on the pre-upgrade storage
  engine.
- Load capture profile lifecycle: started a real capture, confirmed
  the list/detail/preview responses use `type` (not `capture_type`)
  and an integer `node_id`, downloaded the archive, and deleted the
  profile.
- Deleting a profile while it's still capturing: confirmed 409 from
  both the HTTP API and `influxdb3 loadcap delete` ("a capture is
  already in progress"); confirmed deleting the same profile succeeds
  once the capture finishes.
- `influxdb3 manage cleanup-parquet` and `retry-upgrade-to-pacha-tree`:
  ran `--status-only`, `--dry-run`, the `--dry-run`/`--yes` and
  `--dry-run`/`--status-only` conflicts, and an invalid token, and
  confirmed each against the actual CLI output.
- Frontmatter validated for all new/changed CLI reference pages.
- Cloud token claim checked against
  https://docs.influxdata.com/influxdb3/cloud/admin/tokens/.

Claude-Session: https://claude.ai/code/session_01XqThD3yqz8y64vkaTgQDvR

### DIFF
--- a/content/influxdb3/enterprise/reference/cli/influxdb3/_index.md
+++ b/content/influxdb3/enterprise/reference/cli/influxdb3/_index.md
@@ -32,6 +32,7 @@ influxdb3 [GLOBAL-OPTIONS] [COMMAND]
 | [import](/influxdb3/enterprise/reference/cli/influxdb3/import/)   | Manage bulk Parquet imports         |
 | [install](/influxdb3/enterprise/reference/cli/influxdb3/install/) | Install plugins                     |
 | [loadcap](/influxdb3/enterprise/reference/cli/influxdb3/loadcap/) | Capture and manage workload profiles |
+| [manage](/influxdb3/enterprise/reference/cli/influxdb3/manage/)   | Perform cluster maintenance operations |
 | [query](/influxdb3/enterprise/reference/cli/influxdb3/query/)     | Query {{% product-name %}}          |
 | [remove](/influxdb3/enterprise/reference/cli/influxdb3/remove/)   | Remove stopped nodes                |
 | [serve](/influxdb3/enterprise/reference/cli/influxdb3/serve/)     | Run the {{% product-name %}} server |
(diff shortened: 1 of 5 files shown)


hiltontj (commented):
LGTM

PR #7728, opened 2026-09-01
fix(api): clarify trigger arguments in Core and Enterprise

### LINKED ISSUE #7727: InfluxDB 3 API reference: clarify trigger_arguments type and usage
Page: https://docs.influxdata.com/influxdb3/enterprise/api/processing-engine/#operation/PostConfigureProcessingEngineTrigger

The trigger_arguments request-body field only says "Optional arguments passed to the plugin." It is unclear what callers should provide.

The API accepts an optional string-to-string map: Option<HashMap<String, String>>. When present, InfluxDB passes it to the configured plugin entry point as args, a Python dict[str, str]; when omitted, args is None. The names and string values are plugin-specific.

Example valid request member:

"trigger_arguments": {
  "threshold": "90",
  "notify_email": "admin@example.com"
}

The Enterprise OpenAPI schema is close but not quite accurate: it declares type: object and additionalProperties: true, which permits any JSON value. It should constrain additional property values to strings (for example, additionalProperties: { type: string }) and describe that this is the plugin-specific args mapping. Adding the JSON example would make the API reference self-contained.

Core has the same schema and should receive the same correction.

Research:

- influxdb_pro/ent/influxdb3_types/src/http.rs: request type is trigger_arguments: Option<HashMap<String, String>>.
- influxdb_pro/oss/influxdb3_py_api/src/py_conversion.rs: converts it to an optional Python dictionary.
- influxdb_pro/oss/influxdb3_py_api/src/system_py.rs: passes it as the args parameter to WAL, scheduled, and request plugin entry points.
- Existing plugin API reference documents args as a string-to-string mapping.

Related: #7715 / #7716 fixed the API renderer and showed the expected Core-and-Enterprise spec update pattern.

caterryan:
Closes #7727

What changed:
- Define trigger_arguments as an object whose values must be strings in both the Core and Enterprise processing-engine API schemas.
- Explain that it is the plugin-specific args mapping, and that omitting it passes None to the plugin.
- Add threshold and target_table example values so generated request bodies are directly usable.
- Add Cypress coverage for rendering object schema descriptions, JSON examples, and generated request values.

Why:
The server accepts an optional HashMap<String, String> and passes it to the plugin entry point as a Python dict[str, str]. The prior additionalProperties: true schema allowed values the API rejects.

Verification:
- Parsed both YAML specs and asserted the string-map schema and examples.
- Ran sh api-docs/generate-api-docs.sh --no-build.
- Ran npx hugo and inspected both generated processing-engine API pages.
- Ran Redocly lint: Core is valid with 6 existing warnings; Enterprise retains 12 pre-existing errors and 20 warnings, none in the changed schema.
- Ran node cypress/support/run-e2e-specs.js --spec "cypress/e2e/content/api-reference.cy.js" --no-mapping (47 passing).

Note:
Core and Enterprise v3 specs are ported from docs-tooling. Keep the matching source definitions in sync to preserve this correction on future ports.

test2 docs influxdata com_pr-preview_pr-7728_influxdb3_enterprise_api_processing-engine_

### DIFF
--- a/api-docs/influxdb3/core/influxdb3-core-openapi.yaml
+++ b/api-docs/influxdb3/core/influxdb3-core-openapi.yaml
@@ -2904,8 +2904,15 @@ components:
           example: cron:0 0 6 * * 1-5
         trigger_arguments:
           type: object
-          additionalProperties: true
-          description: Optional arguments passed to the plugin.
+          additionalProperties:
+            type: string
+          description: |
+            Optional, plugin-specific arguments passed to the plugin entry point as its `args` dictionary.
+            Provide string key-value pairs. Omit this field to pass `None` as `args`.
+            Refer to the plugin documentation for supported argument names and values.
+          example:
+            threshold: "90"
+            notify_email: admin@example.com
         disabled:
           type: boolean
           default: false
--- a/api-docs/influxdb3/core/influxdb3-core-openapi.yaml
+++ b/api-docs/influxdb3/core/influxdb3-core-openapi.yaml
@@ -2912,7 +2912,7 @@ components:
             Refer to the plugin documentation for supported argument names and values.
           example:
             threshold: "90"
-            notify_email: admin@example.com
+            target_table: alerts
         disabled:
           type: boolean
           default: false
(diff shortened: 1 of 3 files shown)


jstirnaman (approved):
- Added a Cypress test to cover the layout for this schema and using it as a test fixture.
- Ported spec changes upstream.

